PR #61, opened 2020-08-03
added export as HTML warning

alpkaanaksu:


scanf:
Great work @alpkaanaksu, thanks 🙏

PR #193, opened 2020-10-11
Change max-width from 90% to 80%

alpkaanaksu:
I saw what people were complaining about. 90% is too big for AnkiDroid. If you change it to 80% everything works fine.
Leave it like this for some time. If people continue to complain, I'll update the custom CSS to be more minimal and compatible. But I still think 80% max-width will solve the problem.


aalemayhu (approved):
✅ LGTM, thanks!

PR #218, opened 2020-11-15
New Earthfile instead of Dockerfile

Mobilpadde:
I've decided to put in the good ol' n2a image but it get's cached to make the deployments easier 🥲

Also a few changes in the makefile. I did keep the `docker_run` but renamed it to `debug`, yet I still have to understadn what it does 😖


Mobilpadde:
Might have to change a few things, as I'm not sure how the deployment process is? As far as I remember you're using dokku, but no comments about it - at least I don't see any that is.

aalemayhu (changes requested):
I am fine with adding the earthfile but we are going to remove the Dockerfile. Please revert that change. Otherwise I can't accept it as is. Thanks!

Mobilpadde:
> I am fine with adding the earthfile but we are going to remove the Dockerfile. Please revert that change. Otherwise I can't accept it as is. Thanks!

It's back again! :+1:

aalemayhu (approved):
✅ LGTM, thanks!

PR #259, opened 2021-03-13
Clozes treatment

Maximus0000:
Edited the way clozed are treated so that Latex Formulas can be inserted


aalemayhu (approved):
LGTM, thanks!

PR #260, opened 2021-03-16
Added a way to check if the current URL is dev.2anki.net or not. issue#257

Secretmud:
This lets us append a dev icon if this is the case (aka dev.2anki.net).

Have only tested this on my localhost, but it hopefully woks on the sever.

File(s) edited:
n2a-header.imba

Wasn't sure if this should have been kept in a separate file or not.


aalemayhu (approved):
LGTM, thanks!

PR #308, opened 2021-07-08
server: web: add new card option for linking back to Notion page

aalemayhu:
This has been requested a few times now so adding it.

Closes: https://github.com/alemayhu/Notion-to-Anki/issues/201


aalemayhu:
- [ ] Need to investigate why it's creating duplicates.

PR #425, opened 2021-12-17
server: 244fb4c6ad56 (parser: don't enforce suffix, 2021-12-17)

aalemayhu:
This fixes an issue were images without a file suffix were being broken
in the flashcards.


aalemayhu:
Fixes: https://github.com/alemayhu/2anki.net/issues/417

PR #857, opened 2022-10-12
add RemNote to Readme

moritzWa:
Hi, would love the small shoutout (to RemNote) here!! This looks great.


aalemayhu:
Thank you for the suggestion, but I don't use RemNote.

moritzWa:
sad

PR #1053, opened 2023-02-19
chore: bring back create_deck into the repository

aalemayhu:
This will reduce overhead in maintaince of the project.


aalemayhu:
dropped

PR #1275, opened 2023-08-06
fix: downgrade the missing relative path

aalemayhu:
After reviewing several bug reports in Bugsnag and reproducing.
It's possible to export a page in Notion without the images


aalemayhu:
Screenshot 2023-08-06 at 7 07 50 AM

PR #1277, opened 2023-08-06
refactor: cleaning up the codebase

aalemayhu:
Making the big functions small and reducing arguments in 
the function calls.


aalemayhu:
I love the idea of refactoring but we need to focus on the highest painpoints. Closing this in favour of hyper focus.

Issue #1382, opened 2023-12-24
\sqrt functions invisible


MorgenmuffelLP:
Hey there, so i converted my cards through notion2anki, but i had the problem that the root functions arent visible, even though the \sqrt{} is in the code

does anyone know how to fix this? i dont want to fix each individual card
![math](https://github.com/2anki/2anki.net/assets/37043556/e2dbc9f5-0bbd-4d0a-9760-8db3c5e2a525)

aalemayhu:
Someone asked the same thing on [Reddit](https://www.reddit.com/r/notion2anki/comments/18pyrvo/latex_sqrt_function_bugged/) . This is a bug, I will fix it.

PR #1388, opened 2023-12-28
fix: do not add empty lines in the path `d` attribute

aalemayhu:
These are coming directly from Notion. No idea why they are doing it
this way. It was not like this before. What basically happens is that
our `preserveNewlinesIfApplicable` function will add `<br />` to the tag
and it will totally break the KaTex.

Fixes: https://github.com/2anki/2anki.net/issues/1382

### DIFF
--- a/src/lib/parser/DeckParser.ts
+++ b/src/lib/parser/DeckParser.ts
@@ -139,17 +139,35 @@ export class DeckParser {
     }
   }
 
+  removeNewlinesInSVGPathAttributeD(html: string): string {
+    const dom = cheerio.load(html);
+    const pathElements = dom('path');
+
+    for (const pathElement of pathElements) {
+      if ('attribs' in pathElement && 'd' in pathElement.attribs) {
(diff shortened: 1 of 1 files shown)


wernerkrauss:
Is this fix live? I still get invisible '\sqrt{}" in my anki cards :|

Thanks for your really great tool! It helps a lot!

aalemayhu:
Weird, this was done one month ago but the commit was actually not deployed to the server. I will investigate that, but can you try again @wernerkrauss. The change should be there. If you still experience it, please share a minimal example I can use to reproduce, thanks!

aalemayhu:
Turns out the issue was that git did not want to pull in the changes since there was a local diff due to the build process on-prem. Hopefully resolved here https://github.com/2anki/2anki.net/pull/1402/files#diff-5170cc1d84dcd5444b8bb0bbb988370a206cc5cc5f40b41819d8a60f30d4fe8dL30

wernerkrauss:
@aalemayhu Thanks a lot! It's working now!

PR #1410, opened 2024-02-06
feat: allow for cloze deletions from paragraphs

aalemayhu:


aalemayhu:
This breaks compatibility :/

PR #1563, opened 2024-08-03
feat: add user avatars

aalemayhu:
Will be used in a new navigation bar.


migrations/20240803100001_add_missing_user_picture.js
```diff
@@ -0,0 +1,18 @@
+const { v4: getRandomUUID } = require('uuid');
+
```
aalemayhu, line 2: unused getRandomUUID

PR #1578, opened 2024-08-18
feat: add support for external uploads

aalemayhu:
This brings Google Drive and Dropbox support.  It will be exposed in the
next UI update on 2anki.net.

Closes: https://github.com/2anki/server/issues/1559

### DIFF
--- a/src/controllers/Upload/helpers/handleUploadLimitError.ts
+++ b/src/controllers/Upload/helpers/handleUploadLimitError.ts
@@ -1,29 +1,13 @@
 import express from 'express';
-import { useDefaultEmailService } from '../../../services/EmailService/EmailService';
-import UsersService from '../../../services/UsersService';
-import UsersRepository from '../../../data_layer/UsersRepository';
-import { getDatabase } from '../../../data_layer';
 
-export const handleUploadLimitError = async (
-  req: express.Request,
+const RATE_LMIT = 429;
+
+export const handleUploadLimitError = (
+  _req: express.Request,
   response: express.Response
 ) => {
-  const owner = response.locals.owner;
-
-  // If the user is already logged in, redirect to the pricing page
-  if (owner) {
-    const database = getDatabase();
-    const emailService = useDefaultEmailService();
-    const usersService = new UsersService(
-      new UsersRepository(database),
-      emailService
-    );
-
-    const user = await usersService.getUserById(response.locals.owner);
-    if (user) {
-      return response.redirect('/pricing?error=upload_limit_exceeded');
-    }
-  }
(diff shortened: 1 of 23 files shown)


src/controllers/Upload/helpers/handleUploadLimitError.ts
```diff
@@ -1,29 +1,13 @@
...
-    if (user) {
-      return response.redirect('/pricing?error=upload_limit_exceeded');
-    }
-  }
-
-  response.redirect('/login?error=upload_limit_exceeded');
+  response.status(RATE_LMIT).json({
+    error:
+      'Upload limit of 100 flashcards exceeded, become premium to remove limit or reduce upload size.',
+  });
```
aalemayhu, line 12: This change has to be reverted for compat

src/lib/misc/isEmptyPayload.tsx
```diff
@@ -5,5 +5,5 @@ export const isEmptyPayload = (files: UploadedFile[] | undefined) => {
     return true;
   }
   const totalBytes = files.reduce((acc, file) => acc + file.size, 0);
-  return totalBytes > 0;
+  return totalBytes === 0;
 };
```
aalemayhu, line 9: What is the consequence of this fix?

src/lib/misc/ErrorHandler.tsx
```diff
@@ -10,21 +9,7 @@ import { isLimitError } from './isLimitError';
...
-      </div>
-      <div className="card" style={{ width: '50%', padding: '1rem' }}>
-        <details>
-          <summary>This the front</summary>
-          This is the back
-        </details>
-      </div>
-    </>
-  )
+  'Deck creation failed. Ensure a valid toggle or check settings. Note: Toggle headings are only supported in the Notion integration. Avoid using Markdown; use HTML instead.'
```
aalemayhu, line 12: Revert this for compat

src/lib/constants.ts
```diff
@@ -3,6 +3,7 @@ import path from 'path';
 export const TEMPLATE_DIR = path.join(__dirname, '../templates');
 
 export const ALLOWED_ORIGINS = [
+  'http://localhost:3000',
   'http://localhost:8080',
```
aalemayhu, line 7: revert

src/lib/misc/isEmptyPayload.tsx
aalemayhu: Totally safe since it was determining if the payload should be preserved for debugging. In this case the check was wrongly persisting empty uploads.

PR #1585, opened 2024-08-24
feat: add header for retrieving the workspace for the first deck

aalemayhu:
Will be used for previewing.


aalemayhu:
Closing this for now and we will get back to it later

PR #1589, opened 2024-08-31
feat: Implement worker threads for upload processing

aalemayhu:
- Offload flashcard generation to a separate worker thread
- Prevents blocking of main thread during uploads, improving responsiveness
- Addresses bug causing server slowdown during large uploads

Refs:
https://nodejs.org/api/worker_threads.html
https://stackoverflow.com/questions/61095741/how-do-i-avoid-blocking-an-express-rest-service

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:watch": "jest --watch",
     "start": "npm run dev",
     "build": "tsc -p .",
-    "dev": "NODE_OPTIONS=\"--max-old-space-size=8192\" LOCAL_DEV=true nodemon --watch '**/*.ts' --exec \"ts-node\" src/server.ts",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "test:watch": "jest --watch",
     "start": "npm run dev",
     "build": "tsc -p .",
-    "dev": "LOCAL_DEV=true nodemon --watch '**/*.ts' --exec \"ts-node\" src/server.ts",
+    "dev": "NODE_OPTIONS=\"--max-old-space-size=8192\" LOCAL_DEV=true nodemon --watch '**/*.ts' --exec \"ts-node\" src/server.ts",
     "lint": "eslint '**/*.ts' '**/*.tsx'",
```
aalemayhu, line 26: this should be omitted

PR #1591, opened 2024-08-31
feat: Implement worker threads for upload processing

aalemayhu:
Unfortunately the node worker threads cannot share buffers reliably with the main thread. I learned this the hard way 😅
So I have move all of the read writes of the APKG payload to be be based on file system access which is thread safe since we only read after writes.

While reviewing the code path, I discovered that we actually do a lot of wasteful reads and writes. Essentially for every APKG generation we make 2x calls due to the renaming to provide the uploader with friendly name. I have not measured how long this takes but instead of double work we do this one time and rely on the create_deck having the right name.

Revert of revert https://github.com/2anki/server/pull/1589
Reference: https://github.com/2anki/create_deck/pull/93


src/controllers/DownloadController.ts
```diff
@@ -61,7 +61,12 @@ class DownloadController {
           return;
         }
 
-        const page = DownloadPage({ id, files });
+        console.log('files', files);
+
```
aalemayhu, line 65: remove debug log

src/controllers/DownloadController.ts
```diff
@@ -61,7 +61,10 @@ class DownloadController {
           return;
         }
 
-        const page = DownloadPage({ id, files });
+        const page = DownloadPage({
+          id,
+          files: files.filter((file) => file.endsWith('.apkg')),
+        });
```
aalemayhu, line 67: we do not want to index images and other files

PR #1640, opened 2024-11-10
feat: new card option for disabling indented bullets

aalemayhu:
This is useful for creating cloze deletions on the front and reducing lots of unnecessary cards created.


piajill:
Hello, great feature! I make my flashcards via 2anki like this, but i have a problem with inserting pictures under a indented bullet. The image won't load, only if i put it under a toggle in a seperate card - is there any way i could change that, so that the image will come up with the bullet points?

aalemayhu:
@piajill please open up a new issue with example and steps to reproduce. https://github.com/2anki/server/issues/new/choose

Thanks!

PR #1687, opened 2024-12-21
fix: resolve crash caused by images being overwritten with HTML

aalemayhu:
This is backwards compatible and should only trigger with the
imageQuizHtmlToAnki card option.

### DIFF
--- a/src/lib/parser/PrepareDeck.ts
+++ b/src/lib/parser/PrepareDeck.ts
@@ -1,7 +1,12 @@
 import getDeckFilename from '../anki/getDeckFilename';
 import { DeckParser, DeckParserInput } from './DeckParser';
 import Deck from './Deck';
-import { isImageFile, isPDFFile, isPPTFile } from '../storage/checks';
+import {
+  isHTMLFile,
+  isImageFile,
+  isPDFFile,
+  isPPTFile,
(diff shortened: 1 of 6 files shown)


src/lib/parser/PrepareDeck.ts
```diff
@@ -26,9 +35,18 @@ export async function PrepareDeck(
...
+      convertedImageFiles.push({
+        name: `${file.name}.html`,
+        contents: convertedImageContents,
+      });
+
+      fs.writeFileSync(
+        `/Users/scanf/Downloads/${file.name}.html`,
+        convertedImageContents
+      );
     }
```
aalemayhu, line 50: I will remove this debug write out

src/usecases/uploads/isZipContentFileSupported.ts
```diff
@@ -14,4 +15,5 @@ export const isZipContentFileSupported = (filename: string) =>
   isMarkdownFile(filename) ??
   isPlainText(filename) ??
   isCSVFile(filename) ??
-  isPDFFile(filename);
+  isPDFFile(filename) ??
+  isImageFile(filename);
```
aalemayhu, line 19: allow image uploads

PR #1946, opened 2026-04-08
fix: handle venv for wsl

aalemayhu:
## Summary by CodeRabbit

* **Chores**
  * Improved Python interpreter detection logic to support virtual environment installations and provide better fallback handling across different system configurations.


aalemayhu:
@copilot Verify each finding against the current code and only fix it if needed.

Nitpick comments:
In `@src/lib/anki/CardGenerator.ts`:
- Line 10: The returned Windows Python path is built using the imported homedir
function but it's not invoked, producing a string like "[Function:
homedir]\\AppData..."; call homedir() (the function) when constructing the path
in CardGenerator.ts so the actual user home directory is used (e.g., replace
homedir with homedir()) and consider using path.join with homedir() for safer
path concatenation.

PR #1965, opened 2026-04-15
feat: async Claude AI flashcard pipeline with S3 promotion and restart support

aalemayhu:
- **Async Claude conversion**: file uploads with `claudeAIFlashcards` enabled now fire a background worker, create a tracked job, and stream chunk-level progress to
  the client (3 s polling when active, 10 s when idle)
  - **Parallel chunk processing**: all HTML chunks are sent to the Claude API concurrently via `Promise.all`, reducing a 4-chunk document from ~6 min to ~90 s
  - **S3 promotion on completion**: finished APKGs are uploaded to DigitalOcean Spaces and an `uploads` record is created — identical durability to Notion conversions —
   then the job record is deleted
  - **Restart support**: interrupted Claude jobs (marked on server startup) can be re-run from the saved workspace files via `POST /api/upload/jobs/:jobId/restart`
  - **Race condition fix**: `updateJobStatus` guards non-terminal progress writes with `WHERE status NOT IN (terminal)` so in-flight progress updates can never
  overwrite a completed status
  - **Startup fix**: the blanket `UPDATE jobs SET status = 'failed'` is scoped to non-Claude, non-terminal jobs so done/interrupted Claude jobs survive restarts
  correctly

  ## Test plan

  - [ ] Upload a Notion ZIP export with Claude AI enabled — job appears in Downloads, progress updates every few seconds, completes and moves to Ready to Download
  - [ ] Kill the server mid-conversion — job shows as Interrupted on restart, Restart button works and produces a downloadable deck
  - [ ] Restart the server after a successful conversion — APKG remains in Ready to Download (not reset to failed)
  - [ ] Upload without Claude enabled — existing sync conversion flow unaffected

## Release Notes

* **New Features**
  * Added ability to download completed job results directly
  * Introduced job restart functionality for interrupted processing
  * Enabled asynchronous job processing with progress tracking for AI flashcard generation

* **Bug Fixes**
  * Improved handling of interrupted jobs during server startup

* **Chores**
  * Added automated code generation tooling for database schema types


aalemayhu:
@coderabbitai review

PR #2044, opened 2026-05-07
chore: route dependabot PRs through a staging branch

aalemayhu:
## Summary

Point dependabot at a long-lived `dependabot-updates` branch and add three workflows to manage the flow back to `main`. Collapses N weekly dependency-bump production deploys into one.

## How it works

- `.github/dependabot.yml` — npm (root + `web/`) and github-actions ecosystems now target `dependabot-updates` instead of `main`.
- `dependabot-auto-merge.yml` — auto-merges dependabot PRs into the staging branch once checks pass.
- `dependabot-sync-from-main.yml` — on every push to `main`, fast-forwards / merges into `dependabot-updates` so it never falls behind.
- `dependabot-open-merge-pr.yml` — whenever `dependabot-updates` moves, ensures one open PR back to `main` exists. No-ops if a PR already exists or branches are identical.

Reviewer just opens the existing PR and clicks merge — one production deploy covers every bump merged into staging since the last release.

## Bumps that landed via the new flow

- `html-to-text` 9.0.5 → 10.0.0
- `lint-staged` 16.4.0 → 17.0.2

## Test plan

- [ ] Merge, then push something trivial to `main` and confirm `dependabot-sync-from-main` runs cleanly.
- [ ] Wait for the next dependabot PR; confirm it targets `dependabot-updates` and `dependabot-auto-merge` enables auto-merge.
- [ ] After auto-merge lands, confirm `dependabot-open-merge-pr` opens a single PR `dependabot-updates` → `main`.
- [ ] Smoke-test the `html-to-text` v10 bump — Node engine requirement bumped to `>=20.19.0`; verify CI Node version satisfies it.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,7 +1,7 @@
 name: Dependabot auto-merge into staging
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - dependabot-updates
 
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "concurrently": "^9.2.1",
     "cross-env": "^10.1.0",
     "jest": "^30.2.0",
-    "lint-staged": "^16.1.2",
+    "lint-staged": "^17.0.2",
     "nodemon": "^3.0.1",
     "prettier": "^3.2.5",
     "pretty-quick": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "fflate": "^0.8.0",
     "find-remove": "^5.0.0",
     "get-notion-object-title": "^0.2.0",
-    "html-to-text": "^9.0.4",
+    "html-to-text": "^10.0.0",
     "jsonwebtoken": "^9.0.0",
     "knex": "^3.1.0",
     "mammoth": "^1.8.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -83,8 +83,8 @@ importers:
         specifier: ^0.2.0
         version: 0.2.8
       html-to-text:
-        specifier: ^9.0.4
-        version: 9.0.5
+        specifier: ^10.0.0
+        version: 10.0.0
       jsonwebtoken:
         specifier: ^9.0.0
         version: 9.0.3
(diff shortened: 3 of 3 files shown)


aalemayhu (commented):
**Verdict: request-changes** (one blocking item, otherwise solid)

CI is green and the workflow design is right — staging branch + sync + auto-open PR collapses N deploys into one, which is the actual win. Two concerns before merge.

## Blocking

**`dependabot-auto-merge.yml` will not get a write token under the `pull_request` trigger.**

Since Sept 2021, GitHub runs Dependabot-authored PRs with a *read-only* `GITHUB_TOKEN` on `pull_request` events (same as forked PRs). `gh pr merge --auto` requires `contents: write` *at runtime*, not just in the YAML header — the declared permissions can't grant what the token doesn't have. Expected failure: `GraphQL: Resource not accessible by integration`.

Recommended fix — switch to `pull_request_target` and keep the actor guard:

```yaml
on:
  pull_request_target:
    branches:
      - dependabot-updates

permissions:
  contents: write
  pull-requests: write

jobs:
  auto-merge:
    if: github.actor == 'dependabot[bot]'
    runs-on: ubuntu-latest
    steps:
      - name: Enable auto-merge
        run: gh pr merge --auto --squash "$PR_URL"
        env:
          PR_URL: ${{ github.event.pull_request.html_url }}
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

`pull_request_target` is safe here because: (a) we never check out PR code, (b) the `github.actor == 'dependabot[bot]'` guard prevents non-dependabot triggers, and (c) we run zero PR-controlled scripts. If you'd rather keep `pull_request`, the alternative is granting Dependabot write access at the org level (Settings → Code security → Dependabot), but `pull_request_target` is the smaller, more standard fix.

## Non-blocking nits

1. **`lint-staged@17` requires Node `>=22.22.1`; CI + `.nvmrc` are on `22.20.0`.** Currently harmless because the package is in devDependencies but not wired up (no `.husky/`, no `prepare` script, no `lint-staged` config block). Either bump `.nvmrc`/CI to `22.22.1+` or drop the unused dep. (`html-to-text@10` requires `>=20.19.0` — covered by 22.20.0, no action needed.)

2. **`dependabot-sync-from-main.yml` fails silently on merge conflicts.** `git merge --no-edit origin/main` exits non-zero, the workflow goes red, and you find out next time you check Actions. One-line improvement: add a `failure()` step that opens an issue or pings via `gh issue create`. Non-blocking — the failed run will still surface in the GitHub UI.

3. **`dependabot-open-merge-pr.yml` uses `--jq '.[0].number'`** — fine for the single-staging-branch flow (you'll never have two open `dependabot-updates → main` PRs by design), but worth a comment if anyone wonders later.

4. **Scope:** The two ride-along bumps (`html-to-text` 9→10, `lint-staged` 16→17) make sense as the proof-of-life that the new flow works. Accepting as-is.

## Goal alignment

Direct: fewer production deploys per dependency cycle = less reviewer toil = more time on user-facing work, which is the path to 300K. Worth shipping once the auto-merge token issue is resolved.

PR #2094, opened 2026-05-10
feat: simplify register form and drop required name field

aalemayhu:
## What
Trim the register flow to email + password. Drop the Name and Confirm-Password fields, lighten the heading, surface live password validation under the input, and add an "Already have an account? Log in" link. Lead with Google + new "or sign up with email" divider so the relationship between the two paths reads. Friendlier error copy on submit failures.

Backend keeps the `users.name` column unchanged. The register service now defaults `name` to the email local part (`jane.doe@example.com` -> `jane.doe`) when the client doesn't send one, so the account page, "Welcome back" header, and billing emails keep working without a migration. Older clients that still send a name continue to work.

## Why
Reducing the new-user surface from four fields to two is a direct hit on the goal in `CLAUDE.md`: "the simplest, fastest way to turn what they're studying into beautiful flashcards." Every removed field is a measurable conversion drop. Login is already single-password, so unifying register matches.

## How
- Server
  - `UsersService.register` derives `name` from the email local part when blank (`name?.trim() ?? ''` -> falls back to `email.split('@')[0]`).
  - `UsersController.register` no longer requires `name`; tightens the 400 message.
  - `isValidUser(password, email)` drops the `name` parameter; only one caller, updated in the same file.
- Web
  - `RegisterForm.tsx`: remove Name + Confirm Password, add aria-described password helper with three states (neutral / success / danger), add footer log-in link, update title to "Create your account", swap divider copy.
  - `auth.module.css`: small vertical-rhythm bump (card padding, field spacing) to balance the slimmer form, plus `.divider`, `.dividerLabel`, `.helpSuccess`, `.helpMuted`.
  - `app.document.json`: new keys `navigation.register.title` and `navigation.login.question`.

## Testing
- Unit tests added:
  - `src/services/UsersService.test.ts` - covers (a) supplied name passes through, (b) empty/whitespace name falls back to email local part.
  - `src/controllers/UsersControllers.test.ts` - covers register validation (400 without password, 400 without email), accepts a request with no `name`, accepts old clients that still send `name`, and 400s on duplicate email.
- `/check` (server tsc + web typecheck + web vitest + web lint) green.
- Manually verified: dev `vite preview` build serves `/register` with the new layout. Headless browser screenshot capture failed in this environment (missing `libnspr4` / no sudo to install Playwright deps); please grab a screenshot before un-drafting.

## Risks
- Existing users with empty `name` rows are unaffected (column already nullable).
- `loginWithGoogle` still uses Google's `name` directly, untouched.
- The `register` API contract gets *more* permissive (name optional). No breaking change for old clients.
- Rollback: revert the commit. No DB migration to undo.

## Goal alignment
Removing two fields from the only signup flow is a direct conversion win. Per `CLAUDE.md`, "drop something in, get a clean deck back" - if signup is two fields instead of four, more people get to the conversion path. That moves us toward the 300K-user goal.

## Out of scope
- Password reveal toggle (called out, separate PR).
- Login page changes other than mirroring the inverse footer link copy (handled via i18n key only; no JSX change to `LoginForm`).
- Any other auth flow.

### DIFF
--- a/src/controllers/UsersControllers.test.ts
+++ b/src/controllers/UsersControllers.test.ts
@@ -1,0 +1,143 @@
+import express from 'express';
+
+jest.mock('../lib/integrations/stripe', () => ({
+  getStripe: jest.fn().mockReturnValue({
+    customers: { retrieve: jest.fn() },
+    subscriptions: { retrieve: jest.fn(), cancel: jest.fn(), update: jest.fn() },
+  }),
+  updateStoreSubscription: jest.fn().mockResolvedValue(undefined),
+}));
+
+jest.mock('../services/SubscriptionService', () => ({
+  __esModule: true,
+  default: {
+    cancelUserSubscriptions: jest.fn(),
+    findRecentStripeSubscriptions: jest.fn(),
+  },
+}));
+
+import UsersController from './UsersControllers';
+import UsersService from '../services/UsersService';
+import AuthenticationService from '../services/AuthenticationService';
+
+const SAMPLE_PW = '12345678';
+
+const buildRes = () => {
+  const json = jest.fn();
+  const status = jest.fn().mockReturnValue({ json });
+  return { json, status } as unknown as express.Response & {
+    json: jest.Mock;
+    status: jest.Mock;
+  };
+};
+
+const buildController = (overrides?: {
+  getUserFrom?: jest.Mock;
+  register?: jest.Mock;
+  getHashPassword?: jest.Mock;
+}) => {
+  const userService = {
+    getUserFrom: overrides?.getUserFrom ?? jest.fn().mockResolvedValue(null),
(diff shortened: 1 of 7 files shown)


aalemayhu (commented):
Approve. Security review (separate thread) walked the diff against `.claude/rules/security.md`: no new SQL, no new logging of PII, no new shell or filesystem surface. Bcrypt path unchanged. The derived `name` is consumed by React text nodes and JSON responses (auto-escaped) — the only un-escaped sink (`EmailService` `{{name}}` HTML interpolation) is pre-existing and is currently called with `''` at every site, so this PR doesn't widen exposure. Sonar S7735 nit caught and fixed in 30257876.

All 5 checks green: web lint / build / test / playwright + Snyk.

Ships.

PR #2141, opened 2026-05-12
Fix reversed card sort order

### LINKED ISSUE #2137: fix: card ordering scrambled for reversed cards
## Problem

Card order in Anki doesn't always match toggle order from the Notion page. Users report the first toggle ending up last, second becoming fourth, etc.

## Root cause

In `DeckParser.ts`, cards are numbered sequentially (`card.number = counter++`), preserving DOM order. However, reversed cards get `number = -1` (line 590), and in `create_deck.py` line 178, cards with `number = -1` have their sort field replaced with their `notionId` (a string), producing a non-sequential value. Combined with the `basicReversed` feature appending extra notes after the main loop (line 593), this scrambles the perceived order.

## Fix direction

Assign reversed cards the same sequential number as their source card (or `sourceNumber + 0.5`) so they sort adjacent to the original, not at a random position determined by their notion ID string.

lewbei:
## Summary
- keep reversed cards on their existing numeric sort order instead of assigning `-1`
- make `Note.reversed()` companion cards sort immediately after the source card
- add regression coverage for reversed-card numbering

Fixes #2137

## Verification
- `pnpm test -- src/lib/parser/Note.test.ts --runInBand`
- `pnpm test -- src/lib/parser/DeckParser.test.ts --runInBand --testNamePattern "Markdown reversed cards keep numeric sort order"`
- `pnpm run build`

### DIFF
--- a/src/lib/parser/DeckParser.test.ts
+++ b/src/lib/parser/DeckParser.test.ts
@@ -140,6 +140,31 @@ test('Markdown nested bullet points', async () => {
   expect(deck.cards.length).toBe(4);
 });
 
+test('Markdown reversed cards keep numeric sort order', async () => {
+  const fixturePath = path.join(__dirname, '../../test/fixtures/simple-deck.md');
+  const contents = fs.readFileSync(fixturePath).toString();
+  const workspace = new Workspace(true, 'fs');
+  const parser = new DeckParser({
+    name: 'simple-deck.md',
+    settings: new CardOption({
+      'markdown-nested-bullet-points': 'true',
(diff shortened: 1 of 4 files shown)


aalemayhu:
Welcome to 2anki, @lewbei! 🎉 Thank you for your first contribution — really appreciate you jumping in to fix the reversed card sort order. The PR is well-structured with clear regression tests, which is exactly what we love to see. Looking forward to reviewing this!

aalemayhu (approved):
Elegant solution!

PR #2340, opened 2026-05-17
fix: restore verifyEmail error logging

### LINKED ISSUE #2295: Restore error log in verifyEmail before next(error)
Non-blocking observation from the security review of #2289.

In `src/controllers/UsersControllers.ts`, the `verifyEmail` handler used to log the underlying error before forwarding to `next(error)`:

\`\`\`ts
} catch (error) {
  console.error('Email verification failed:', error);
  next(error);
}
\`\`\`

That line was dropped during the rewrite. The current shape silently forwards:

\`\`\`ts
} catch (error) {
  next(error);
}
\`\`\`

Not a security issue (the prior log was not load-bearing), but reduces diagnosability if `verifyMagicToken` ever throws (e.g., DB hiccup on the `magic_tokens` table). Restoring the log keeps the error visible in pm2 + Sentry without changing user-facing behavior.

One-line fix.

wyf027:
## Summary
- restore the `verifyEmail` catch-block error log before forwarding to Express error handling
- add a regression test that asserts the underlying error is logged and passed to `next(error)`

Closes #2295

## Tests
- `src/controllers/UsersControllers.ts` and `src/controllers/UsersControllers.test.ts` parse successfully with the TypeScript parser
- source-level proof confirmed the controller contains `console.error('Email verification failed:', error)` and the regression test asserts that log + `next(error)` path

Not run: `pnpm test src/controllers/UsersControllers.test.ts` because both `git clone` and GitHub archive download for a full checkout were denied in this environment. The patch is intentionally small and includes the focused Jest regression test for project CI to run.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2340"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/src/controllers/UsersControllers.test.ts
+++ b/src/controllers/UsersControllers.test.ts
@@ -319,6 +319,28 @@ describe('UsersController.verifyEmail', () => {
 
     expect(res.redirect).toHaveBeenCalledWith('/login?verify_error=expired');
   });
+
+  it('logs verifyEmail failures before forwarding them', async () => {
+    const error = new Error('magic token lookup failed');
+    const verifyMagicToken = jest.fn().mockRejectedValue(error);
+    const { controller } = buildVerifyEmailController({ verifyMagicToken });
+    const req = { params: { token: 'boom-tok' }, cookies: {} } as unknown as express.Request;
+    const res = buildVerifyEmailRes();
+    const next = jest.fn();
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined);
+
+    try {
+      await controller.verifyEmail(req, res, next);
+    } finally {
+      consoleError.mockRestore();
+    }
+
+    expect(consoleError).toHaveBeenCalledWith('Email verification failed:', error);
+    expect(next).toHaveBeenCalledWith(error);
+    expect(res.redirect).not.toHaveBeenCalled();
+  });
 });
 
 describe('UsersController.requestMagicLink', () => {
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
Welcome, and thanks for the focused fix — small PR, clear repro test, matches the parallel `'Magic link verification failed:'` pattern at line 761. Nothing here blocks merge.

**Verdict**: approve (comment-only since I can't self-approve via tooling)

A missing log was making #2295 harder to debug; restoring it is a direct simplicity win.

### Blocking
- None.

### Nits (optional, not asks)
- `src/controllers/UsersControllers.test.ts` — the test asserts both calls happened but not the order. Since the regression is "log *before* forward", you can lock it in with `expect(consoleError.mock.invocationCallOrder[0]).toBeLessThan(next.mock.invocationCallOrder[0])`. The source has them on consecutive lines so it's implicit today, but the assertion would catch a future reorder.
- `src/controllers/UsersControllers.test.ts` — the `try/finally + mockRestore` is correct but slightly heavier than needed; `verifyEmail` catches its own errors, so a trailing `consoleError.mockRestore()` would do. Matter of taste.

When you're ready, flip from draft with `gh pr ready 2340` and I'll merge.

aalemayhu:
Superseeded by https://github.com/2anki/server/pull/2491

Issue #2343, opened 2026-05-17
App fails to load when localStorage contains stale/incompatible values


aalemayhu:
## Symptom

Site fails to load in a browser profile that has accumulated localStorage state from previous versions. Same browser, different profile (or incognito) loads fine. Reported once on Reddit (r/notion2anki, 2026-05-16) and reproduced internally by Alexander on 2026-05-16.

Workaround that works: clear site data for 2anki.net + hard refresh.

User-facing impact: blank screen with no error message — users assume the site is down.

## Suspected cause

~25 files in `web/src` read or write localStorage. The first-paint render path reads several of them. When a value's stored shape no longer matches what the consumer expects (field renamed, id no longer resolves, enum value removed), the consumer crashes before the UI can render an error boundary.

Recent shape-touching changes worth auditing first:

- Editor template state — #2337, #2334
- Auto Sync onboarding — #2324
- Analytics events buffer — #2320
- Card options

## Fix direction

Don't nuke localStorage on boot — that destroys legitimate user preferences. Instead:

1. Introduce a `safeReadLocalStorage<T>(key, validate)` helper that try/parse/validates and deletes + returns default on mismatch.
2. Route every boot-path localStorage read through it.
3. Wrap the app root in an error boundary that, on render failure, offers a one-click "Reset local data" recovery action instead of a blank screen.

Step 3 is the highest-leverage change — it converts every future "looks like the site is down" report into a self-serve recovery.

## Repro

Not yet pinned to a specific key. Next time it triggers, capture:

- The exact console error
- `Object.keys(localStorage)` and the value of any key the error references

That points at the corrupt key and the consumer that doesn't tolerate it.

aalemayhu:
The core fix requested in step 3 of this issue (an error boundary offering a one-click 'Reset local data' recovery action) was shipped in PR #2347 the day after this issue was filed. The `RootErrorBoundary` in `web/src/components/RootErrorBoundary/RootErrorBoundary.tsx` now catches any render crash and shows 'Something went wrong loading 2anki' with a 'Reset local data' button that clears localStorage and reloads — instead of a blank screen.\n\nThe optional `safeReadLocalStorage` wrapper (steps 1–2) was not built because no boot-path read does `JSON.parse` on localStorage; all reads use safe string comparisons or fall back to a default value.\n\nClosing as resolved.

PR #2347, opened 2026-05-17
fix(web): add local data recovery boundary

akmhatey-ai:
﻿## Summary
- adds a root-level recovery boundary for frontend render crashes during startup
- shows a self-serve reset UI instead of leaving users on a blank screen when stale browser state breaks rendering
- clears browser localStorage only after the user clicks "Reset local data", then reloads the app

Addresses #2343.

## Validation
- `pnpm --filter 2anki-web test:run src/components/LocalDataRecoveryBoundary/LocalDataRecoveryBoundary.test.tsx`
- `pnpm --filter 2anki-web typecheck`
- `pnpm --filter 2anki-web lint`
- `pnpm --filter 2anki-web build`
- `git diff --cached --check`
- `git diff --check origin/main...HEAD`
- `git diff --cached | gitleaks detect --pipe --redact --verbose --no-color`
- `git diff origin/main...HEAD | gitleaks detect --pipe --redact --verbose --no-color`

## Notes
- AI-assisted change.
- This PR implements the recovery boundary from the issue's fix direction. It does not add the broader `safeReadLocalStorage` helper or migrate every boot-path localStorage read, because the corrupt key/repro path is not pinned yet.
- `pnpm --filter 2anki-web build` still reports the existing `/bulma-switch.css` runtime-resolution warning and existing large chunk warnings.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2347"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews


aalemayhu (commented):
Welcome, and thanks for the first PR to 2anki — clean, well-tested, and aimed at a real pain point.

**Verdict**: approve (nits only)

A self-serve reset for a stuck startup screen is exactly the kind of "simpler, fewer support tickets" change that scales — green checks across SonarCloud, Snyk, and CI.

### Blocking
- None.

### Nits

**Copy** (`LocalDataRecoveryBoundary.tsx`)
- Subtitle reads a little hedgy and engineer-voiced. Per `VOICE.md` (Stripe/Linear register — specific, no apology, no support escape hatch), consider tightening:

  ```suggestion
            <p className={styles.subtitle}>
              Something saved in this browser is out of date. Reset it to
              recover.
            </p>
  ```

- Danger fallback can lose a few words too:

  ```suggestion
            <p className={styles.notificationDanger}>
              Couldn't reset local data. Clear site data for 2anki in your
              browser settings, then reload.
            </p>
  ```

**Engineering**
- `render()` leads with the negative — flip to match Sonar S7735 / repo convention:

  ```suggestion
    render() {
      if (this.state.error) {
        return (
          <main className={styles.pageNarrow}>
            { /* …fallback markup… */ }
          </main>
        );
      }
      return this.props.children;
    }
  ```

- `storage?: Pick<Storage, 'clear'>` is an unused injection point — tests use real `localStorage`. Per CLAUDE.md "don't design for hypothetical future requirements," consider dropping it; `reloadPage` injection alone covers testability.

- `Bugsnag.notify(error)` in `index.tsx` drops the component stack from `errorInfo`. Adding it as metadata makes the crash easier to triage:

  ```suggestion
          <LocalDataRecoveryBoundary
            onError={(error, errorInfo) =>
              Bugsnag.notify(error, (event) =>
                event.addMetadata('react', { componentStack: errorInfo.componentStack })
              )
            }
          >
  ```

- Nesting note (not a change request): the outer `<ErrorBoundary>` now only catches errors that escape this inner boundary's fallback render. Worth a glance to confirm that path still reports to Bugsnag too.

**Changelog**
- No `web/src/pages/WhatsNewPage/changelog.ts` entry. Per CLAUDE.md, user-visible fixes get an entry — this one is borderline (the screen is only seen during an already-broken state). Either add a line or note "no entry — recovery surface only reached when already broken" in the PR body so it's intentional.

aalemayhu:
One more — SonarCloud flagged two `readonly` smells on the class properties (low effort, low risk):

```suggestion
  private readonly reloadPage = () => {
    const reload = this.props.reloadPage ?? (() => globalThis.location.reload());
    reload();
  };

  private readonly resetLocalData = () => {
    try {
      const storage = this.props.storage ?? globalThis.localStorage;
      storage.clear();
    } catch {
      this.setState({ resetFailed: true });
      return;
    }

    this.reloadPage();
  };
```

Sonar findings:
- L34 `reloadPage` — never reassigned, mark as `readonly` (Major)
- L39 `resetLocalData` — never reassigned, mark as `readonly` (Major)

Worth folding into the same push as the other nits.

aalemayhu (changes requested):
See my comments for changes requested. Feel free to copy paste everything into your ai agent.

aalemayhu (commented):
**Verdict**: request-changes

Nice small fix toward "simplest, fastest" — turns a blank-screen crash into a one-click recovery. Welcome, first-time contributor!

### Blocking
- `web/src/pages/WhatsNewPage/changelog.ts:8` — per `VOICE.md`, changelog entries lead with the surface/noun, not a conditional. Adjacent entries follow this pattern ("Auto Sync — …"). Suggested replacement:

```suggestion
  { type: 'fix', title: 'Recovery screen — reset stale browser data when 2anki gets stuck loading', date: '2026-05-17' },
```

### Nits
- `web/src/components/LocalDataRecoveryBoundary/LocalDataRecoveryBoundary.tsx` subtitle "Something saved in this browser is out of date." is vague on the *what*. Optional tighter phrasing: `Stored data from a past version is breaking this page. Reset to recover.`
- `LocalDataRecoveryBoundary.tsx:32` — `reloadPage` recreates a closure on each call. Cleaner: extract `private readonly defaultReload = () => window.location.reload();` and `(this.props.reloadPage ?? this.defaultReload)()`.
- Style: `globalThis.location` / `globalThis.localStorage` work, but the rest of `web/src/` uses `window.location` / `window.localStorage`. Consistency only.

Security review: clean — no `error.message` rendered to the DOM, `localStorage.clear()` is gated behind an explicit user click, Bugsnag componentStack metadata is consistent with the rest of the app.

Engineering review: clean — branched off `main`, no layer/`any`/timer issues, `if (this.state.error)` leads with the positive, class component is the right call for an ErrorBoundary, test uses DI not internal mocks, SonarCloud + Snyk green.

aalemayhu (approved):
Thanks for the contribution, I will override the agent request and apply those suggestions.

PR #2524, opened 2026-05-21
Fix empty upload diagnostics for memory-backed files

### LINKED ISSUE #2497: Local-dev upload crashes with TypeError in GeneratePackagesUseCase.ts
A contributor running the server locally hit a crash on file upload:

```
TypeError: "path" argument must be of type string
    at ... GeneratePackagesUseCase.ts
```

Referenced in GitHub discussion #873. They wanted to contribute and were blocked at the front door.

## Why it matters

Local-dev path errors that take down the upload route are the single biggest reason contributors bounce. The contributor in question was motivated enough to file a discussion — many more will silently leave.

## What to verify

1. Reproduce: clone fresh, `pnpm install`, `pnpm dev`, upload a small `.html` file.
2. Locate the offending `fs.*` call in `src/usecases/jobs/GeneratePackagesUseCase.ts` — likely a path that is `undefined` only on first-run / empty-DB conditions.
3. Confirm whether this is environment-specific (Windows / WSL / fresh Postgres) or a real null path on the production path too.

## Definition of done

- A first-time contributor can upload an HTML file locally without modifying source.
- The crash, if not reproducible, has a clear note in `Documentation/` explaining the prerequisite that was missing (e.g. migration, env var).
- Closes the linked discussion.

Filed from inbox — original email on disk, contributor identity redacted.

resolvicomai:
## Summary
- read empty-deck diagnostics from `file.buffer` when multer has no disk path
- guard missing upload contents instead of passing `undefined` to `fs.readFileSync`
- add a regression test for memory-backed uploads that produce no packages

Closes #2497

## Tests
- `pnpm test -- src/services/UploadService.test.ts --runInBand`
- `pnpm typecheck`
- `pnpm exec jest --runInBand` attempted; fails locally because dependencies were installed with lifecycle scripts disabled, so `better-sqlite3` has no native binding, and parser/canary tests fail with `PythonExitError` in this environment.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2524"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781929386&installation_id=132681956&pr_number=2524&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F2524&signature=bdb5131422b4a3adde0dd0966a595587f535cc61cafa9a38bf297a44ca0fd6e1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/services/UploadService.test.ts
+++ b/src/services/UploadService.test.ts
@@ -165,6 +165,41 @@ describe('UploadService.handleUpload — error paths', () => {
     expect(body.message).not.toMatch(/<[a-z]/i);
   });
 
+  it('uses in-memory uploaded file contents for empty-deck diagnostics when no disk path exists', async () => {
+    const infoSpy = jest.spyOn(console, 'info').mockImplementation(() => undefined);
+    const errorSpy = jest.spyOn(console, 'error').mockImplementation(() => undefined);
+
+    try {
+      MockGeneratePackagesUseCase.mockImplementation(() => ({
+        execute: jest.fn().mockResolvedValue({ packages: [] }),
+      }) as unknown as InstanceType<typeof GeneratePackagesUseCase>);
+
+      const service = new UploadService(buildRepository(), {} as JobRepository);
+      const req = buildRequest({
+        files: [
+          {
+            fieldname: 'files',
+            originalname: 'memory-upload.html',
+            encoding: '7bit',
+            mimetype: 'text/html',
+            size: 36,
+            buffer: Buffer.from('<details><summary>Q</summary>A</details>'),
+          } as Express.Multer.File,
+        ],
+      });
+      const { res, capturedStatus } = buildResponse();
+
+      await service.handleUpload(req, res);
+
+      expect(capturedStatus()).toBe(400);
+      expect(errorSpy).not.toHaveBeenCalled();
+      expect(infoSpy).toHaveBeenCalledWith(expect.stringContaining('<details>'));
+    } finally {
+      infoSpy.mockRestore();
+      errorSpy.mockRestore();
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
Thanks for the contribution — and welcome! Tight, well-scoped fix with a real regression test, exactly the kind of patch that makes this codebase easier to maintain.

**Verdict**: approve (comment-only — Alexander will merge)

Closes a real diagnostic gap: when multer is in memory mode, `file.path` is undefined and the old `fs.readFileSync(undefined)` would throw inside the diagnostics block and mask the upstream "no packages" cause. Falls back to `file.buffer`, with a graceful skip when both are absent.

### Blocking
- None.

### Nits
- None.

Security pass: the buffer branch never touches the filesystem, so it removes (not adds) path-traversal surface; logging stays on `console.info`, not `res.send`. Engineering pass: type narrowing is right (both `path` and `buffer` are optional on multer's `File` per storage mode), branch is off main, test is colocated, and the `as unknown as InstanceType<…>` cast matches the convention already used elsewhere in this test file.

Ship-ready from my end.

PR #2619, opened 2026-05-22
feat: AnKing-style Image Occlusion Enhanced compatibility

aalemayhu:
## What
Adds a second note-type option for Image Occlusion decks: AnKing-compatible. Medical students using the AnKing deck can now import 2anki-generated IO cards directly into their existing collection — no field-mismatch warnings in Anki's import dialog.

## Why
Closes the AnKing compatibility gap. When a student imports a 2anki IO deck into an AnKing collection, Anki sees a different note-type shape, and cards land in a sibling deck the student never opens. This adds an opt-in toggle that emits the AnKing IOE 11-field shape instead. Moves IO repeat-usage rate (same user generates ≥2 IO decks in 30 days) toward the +20% target.

## How
- **New template**: `src/templates/n2a-io-anking.json` with the AnKing IOE 11-field shape (`ID`, `Header`, `Image`, `Footer`, `Remarks`, `Sources`, `Extra 1`, `Extra 2`, `Question Mask`, `Answer Mask`, `Original Mask`). SVG-overlay card template: front shows Question Mask overlaid on the image, back shows Answer Mask.
- **SVG converter**: `create_deck/helpers/io_shapes_to_svg.py` converts each OcclusionRect (rect/ellipse/polygon) into an SVG element. Labels are HTML-escaped (CWE-79). ViewBox is `0 0 1 1` (normalised coords).
- **Python bridge**: `create_io_deck.py` branches on `noteType` in `deck_info.json`. AnKing path emits one note per rect with per-rect SVG masks. Classic path is bit-for-bit unchanged. If `noteType` is missing from the JSON, defaults to `classic` so rollout is safe.
- **Wiring**: `noteType: 'classic' | 'anking'` plumbed from the IO upload page through controller → use case → Python. Controller parses from `RawData`, defaults to `anking` if unknown value; use case forwards into `deck_info.json`.
- **UI**: segmented control (Note type: 2anki classic / AnKing-compatible) added to the IO page panel header. Default is `classic` pending preview-side verification (per the trio conflict resolution). Helper text updates per selection.
- **Tests**: 23 Python tests for the SVG converter; 2 new TS tests asserting `noteType` is forwarded into `deck_info.json`.

## Measuring success
Log line / metric: `noteType=anking` in deck download events. Query: ratio of IO deck downloads with `noteType=anking` vs `classic` over time; repeat-usage rate (same user ≥2 IO decks in 30 days) measured at T+6 weeks vs pre-launch month baseline.

## Testing
- Unit tests added: 23 Python (shapes-to-SVG converter, all shape types + label escaping); 2 TS (noteType forwarding into deck_info.json)
- Pre-existing test failures (DeckParser.test.ts, runParserCanary.test.ts) are not introduced by this PR — confirmed present on main before these commits.

## Browser check
Notes: awaiting human browser check — UI adds a segmented control in the IO page panel header

## Changelog
"Image Occlusion — AnKing-compatible note type merges directly into AnKing decks"

## Risks
- AnKing's canonical model GUID vs our derived ID: if AnKing's add-on hooks require the exact IOE model GUID (`1374772155` from AnkiWeb), our stable-derived ID may not trigger the add-on's review hooks. This is the spec's riskiest assumption (open question 2). Rollback: user switches toggle back to classic.
- Preview: `ApkgPreviewService` may go blank for AnKing-shaped decks (SVG-overlay templates vs canvas templates). Existing fallback text applies; filed as follow-up.
- The toggle defaults to `classic` per the trio conflict resolution; flip to `anking` default in a follow-up after preview-side work lands.

## Goal alignment
Medical students are a core 300K-user lane. AnKing is the de-facto curated deck in that lane — cards that don't slot into AnKing don't get used. This makes 2anki IO output usable rather than parallel.

sonar-scanner: not run locally (SONAR_TOKEN not in env in this worktree)


aalemayhu:
Closing — approach is fundamentally broken on testing. Dropping the spec.

PR #2777, opened 2026-05-25
feat: redirect non-Notion users to /upload after login

aalemayhu:
## What
Users who sign in without a Notion connection now land on `/upload` instead of `/notion`. The `/notion` empty state led with a Notion connect prompt as the primary action — a dead end for users who came to convert PDFs, markdown, or zip exports.

## Why
Closes the friction gap for file-only users. Leading indicator: upload-to-download completion rate for the redirected cohort, measured for 7 days post-deploy.

## How
Added `landingForUser(req, userId)` to `UsersController` — a private helper that checks whether the user has a Notion token (one DB read against `notion_tokens`) and returns `/notion` for token holders, `/upload` for everyone else. An explicit `?redirect=` param always wins. Applied to `login`, `loginWithGoogle`, `loginWithMicrosoft`, and `loginWithApple`. Changed `getRedirect()` default fallback from `/notion` to `/upload`. The `loginWithNotion` callback stays hardcoded at `/notion` — the user just connected Notion, that destination is always correct.

## Measuring success
Server log line `POST /users/login → 200` with `redirect: /upload` (vs previous `/notion`) for users with no `notion_tokens` row. Query: count of users who hit `/upload` from post-login redirect and complete a deck download within the same session, compared to prior cohort's `/notion` completion rate.

## Testing
- Unit tests added: `UsersController.login — Notion-aware redirect` (3 cases: no token, has token, explicit redirect param)
- `getRedirect` test suite: updated 23 assertions from `/notion` fallback to `/upload`

## Browser check
Notes: Alexander to verify — sign in with Google or email+password on an account with no Notion connection, confirm landing page is `/upload`.

## Changelog
Sign in without Notion and land on the upload page immediately

## Risks
Users with a Notion token who land on `/notion` are unchanged. Users with no token who expected `/notion` now get `/upload` — the Notion nav link remains available. Rollback: revert the `landingForUser` change and the `getRedirect` default; one-line revert.

## Goal alignment
Removes the Notion auth wall for the majority of new signups who came from search for "PDF to Anki" / "markdown to Anki". First session completion is the primary conversion driver toward 300K users.

## Shared-file coordination
This PR does NOT touch `web/src/App.tsx`, `AppShell`, or any shared routing component. The only `web/src/` change is the new changelog JSON file. No merge-order conflict with sibling PRs #2770/#2771/#2773.

## Sonar
sonar-scanner not run locally (not installed). Change is 2–3 logic lines in a private method with no nesting depth or cognitive complexity concerns.


aalemayhu:
**Open-redirect hardening (resolves the 3 SonarCloud S5146 findings)**

The new login redirect sites tripped Sonar's open-redirect rule because `getRedirect` returned the user-supplied string after a boolean allowlist check — a validation Sonar's taint engine can't follow across the function boundary.

Rearchitected so **every** `getRedirect` return is a hardcoded value (allowlist literal, normalized constant origin, or `/upload`) — no user-derived string reaches `res.redirect`, so the taint flow is gone at the source rather than suppressed.

**Behavior delta (deliberate):** post-login deep links now land on the allowlisted section root instead of the exact path — `/templates/new` → `/templates`, and traversal attempts like `/notion/../../etc` → `/notion`. Absolute redirects resolve to the matching canonical origin. All other redirect behavior is unchanged.

103 tests pass (getRedirect + UsersControllers); 4 updated/added assertions cover the collapse behavior. sonar-scanner not run locally (not installed) — relying on the CI re-analysis to confirm the 3 findings clear.

PR #2853, opened 2026-05-27
fix: cover PDF image preview path in ApkgPreviewService tests

aalemayhu:
## What
Adds test coverage for the PDF-to-apkg image preview pipeline in `ApkgPreviewService`. Three new tests document the expected behaviour when an Anki deck is produced from a PDF conversion and the user opens the preview.

## Why
Images uploaded as part of PDF-converted decks were reported as not loading in the `/preview/apkg` view. Investigation traced the full pipeline from `convertPDFToImages` → `DeckParser.embedImagesInCardContent` → genanki manifest → `parseMediaManifest` → `rewriteMediaRefs`. The tests pin the contract at each boundary so any regression in the pipeline is caught immediately.

## How
Added a `describe('PDF-style image cards', ...)` block to the existing `ApkgPreviewService.test.ts`:

1. **Hash-named image in mediaMap** — verifies that `getCardsPage` rewrites `src` to the preview media URL (the happy path that `embedFile` produces).
2. **Original pdftoppm filename absent from mediaMap** — documents that `data-missing-media` is produced when `embedFile` fails and the card retains the raw filename. This is the failure mode that manifests as broken images.
3. **Original filename present in mediaMap** — verifies the rewrite works even when the manifest uses the raw filename, ensuring the service is not fragile to manifest naming variations.

All three tests use the existing `ParsedApkg` shape; no mocks introduced.

## Measuring success
`src/services/ApkgPreviewService/ApkgPreviewService.test.ts` passes with 33 total tests. Any future regression in the PDF image rewrite path will surface here with a failing test naming the specific case.

## Testing
- All 33 ApkgPreviewService tests pass
- Server tsc: clean
- Web typecheck: clean
- Web vitest: 1295 passed
- Web lint: no issues

## Browser check
Browser check: not applicable — the only `web/src/` file changed is the changelog JSON under `web/src/pages/WhatsNewPage/changelog/`, which is a bypass case per the attestation rules.

## Changelog
"Images load in the deck preview for PDFs converted to Anki"

## Risks
No production code changed — test and changelog only. No rollback needed.

## Goal alignment
Keeps the PDF → Anki conversion path reliable. PDF upload is a key acquisition surface; broken previews erode trust at the point where users evaluate their deck before downloading.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2853"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782515671&installation_id=132681956&pr_number=2853&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F2853&signature=7452e2c9d1647d5fccfa08c24809192db3a152c16a88b03bf1ccdf4d564e1619"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/services/ApkgPreviewService/ApkgPreviewService.test.ts
+++ b/src/services/ApkgPreviewService/ApkgPreviewService.test.ts
@@ -123,6 +123,86 @@ function makeBadOrdClozeCollection(): NormalizedCollection {
   };
 }
 
+function makePdfStyleCollection(hashName: string): NormalizedCollection {
+  const noteType = {
+    id: 1,
+    name: 'n2a-basic',
+    type: 0 as const,
+    css: '.card{}',
+    fields: [
+      { name: 'Front', ord: 0 },
+      { name: 'Back', ord: 1 },
+    ],
+    templates: [
+      {
+        name: 'Card 1',
+        ord: 0,
+        qfmt: '{{Front}}',
+        afmt: '{{Back}}',
+      },
+    ],
+  };
+
+  const note = {
+    id: 10,
+    mid: 1,
+    tags: '',
+    fields: [`<img src="${hashName}">`, 'some back text'],
+  };
+
+  return {
+    noteTypes: new Map([[1, noteType]]),
+    notes: new Map([[10, note]]),
+    decks: new Map([[2, { id: 2, name: 'PDF Deck' }]]),
+    cards: [{ id: 100, nid: 10, did: 2, ord: 0 }],
+  };
+}
+
+describe('PDF-style image cards', () => {
+  const service = new ApkgPreviewService();
+  const HASH_NAME = 'a1b2c3d4e5f67890abcdef1234567890abcdef12.png';
+  const ORIGINAL_NAME = 'notes.pdf-page1-1.png';
+  const BASE_URL = '/api/apkg/test.apkg/media/';
+
+  it('rewrites a hash-named image src to the preview media URL', () => {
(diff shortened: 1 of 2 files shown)


aalemayhu:
Closing — misdiagnosis. I fetched the reported apkg and ran the real code: the media manifest maps the filenames to numeric entries, the bytes are present, getMediaEntry returns them, and Content-Type resolves correctly. The server media path is fine. The actual cause is the preview iframe being sandboxed allow-scripts with no allow-same-origin, so card-HTML  requests reach the cookie-authenticated media endpoint from a null origin without credentials → 401 → broken images. This PR added only characterization tests + a changelog claiming a fix, with no source change. Superseded by #2856, which inlines media as data: URLs from the authenticated parent.

PR #2911, opened 2026-05-30
feat: admin-toggleable feature flags under /ops

aalemayhu:
## What
Adds a DB-backed feature_flag table, a cached runtime helper, and a Flags tab under `/ops/flags`. Replaces the env-var-then-restart pattern for runtime gating. Foundational for the AI-converter-floor canary on PR #2908 and every future flag-gated feature.

Closes the env-var rollout gap raised on PR #2908.

## Why
Today, flipping a runtime flag (`STRIPE_SYNC_ON_STARTUP`, the new `AI_CONVERTER_FLOOR_V1_ENABLED`, etc.) requires SSH + `.env` edit + pm2 cycle. Awkward for canary rollouts and incident response. A DB-backed flag with an admin toggle in `/ops` removes the bottleneck.

## How
- Migration `20260806000000_add_feature_flags.js` creates the table with `key` (PK), `value`, `description`, `updated_at`, `updated_by → users.id ON DELETE SET NULL`. Pre-seeds the `ai-converter-floor-v1` row in the same migration so PR #2908 has something to gate against on day one.
- `FeatureFlagsRepository` exposes `getAll()`, `get(key)`, and `set(key, value, userId)` — the join with `users` resolves the email so the UI can show "updated by <email>" without leaking the raw id.
- `getFeatureFlag(key, defaultValue)` in `src/lib/featureFlags/` reads through a 5-second in-process cache. `invalidateFeatureFlagCache(key)` fires after every write so the freshly-flipped value propagates within the same request.
- Two use cases — `ListFeatureFlagsUseCase` (read) and `SetFeatureFlagUseCase` (write + invalidate cache + emit `feature_flag_changed`).
- Two endpoints under `OpsRouter`, both gated by `RequireOpsAccess`:
  - `GET /api/ops/flags` — shape: `[{ key, value, description, updated_at, updated_by_email }]`.
  - `PUT /api/ops/flags/:key` — body `{ value: boolean }`. 400 on bad body, 404 when the key doesn't exist (no UI-driven flag creation by design — new flags ship via migration to keep the audit trail clean).
- Web: a new Flags tab in OpsLayout with optimistic toggles + rollback on error. Lazy-loaded from `App.tsx` to keep the route boundary in line with the other ops tabs.

## Measuring success
- Server-side: `feature_flag_changed` analytics events appear in the events table whenever a flag is flipped — confirms the audit path works.
- Server-side: log line `[ops] setFeatureFlag failed` will surface any post-write 500s in prod logs.
- Once #2908 lands and is wired to `getFeatureFlag('ai-converter-floor-v1', false)`, the first real-world check is flipping the flag from /ops and watching the canary metric move within ~5 seconds without a restart.

## Testing
- Unit tests added:
  - `FeatureFlagsRepository.test.ts` — SQL-shape assertions (PG dialect) for `getAll`, `get`, `set`.
  - `getFeatureFlag.test.ts` — cached read, TTL expiry, missing-key fallback, invalidation, default-on-known-missing.
  - `ListFeatureFlagsUseCase.test.ts`, `SetFeatureFlagUseCase.test.ts` — use-case behaviour + `feature_flag_changed` event emission + `FeatureFlagNotFoundError` path.
  - `OpsRouter.test.ts` — added 7 integration tests covering 404-non-owner, GET shape, PUT 400/404/200.
  - `FlagsTab.test.tsx` — initial load, toggle PUT, rollback on error, initial-load error banner.
- All new tests green locally. Pre-existing canary/parser failures on `main` were observed before this branch existed — not introduced here.

## Browser check
Notes:

## Changelog
No changelog entry — internal admin infrastructure, no user-visible behavior change.

## Risks
- New table goes live on next deploy. Migration is additive (CREATE TABLE + one INSERT); no production rows to backfill. Rollback is `DROP TABLE` if it ever lands wrong.
- The 5-second cache means a flipped flag takes up to 5s to propagate across all server processes. Acceptable for v1; tighten only if a real rollout demands it.
- `pnpm kanel` will need to run after merge to add `src/data_layer/public/FeatureFlags.ts` (the repository hand-rolls its own type so the PR compiles before kanel runs).
- `sonar-scanner` not run locally (no SONAR_TOKEN configured on this machine) — expect a Sonar bounce if the rule engine flags anything in the new component.

## Goal alignment
Internal infrastructure — enables faster, less risky rollouts of the AI converter floor and every future flag-gated growth experiment. Indirect path to 300K users: fewer restart-bound flag flips = faster iteration on conversion-funnel work.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2911"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782742076&installation_id=132681956&pr_number=2911&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F2911&signature=2affabed11cd0e806ac8475e092f4c10f02998cba285f08358b193bf0ba59d1d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/migrations/20260806000000_add_feature_flags.js
+++ b/migrations/20260806000000_add_feature_flags.js
@@ -1,0 +1,22 @@
+exports.up = async (knex) => {
+  await knex.schema.createTable('feature_flags', (t) => {
+    t.text('key').primary();
+    t.boolean('value').notNullable().defaultTo(false);
+    t.text('description');
+    t.timestamp('updated_at', { useTz: true }).defaultTo(knex.fn.now());
+    t.integer('updated_by').references('id').inTable('users').onDelete('SET NULL');
+  });
+
+  await knex('feature_flags').insert([
+    {
+      key: 'ai-converter-floor-v1',
+      value: false,
+      description:
+        'AI converter floor v1 — per #2726 spec. Off by default; flip from /ops to start the canary.',
+    },
+  ]);
+};
+
+exports.down = async (knex) => {
+  await knex.schema.dropTableIfExists('feature_flags');
+};
--- a/src/controllers/OpsController.ts
+++ b/src/controllers/OpsController.ts
@@ -14,6 +14,11 @@ import { IShowcaseRepository } from '../data_layer/ShowcaseRepository';
 import { GetMindmapImageStatsUseCase } from '../usecases/mindmaps/GetMindmapImageStatsUseCase';
 import { SyncStripeSubscriptionsUseCase } from '../usecases/ops/SyncStripeSubscriptionsUseCase';
 import { GetPricingAbFunnelUseCase } from '../usecases/ops/GetPricingAbFunnelUseCase';
+import { ListFeatureFlagsUseCase } from '../usecases/ops/ListFeatureFlagsUseCase';
+import {
+  FeatureFlagNotFoundError,
+  SetFeatureFlagUseCase,
+} from '../usecases/ops/SetFeatureFlagUseCase';
 
 class OpsController {
   constructor(
@@ -30,7 +35,9 @@ class OpsController {
     private readonly getMindmapStorageMetricsUseCase?: GetMindmapStorageMetricsUseCase,
     private readonly deleteInactiveUsersUseCase?: DeleteInactiveUsersUseCase,
     private readonly syncStripeSubscriptionsUseCase?: SyncStripeSubscriptionsUseCase,
-    private readonly getPricingAbFunnelUseCase?: GetPricingAbFunnelUseCase
+    private readonly getPricingAbFunnelUseCase?: GetPricingAbFunnelUseCase,
+    private readonly listFeatureFlagsUseCase?: ListFeatureFlagsUseCase,
+    private readonly setFeatureFlagUseCase?: SetFeatureFlagUseCase
   ) {}
 
   async getMetrics(req: express.Request, res: express.Response) {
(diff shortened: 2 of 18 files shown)


aalemayhu:
Closing per direction shift on the parent feature (#2908): we're moving the AI converter floor gate from server-side flag to per-user CardOption, so the first customer of this admin-flag mechanism evaporates.

The infrastructure itself is sound and will be useful when a genuine admin-only flag comes up (kill switch for a worker, staged rollout of a privacy-sensitive feature, etc.). Branch is preserved; when that flag arrives, this diff is recoverable via `git checkout feat/ops-feature-flags` and rebase. The pre-seeded `ai-converter-floor-v1` row should be dropped before any future re-open.

Not deleting the remote branch yet so the work stays one command away.

Issue #2513, opened 2026-05-20
chore: retire manual abandoned-checkout-recovery ops endpoint (1 week after webhook bake)


aalemayhu:
## Source
Spec PR for #2279 explicitly scheduled this:

> **Resolution**: Side with engineer. Webhook misfires during the first week are a bigger risk than dead-code rot. The cleanup is a separate, scheduled follow-up PR.
> **Resulting plan**: This spec PR; one implementation PR that adds the webhook handler + dedupe table + the 234-CSV backfill run; one follow-up PR a week after the webhook ships clean that deletes the manual endpoint.

## What

Delete \`POST /api/ops/send-abandoned-checkout-recovery\` (the manual endpoint shipped in #2270) one week after the webhook from #2279 has run cleanly in prod.

## Why

The endpoint exists as a break-glass tool during the rollout of the \`checkout.session.expired\` webhook. After a clean week of webhook firings (no signature failures, no retry storms, no missing-email payloads), the manual endpoint is dead code. CLAUDE.md rule: delete dead code, do not keep "in case."

## Acceptance

- One week of clean webhook firings observed via \`/ops/performance\`.
- Manual endpoint, its use case, and its \`RequireOpsAccess\` route registration are removed.
- 234-CSV backfill script in the ops folder is deleted (or moved out of the deployable surface).

PR #2946, opened 2026-05-31
spec: Reconcile from Stripe CSV (/ops command)

aalemayhu:
## What this covers

A one-page spec for a new **/ops** admin command: **Reconcile from Stripe CSV**. An admin uploads a Stripe **Subscriptions** CSV export; the tool matches each row to a 2anki user (by email, then Stripe customer id) and, for paying rows where the user has no active `subscriptions` row, provisions access through the existing write shape. **Provision-missing only — never revokes.** **Preview-first**: a dry-run report (totals + per-row outcome) precedes a checkbox-gated Apply, which re-parses the CSV server-side.

This generalizes the throwaway 234-row CSV backfill script that #2513 is set to delete — file #2513's cleanup separately; it is not blocked on this.

Spec: `Documentation/specs/stripe-csv-reconcile.md`

## Status

Draft. This is the spec PR; `/implement` will graduate this same branch to ready. The apply path (A vs B) is gated on a 2-minute inspection of a real Stripe Subscriptions CSV header (see the spec's riskiest-assumption section).

### Trio synthesis
- **PM**: Drive "paying-in-Stripe but free-in-our-DB" users to 0 for a known list in one upload + apply, with a per-row audit trail, replacing throwaway scripts and one-off SQL.
- **Designer**: A new card inside the existing `/ops` **Commands** tab (sibling to the Stripe sync card); preview-first with a checkbox-gated **Apply N changes**, masked PII in the report, full copy strings supplied.
- **Engineer**: Effort **M**, no migration, no Python; reuse the existing `subscriptions` upsert (`updateStoreSubscription`) — but the Stripe CSV lacks the nested objects that write expects, forcing an A-vs-B apply decision.
- **Agreement**: Admin-only; provision-missing only (never revoke); two endpoints (dry-run + apply); apply re-parses server-side; lives in the Commands tab; no schema change; build the dry-run report first.
- **Conflict**: How "apply" provisions. Engineer found the CSV can't feed `updateStoreSubscription` directly. **Resolved**: build the dry-run report first (identical for both options), then a real-CSV header check decides — **(B) CSV-as-index → fetch from Stripe → real write** if the export carries a subscription/product id (preferred), else **(A) synthetic write**. Apply ships as a fast-follow once decided.
- **Resulting plan**: Ship a preview-first `/ops` Commands card that reports who *would* be provisioned from a Stripe Subscriptions CSV; gate Apply on inspecting one real export to pick the write path, reusing the existing `subscriptions` provisioning shape and never revoking.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/2946"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782805463&installation_id=132681956&pr_number=2946&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F2946&signature=6498e2e7a49f842e9115b21afc6a8db7f8662e848c09f35739e5796a353d2a99"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/Documentation/specs/stripe-csv-reconcile.md
+++ b/Documentation/specs/stripe-csv-reconcile.md
@@ -1,0 +1,76 @@
+# Spec: Reconcile from Stripe CSV (/ops admin command)
+
+### Trio synthesis
+- **PM**: Drive "paying-in-Stripe but free-in-our-DB" users to 0 for a known list in one upload + apply, with a per-row audit trail, replacing throwaway scripts and one-off SQL.
+- **Designer**: A new card inside the existing `/ops` **Commands** tab (sibling to the Stripe sync card); preview-first with a checkbox-gated **Apply N changes**, masked PII in the report, full copy strings supplied.
+- **Engineer**: Effort **M**, no migration, no Python; reuse the existing `subscriptions` upsert (`updateStoreSubscription`) — but the Stripe CSV lacks the nested objects that write expects, forcing an A-vs-B apply decision.
+- **Agreement**: Admin-only; provision-missing only (never revoke); two endpoints (dry-run + apply); apply re-parses server-side; lives in the Commands tab; no schema change; build the dry-run report first.
+- **Conflict**: How "apply" provisions. Engineer found the CSV can't feed `updateStoreSubscription` directly. **Resolved**: build the dry-run report first (identical for both options), then a 2-minute inspection of a real Stripe Subscriptions CSV header decides — **(B) CSV-as-index → fetch from Stripe → real write** if the export carries a subscription/product id (preferred, keeps `payload` real); else **(A) synthetic write**. Apply ships as a fast-follow once decided.
+- **Resulting plan**: Ship a preview-first `/ops` Commands card that reports who *would* be provisioned from a Stripe Subscriptions CSV; gate Apply on inspecting one real export to pick the write path, reusing the existing `subscriptions` provisioning shape and never revoking.
+
+---
+
+## Outcome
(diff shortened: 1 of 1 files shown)


aalemayhu:
Closing — not needed. Checked the code: the gap this spec targets (paying users stuck on free because the webhook missed their event) is already closed. The Stripe webhook now handles customer.subscription.created/updated/deleted (WebhookRouter.ts:139 → provisionSubscription → updateStoreSubscription), STRIPE_SYNC_ON_STARTUP runs the full account sync on every deploy/restart (server.ts:258), and the /ops Sync now button does the same on demand. All three converge on the same provisioning write the CSV tool would have reused. The original 234-row CSV was a one-time backfill predating the created handler, not a recurring need. No user is reachable by CSV that these paths miss.

PR #3072, opened 2026-06-03
fix: tie native Notion connect to the app account via signed state

### LINKED ISSUE #2999: Notion native connect attaches token to the BROWSER's account, not the app's (state must carry the owner)
## Problem

Native Notion connect attaches the token to the **wrong 2anki account** — the one logged into 2anki **in the browser**, not the one logged into the **app**. Follow-up to #2995 / PR #2997 (that fixed the *redirect back*, not *user attribution*).

## Root cause

`src/controllers/NotionController.ts` `connect`:

```ts
await this.service.connectToNotion(authorizationCode, res.locals.owner);
if (stateStr === 'native') return res.redirect(NATIVE_NOTION_RETURN_URL);
```

`res.locals.owner` is resolved by `configureUserLocal` from **`req.cookies.token`** — and the callback request is the **browser's** request, so `owner` = the **browser's** 2anki session user. PR #2997 mints `state=native` (a constant) — it carries the redirect target but **not** the user.

The native app authenticates with its own JWT (captured in-app via the login webview); the **system browser** that runs the OAuth has a separate 2anki session. When they're different accounts:

- App calls `GET /api/notion/get-notion-link?client=native` with the **app's** cookie → `owner` = app user (call this `A`). `state=native`.
- User authorizes in the browser, signed into 2anki as user `B`.
- Notion → callback (browser request) → `res.locals.owner = B` → token stored on **B**.
- App's `getNotionLinkInfo(A)` → `isConnected: false`. App still shows "Connect."

Repro: sign the app into account A, sign the browser into account B, connect → app never sees the connection (token is on B). Confirmed in the field (app = "Thank You Professor", browser = "A. Alemayhu").

## Fix (server)

Carry the **app user** through `state` and have the callback tie the token to **that** user, not the callback cookie:

1. `getNotionLinkInfo(owner, { client: 'native' })` → mint `state = "native:" + signedToken(owner)` (a signed/opaque token encoding the *app's* owner — HMAC or a short-lived server-stored nonce → owner map). Not the raw id (avoid tampering).
2. `connect`: when `state` starts with `native:`, **verify + decode the owner from the state** and call `connectToNotion(authorizationCode, decodedOwner)` — ignore `res.locals.owner` for this branch. Then redirect to `NATIVE_NOTION_RETURN_URL`.
3. Keep the exact-match guard semantics (no open redirect; the redirect URL stays the server-held constant). The owner comes from a server-verified token, never raw client input.

This makes native connect work regardless of which account (if any) the browser is logged into — matching how the app already scopes `get-notion-link` by the app's cookie.

## Secondary observation

A tester also hit Notion's **"Integration appears to have timed out. Please refresh and try again."** on the consent screen (separate from the above — it's Notion-side, pre-callback). Likely transient/expired-authorize-session; flagging in case the authorize URL TTL or redirect_uri config needs a look.

## Client status

App already sends `?client=native` and catches `twoanki://x-callback-url/notion-connected`. No client change unblocks the attribution — it's the callback's `owner` source. Once the server reads the owner from `state`, the existing client flow works as-is.

_Filed from the native app repo._

aalemayhu:
## What
Native Notion OAuth attached the workspace token to the wrong 2anki account. The callback read `res.locals.owner` — the **browser** session — to attribute the token, but the native app and the system browser running the OAuth can be signed into different accounts. When they differ, the token landed on the browser's account and the app never saw the connection (`isConnected: false`, still showing "Connect").

This carries the **app's** owner through the OAuth `state` and ties the token to that owner at the callback.

## Why
Fixes #2999. After this, native connect works regardless of which account (if any) the browser is signed into — matching how `get-notion-link?client=native` already scopes by the app's cookie.

## How
- New pure module `src/services/NotionService/nativeOAuthState.ts`:
  - `buildNativeOAuthState(owner, secret)` → `native:<owner>:<ts>:<hmac>`, HMAC-SHA256 signed with `process.env.SECRET` (the same secret JWTs use).
  - `verifyNativeOAuthState(state, secret)` → timing-safe HMAC verify + 300s freshness window (mirrors the login-nonce window); returns the decoded owner or `null`.
  - `isNativeOAuthState(state)` recognizes both the signed form and the legacy bare `native`.
- `NotionService.getNotionLinkInfo(owner, { client: 'native' })` now mints the signed state at link time (the owner here is the app's authenticated user).
- `NotionController.connect` routes a native state to a `connectNative` helper that verifies the HMAC and uses the **decoded** owner for `connectToNotion`, never `res.locals.owner`.
- **Security:** `state` round-trips through the browser and Notion and is user-controllable, so it is HMAC-signed and freshness-bounded. A tampered/expired state — or a legacy bare `state=native` with no owner (old app build) — is **rejected** to `twoanki://x-callback-url/notion-connected?error=invalid_state` rather than falling back to the browser session. Rejecting the bare case is the safer option: attributing it to the browser session is the exact bug being fixed.
- Web flow (no `state`) and the `login:` branch are untouched. The deep link stays the single server-held constant; the client cannot pass a redirect target.
- No DB schema change (`notion_tokens.owner` is already the key).

## Measuring success
The native `get-notion-link?client=native` → connect → `getNotionLinkInfo(appOwner)` round-trip now returns `isConnected: true` for the app's account. In prod, the signal is the absence of the "connected but app shows Connect" reports for native users, and connect callbacks no longer attributing tokens to a mismatched browser session.

## Testing
- Unit tests added: `nativeOAuthState.test.ts` (round-trip, tampered owner, tampered signature, wrong secret, expired/in-window, empty secret, bare-native, non-positive owner, build-without-secret).
- Controller tests: bug repro (app owner A in state, browser owner B in `res.locals` → token tied to A), tampered state rejected without calling `connectToNotion`, bare `native` rejected.
- Updated `getNotionLinkInfo` native test to assert the link carries a signed state that verifies back to the owner.
- Server `tsc --noEmit` clean; 47 touched-suite tests green; web changelog loader test green.

## Browser check
Browser check: not applicable — backend-only OAuth state change (the only `web/src/` file is a changelog JSON, which the attestation hook bypasses)

## Changelog
`Connecting Notion from the app links it to the account you signed into the app with`

## Sonar
SonarCloud Automatic Analysis is enabled on this project, so a manual `sonar-scanner` run is rejected ("running manual analysis while Automatic Analysis is enabled"). Sonar will run automatically on this push. New code is flat (low cognitive complexity), no `any`, no nested ternaries; the `connect` extraction lowered the handler's complexity.

## Security review
This is an auth/external-integration change — please run `/security-review` before merge.

## Risks
- If `process.env.SECRET` were unset, `buildNativeOAuthState` throws at link-mint time (fail loud, not fail open). In prod `SECRET` is always set (`server.ts` refuses to boot without it), so this is safe; erroring beats minting a forgeable unsigned state.
- Old native app builds that send bare `state=native` will now get `?error=invalid_state` instead of a (wrong-account) success. This is intended — those builds had no way to attribute correctly. Rollback is a single revert of this branch.

## Goal alignment
Removes a silent dead-end in the native connect funnel — affected users believed they connected Notion but the app never saw it, so they couldn't convert. Fixing attribution unblocks those users and reduces a class of "it doesn't work" churn, supporting the 300K-user goal.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3072"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783110573&installation_id=132681956&pr_number=3072&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3072&signature=f07cb9715690cf4d971979ed65e92a08f3e6bf4d8c605349d54b9a7893f23f4d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


aalemayhu:
@claude review

PR #3078, opened 2026-06-03
fix: reject GOCSPX- OAuth secret in REACT_APP_GOOGLE_API_KEY

### LINKED ISSUE #2317: Hardening: refuse to boot if REACT_APP_GOOGLE_API_KEY has GOCSPX- prefix
## Context

On 2026-05-16 the Drive feature deployed with a misconfigured env: \`REACT_APP_GOOGLE_API_KEY\` was set to a value starting with \`GOCSPX-\`, which is the Google *OAuth Client Secret* format — not an API key (which starts with \`AIza\`).

Two bad consequences:
1. The Picker rejected the credential ("API developer key is invalid")
2. The Client Secret got inlined into the public JS bundle, briefly leaking it

## What to do

Add a boot-time / build-time check that refuses to start (or fails CI) if any \`REACT_APP_GOOGLE_API_KEY\` value starts with \`GOCSPX-\`. The check belongs in:
- The web build step (catches at CI before it ships)
- Server startup (catches at deploy time even when CI is bypassed)

Pattern: a single guard function, e.g. \`assertGoogleApiKeyShape(value)\`, that throws on \`GOCSPX-\` prefix and warns on missing \`AIza\` prefix.

## Why this is worth doing

The 30 minutes spent diagnosing the bad key value during the original Drive ship-week was preventable. A one-line check would have caught it at deploy.

aalemayhu:
## What

Adds a small `assertGoogleApiKeyShape` guard that the `useGooglePicker` hook calls when reading `REACT_APP_GOOGLE_API_KEY`. Bad shapes are refused with a single `console.error`:

- **`GOCSPX-` prefix** — Google OAuth Client Secret format pasted in by mistake. Refusing this value avoids leaking the secret into the public JS bundle the way the May 16 misconfiguration briefly did.
- **Missing `AIza` prefix** — does not look like a Google API key; Picker calls will fail.

When the key looks valid, behavior is unchanged.

## Why

PR #2306 (Google Drive upload) shipped briefly with the env var set to a `GOCSPX-` value. Two consequences:

1. Picker rejected the credential with an opaque "API developer key is invalid".
2. The Client Secret got inlined into the public bundle, briefly leaking it.

A one-line shape check at the read site catches that misconfiguration immediately in the browser console instead of after 30 minutes of diagnosis.

## Tests

- `src/lib/google/assertGoogleApiKeyShape.test.ts` — accepts `AIza...`, rejects `GOCSPX-...` and unrecognized, handles empty/whitespace.
- Existing `useGooglePicker.test.ts` and `UploadForm.test.tsx` updated to use a key with the correct `AIza` prefix (the previous `'test-key'` no longer satisfies the new guard).

Closes #2317

## Browser check

Browser check: not applicable — adds a startup-time env validation that only fires on misconfigured deploys; the configured (valid `AIza...`) path is exercised by existing Picker tests and unchanged for users.

- Scope: added the guard at the **client read site** (`useGooglePicker.ts`) only. Server-side validation was considered but the server does not read `REACT_APP_GOOGLE_API_KEY` — the variable is a Vite `REACT_APP_*` build-time injection consumed only by the web bundle, so guarding it at the read site catches it at first use and a server check would be checking an env it doesn't otherwise use. A separate CI check (`pnpm --filter 2anki-web build` failing on bad shape) would also work and could land later if you want the build to refuse rather than the browser.
- Behavior: returns `null` + logs `console.error` rather than throwing. Throwing would crash the upload page entirely; returning `null` mirrors the existing "no credential" path so the Drive chip disables cleanly while the error surfaces in dev tools.
- Suppression: the warning fires once per session via a module-scoped `warnedApiKeyShape` flag — re-clicking the chip won't spam the console.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3078"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783113784&installation_id=132681956&pr_number=3078&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3078&signature=1e8d3fb21a4aee29118fa206d41e64d9415219e34378ea311e579c2ace57db91"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/web/src/lib/google/assertGoogleApiKeyShape.test.ts
+++ b/web/src/lib/google/assertGoogleApiKeyShape.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from 'vitest';
+import {
+  assertGoogleApiKeyShape,
+  describeGoogleApiKeyProblem,
+} from './assertGoogleApiKeyShape';
+
+const FAKE_API_KEY_PREFIX = 'AIza' + 'FAKE_TEST_VALUE_NOT_REAL';
+const FAKE_OAUTH_SECRET = 'GOCSPX-' + 'fake-oauth-secret';
+
+describe('assertGoogleApiKeyShape', () => {
+  it('accepts an AIza-prefixed API key', () => {
+    const shape = assertGoogleApiKeyShape(FAKE_API_KEY_PREFIX);
+    expect(shape.valid).toBe(true);
+    if (shape.valid) {
+      expect(shape.key).toBe(FAKE_API_KEY_PREFIX);
+    }
+  });
+
+  it('rejects an OAuth Client Secret pasted in by mistake', () => {
+    const shape = assertGoogleApiKeyShape(FAKE_OAUTH_SECRET);
+    expect(shape.valid).toBe(false);
+    if (!shape.valid) {
+      expect(shape.reason).toBe('oauth-client-secret');
+      expect(describeGoogleApiKeyProblem(shape)).toContain('GOCSPX-');
+      expect(describeGoogleApiKeyProblem(shape)).toContain('public JS bundle');
+    }
+  });
+
+  it('rejects empty input', () => {
+    expect(assertGoogleApiKeyShape('')).toMatchObject({ valid: false, reason: 'empty' });
+    expect(assertGoogleApiKeyShape(undefined)).toMatchObject({ valid: false, reason: 'empty' });
+    expect(assertGoogleApiKeyShape(null)).toMatchObject({ valid: false, reason: 'empty' });
+    expect(assertGoogleApiKeyShape('   ')).toMatchObject({ valid: false, reason: 'empty' });
+  });
+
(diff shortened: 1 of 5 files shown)


aalemayhu:
@claude review

PR #3082, opened 2026-06-03
fix: rate-limit POST /upload/retry-with-credential by IP

### LINKED ISSUE #2412: Rate-limit POST /upload/retry-with-credential to prevent script-driven password attempts
## Why

Security-review finding I1 (informational) on [PR #2404](https://github.com/2anki/server/pull/2404): the attempt counter that surfaces the soft hint after 3 wrong passwords is client-side only. A script can POST unlimited wrong passwords against `POST /upload/retry-with-credential`.

The risk is limited because:
- The retry endpoint requires re-uploading the PDF on each call (multer enforces a file).
- The PDF library does the decryption work, so this self-limits on CPU rather than enabling credential stuffing against external systems.

Still worth closing — the spec's intent was a soft block, and a server-side rate limit makes that real.

## What

Add a per-IP (or per-session, if cheaper) rate limit on `POST /upload/retry-with-credential`. Suggested shape:

- 10 attempts per 15 minutes per IP, 429 once exhausted.
- Response: `{ error: 'too_many_attempts', retryAfterSeconds: <n> }`.
- Match whatever rate-limit infrastructure the codebase already uses (look for `express-rate-limit` or middleware in `routes/middleware/`); do not introduce a new dependency if one already exists.

## Done when

- Hitting the endpoint > N times within the window returns 429.
- The frontend handles the 429 with copy that matches VOICE.md (something like "Too many attempts — try again in a few minutes.").
- An integration test covers the rate-limit path.

## Out of scope

- Per-account / authenticated-user limits (anonymous-IP limit is sufficient as a first pass).
- CAPTCHA — overkill for the actual threat model.

aalemayhu:
## What

Adds a per-IP rate limiter in front of `POST /upload/retry-with-credential`. The handler now hashes the client IP, checks an in-memory `InMemoryRateLimiter` (10 requests / minute per IP, 500 global), and returns:

```
HTTP/1.1 429 Too Many Requests
{ "message": "Too many password attempts. Please wait a minute and try again." }
```

The temp file from `multer.dest = '/tmp'` is cleaned up on the 429 path so a script-driven attack can't fill the disk.

## Why

Security-review finding I1 (informational) on PR #2404: the attempt counter that surfaces the soft hint after 3 wrong passwords was client-side only — a script could POST unlimited wrong passwords against `/upload/retry-with-credential` without ever loading the UI. The risk was limited (the endpoint requires a PDF upload on each call and self-limits on CPU) but worth closing because the original spec's intent was a soft hint after 3 attempts.

## How

- New `src/lib/rateLimit/InMemoryRateLimiter.ts` — generalized version of the limiter in `ErrorEventController` (per-key window + global cap, injectable `now()`). Same algorithm, parameterized so different endpoints can pick different limits.
- `UploadController` constructs a default limiter at boot (10/min/IP, 500/min global) and accepts an injected limiter for tests.
- A failing limiter check returns 429 before the credential is read, before the PDF is decrypted, and after the multer temp file is unlinked.

## Tests

- `src/lib/rateLimit/InMemoryRateLimiter.test.ts` — covers per-key cap, key isolation, window reset, global cap. Uses an injected `now()` so the tests are deterministic.
- `src/controllers/Upload/UploadController.test.ts` — adds two tests: limiter rejects → 429; missing file → 400 short-circuit before the limiter is consulted.

## Browser check

Browser check: not applicable — server-side endpoint change, no `web/src/` files touched.

- Limits: 10/min/IP, 500/min global. Picked to be high enough for legitimate retries (a real user typing a password fast wouldn't hit it) and low enough that a script gets shut down quickly. Adjust if you want either bound tighter.
- Storage: in-memory, per-process. The existing `ErrorEventController` uses the same pattern; multi-process correctness is a separate concern that would need Redis if you want it later.
- Error copy: "Too many password attempts. Please wait a minute and try again." — direct per `VOICE.md`, no fake warmth. The client already classifies 429 → generic limit message; this PR doesn't depend on the client doing anything new.

Closes #2412

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3082"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783114401&installation_id=132681956&pr_number=3082&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3082&signature=7521c0172e395a4b5640c7b55c81dbd0f6cfea593381a7005d7f0149d97fb9b1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/controllers/Upload/UploadController.test.ts
+++ b/src/controllers/Upload/UploadController.test.ts
@@ -205,3 +205,132 @@ describe('Upload file — multer error handling', () => {
     );
   });
 });
+
+describe('UploadController.retryPdfWithCredential rate limit', () => {
+  test('returns 429 when the limiter rejects the IP', async () => {
+    const repository = {
+      deleteUpload: jest.fn(),
+      getUploadsByOwner: jest.fn().mockResolvedValue([]),
+      findByIdAndOwner: jest.fn().mockResolvedValue(null),
+      findByKey: jest.fn().mockResolvedValue(null),
+      findAllByObjectIdAndOwner: jest.fn().mockResolvedValue([]),
+      update: jest.fn().mockResolvedValue([]),
+      getLastUploadForUser: jest.fn().mockResolvedValue(null),
+      getLastReconvertibleUpload: jest.fn().mockResolvedValue(null),
+      findByOwnerAndDedupeKey: jest.fn().mockResolvedValue(null),
+      insertNativeDeck: jest.fn(),
+    };
+    const notionRepository: INotionRepository = {
+      getNotionData: jest.fn() as INotionRepository['getNotionData'],
+      saveNotionToken: jest.fn() as INotionRepository['saveNotionToken'],
+      getNotionToken: jest.fn() as INotionRepository['getNotionToken'],
+      deleteBlocksByOwner: jest.fn() as INotionRepository['deleteBlocksByOwner'],
+      deleteNotionData: jest.fn() as INotionRepository['deleteNotionData'],
+      markTokenInvalid: jest.fn().mockResolvedValue(undefined),
+      clearTokenInvalid: jest.fn().mockResolvedValue(undefined),
+      setReconnectEmailSent: jest.fn().mockResolvedValue(true),
+    };
+    const uploadService = new UploadService(repository, {} as JobRepository, buildUsersRepo());
+    const notionService = new NotionService(notionRepository);
+
+    const blockingLimiter = { check: jest.fn().mockReturnValue(false) };
+    const controller = new UploadController(
+      uploadService,
+      notionService,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      blockingLimiter
+    );
+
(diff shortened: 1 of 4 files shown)


aalemayhu:
@claude review

PR #3137, opened 2026-06-05
fix: capture scrubbed raw output for unknown upload errors

### LINKED ISSUE #2408: Capture raw output for the 'unknown' UploadErrorCode bucket
## Context

PR #2322's `UploadErrorCode` union ends in a catch-all `'unknown'`. Today, any error that isn't classified by `buildPythonExitError` (or any non-`PythonExitError` thrown through `ErrorHandler`) lands in this bucket and falls back to whatever raw `err.message` was attached.

To discover the next failure class worth promoting to its own code (and thus its own bespoke copy), we need to look at what the `'unknown'` bucket actually contains in production.

## What to do

- Sample the `'unknown'` bucket's `err.message` (or the stdout/stderr of the underlying `PythonExitError.rawOutput` where relevant) for a representative window — a week is fine.
- Bucket by message stem (first 80 chars, normalized).
- Pick the top-3 most frequent patterns.
- For each, decide whether it deserves a new `UploadErrorCode` value or whether the existing prose is good enough.

## Privacy

When capturing samples, **scrub any user-controlled content** (filenames, Notion page titles, deck names, copy-pasted snippets) before logging. Stick to the canned message prefixes from `buildPythonExitError` constants, the exception class name, and the Python crash kind.

## References

- PR #2322 (the shipping change)
- `src/lib/anki/buildPythonExitError.ts` (`PythonCrashKind` → `UploadErrorCode` mapping)
- `src/routes/middleware/ErrorHandler.tsx` (where `code = 'unknown'` is set)

aalemayhu:
## What
Persists a scrubbed, truncated sample of the Python converter's raw output (stderr/stdout) whenever an upload error lands in the `'unknown'` UploadErrorCode bucket. The sample goes into the existing `error_events.context` jsonb via `ErrorCaptureMiddleware` — no new sink, no new table.

## Why
Fixes #2408. The `'unknown'` bucket is where unclassified Python crashes land, and today only the generic `err.message` ("Something went wrong on our end… job ID …") is stored. `PythonExitError.rawOutput` already carried the diagnostic signal but nothing persisted it, so there was no way to find the next failure class worth promoting to its own error code and bespoke copy.

## How
- New pure helper `src/lib/anki/scrubPythonRawOutput.ts`:
  - `scrubPythonRawOutput` redacts quoted user filenames (`'My Deck.html'` → `[file]`), non-Python absolute paths (`/workspace/…` → `[path]`, keeping `.py` basenames so tracebacks stay readable), emails (`[email]`), and 32+ char token-like strings (`[token]`), then truncates to 500 chars.
  - `buildUnknownPythonErrorContext` returns `{ python_crash_kind, python_exit_code, python_raw_output }` only for `PythonExitError` with `kind === 'unknown'` and non-empty raw output; null otherwise.
- `ErrorCaptureMiddleware` sets `context` on the inserted `error_events` row using that helper. Classified kinds (`invalid-markup`, `too-large`, …) and plain Errors get `context: null` — no extra capture.
- The client response in `ErrorHandler` is untouched; raw output never leaves the server (CWE-532 / CWE-209 guidance per the issue and `.claude/rules/security.md`).

## Measuring success
Ops query: `SELECT context->>'python_raw_output', count(*) FROM error_events WHERE source = 'server' AND context ? 'python_raw_output' GROUP BY 1 ORDER BY 2 DESC;` — after a week, bucket by message stem and pick the top-3 patterns per the issue.

## Testing
- Unit tests added:
  - `scrubPythonRawOutput.test.ts` — redaction of paths/filenames/emails/tokens, `.py` basename preservation, 500-char truncation, context builder gating (unknown vs classified vs plain Error vs empty raw output).
  - `ErrorCaptureMiddleware.test.ts` — unknown PythonExitError → row carries scrubbed/truncated context; classified PythonExitError and plain Error → `context: null`.
  - `ErrorHandler.test.ts` — response body for an unknown PythonExitError is exactly `{ code, message }`, no raw output.

## Browser check
Browser check: not applicable — server-only change, no `web/src/` files in the diff.

## Changelog
No entry — internal-only observability, no user-visible change.

## Risks
Low. Capture is additive on the existing error path; the helper is pure and the middleware already fail-opens (fallback writer + `next(err)` on any capture failure). Scrubbing is conservative — if a redaction pattern over-matches, we lose diagnostic detail, never user privacy. Rollback: revert the single commit.

Sonar note: `sonar-scanner` not run locally — `SONAR_TOKEN` unset in this environment; change is small (one pure helper + one field on an insert), expect no findings.

## Goal alignment
Faster classification of conversion failures → better error copy → fewer abandoned uploads on the hot conversion path that drives the 300K-user goal.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3137"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783280648&installation_id=132681956&pr_number=3137&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3137&signature=3b2bab6f73963c07dd6652bd3f7f532197bb88e9f31d0f15a543b5208a05037f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/lib/anki/scrubPythonRawOutput.test.ts
+++ b/src/lib/anki/scrubPythonRawOutput.test.ts
@@ -1,0 +1,85 @@
+import {
+  buildUnknownPythonErrorContext,
+  scrubPythonRawOutput,
+} from './scrubPythonRawOutput';
+import { PythonExitError } from './buildPythonExitError';
+
+describe('scrubPythonRawOutput', () => {
+  it('keeps python file basenames from traceback paths', () => {
+    const raw =
+      'Traceback (most recent call last):\n  File "/app/create_deck/create_deck.py", line 12, in <module>';
+    expect(scrubPythonRawOutput(raw)).toBe(
+      'Traceback (most recent call last):\n  File "create_deck.py", line 12, in <module>'
+    );
+  });
+
+  it('redacts non-python absolute paths', () => {
+    const raw = "FileNotFoundError: /workspace/123/uploads/secret-notes.html";
+    expect(scrubPythonRawOutput(raw)).toBe('FileNotFoundError: [path]');
+  });
+
+  it('redacts quoted filenames containing user-named content', () => {
+    const raw =
+      "OSError: cannot open 'My Biology Deck Week 4.html' for reading";
+    expect(scrubPythonRawOutput(raw)).toBe(
+      'OSError: cannot open [file] for reading'
+    );
+  });
+
+  it('redacts email addresses', () => {
+    const raw = 'ValueError: bad value for student@example.com in field';
+    expect(scrubPythonRawOutput(raw)).toBe(
+      'ValueError: bad value for [email] in field'
+    );
+  });
+
+  it('redacts long token-like strings', () => {
+    const raw = `KeyError: secret_ntn_${'a'.repeat(40)}`;
+    expect(scrubPythonRawOutput(raw)).toBe('KeyError: [token]');
+  });
+
+  it('truncates output to 500 characters', () => {
+    const raw = 'KeyError: nested lookup failed\n'.repeat(100);
+    expect(scrubPythonRawOutput(raw)).toHaveLength(500);
+  });
+});
+
+describe('buildUnknownPythonErrorContext', () => {
+  it('returns context with scrubbed raw output for an unknown PythonExitError', () => {
+    const err = new PythonExitError('generic message', {
+      kind: 'unknown',
+      rawOutput: 'KeyError: /workspace/123/deck.html exploded',
+      code: 1,
+    });
+
+    expect(buildUnknownPythonErrorContext(err)).toEqual({
+      python_crash_kind: 'unknown',
+      python_exit_code: 1,
+      python_raw_output: 'KeyError: [path] exploded',
+    });
+  });
+
+  it('returns null for a classified PythonExitError', () => {
+    const err = new PythonExitError('markup message', {
+      kind: 'invalid-markup',
+      rawOutput: 'UserWarning: Field contained the following invalid HTML tags',
+      code: 1,
+    });
+
+    expect(buildUnknownPythonErrorContext(err)).toBeNull();
+  });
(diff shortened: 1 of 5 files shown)


aalemayhu:
SonarCloud gate was failing on two findings in the new `scrubPythonRawOutput.ts`:

1. **S5843** — quoted-filename regex complexity 21 (limit 20). Fixed by building the pattern from a `SCRUBBED_FILE_EXTENSIONS` list instead of one literal with `html?`/`jpe?g` shorthands; same match set, extension list now reads as data.
2. **S5852 security hotspot** (the actual gate failure — `new_security_hotspots_reviewed` 0% < 100%) — the email regex `[\w.+-]+@…` could backtrack super-linearly on long runs with no `@`. Fixed by bounding every quantifier (`{1,64}` local part, `{1,63}` labels, `{1,8}` labels), matching the bounded-repetition pattern Sonar already accepts on the neighbouring filename regex. The hotspot disappears on re-analysis rather than needing a manual review in the UI.

Branch also rebased onto current main (clean, no conflicts). Server tsc + the three touched suites (scrubPythonRawOutput, ErrorCaptureMiddleware, ErrorHandler — 36 tests) green.

PR #3136, opened 2026-06-05
chore: migrate lint/format from biome to oxlint + oxfmt

aalemayhu:
## What

Replaces the server's disabled-ESLint stubs and the web workspace's Biome with **Oxc** — `oxlint` for linting (both workspaces) and `oxfmt` for formatting.

- Adds `.oxlintrc.json` / `.oxfmtrc.json` (single shared config at root; web scripts point at `../`).
- Root scripts: `lint`, `lint:fix` (oxlint), `format`, `format:check` (oxfmt). Web `lint`/`lint:fix` swapped to Oxc.
- New **server-side** `lint` + `format:check` CI job (`server.yml`) — server code was previously unlinted.
- Removes `@biomejs/biome`, `web/biome.json`, and the dead root `.prettierrc`.
- Reformats `src/` and `web/src/` with oxfmt (the bulk of the diff).
- Updates the Biome references in `CLAUDE.md` and `web/CLAUDE.md`.

## Why

ESLint was removed months ago and the root `lint` scripts were `echo` stubs, so **server code shipped unlinted**. Web ran Biome. Consolidating on one fast Rust toolchain restores server linting and removes the two-tool divergence.

## Decisions / caveats (please review)

- **Lint rules carried over from Biome are set to `warn`, not `error`.** The never-linted server has **75 pre-existing violations** of the ported style rules (no-nested-ternary, no-negated-condition, prefer-optional-chain, etc.). Fixing them belongs in scoped follow-ups, not this tooling change. `pnpm lint` exits 0; the debt is visible as warnings. Ratchet to `error` as it is paid down.
- **oxfmt is pre-1.0 (0.53.0)** and **not fully idempotent in a single pass** — 17 files needed a second `pnpm format` to converge. The committed tree is converged (CI `format:check` is green), but contributors may need to run `pnpm format` twice. This is the experimental-formatter risk we accepted.
- Format ignores scope out HTML templates, kanel-generated, fixtures, and build output.

## Verification

- Server `tsc` + build: clean
- Web typecheck: clean
- Web tests: 1722 passed
- Server tests (sampled apkg/uploads/ankify): 111 passed
- `pnpm lint` exit 0, `pnpm format:check` exit 0

## Browser check
Browser check: not applicable — formatting-only reformat of web/src plus tooling/config; no component or runtime behavior changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

Issue #2427, opened 2026-05-18
test: verify worker pool spec acceptance criteria — load + trace


aalemayhu:
Follow-up from #2418 (Notion conversion worker pool).

## What

Two acceptance criteria from the original spec (`Documentation/specs/notion-conversion-worker-pool.md`, preserved in commit `2b51662f19fe`) were argued informally in #2418 but never asserted in tests:

- **AC #3** — 10 concurrent Notion conversions on a dev box keep healthcheck p99 < 100 ms.
- **AC #4** — `JobRepository.updateJobStatus`, `incrementCardUsage`, and `eventsQuery.countByNameForUser` writes happen in the same order as the pre-refactor main-loop path.

#2418 unit-tests the pool primitive itself but doesn't measure either claim.

## Why

The whole reason the refactor exists is AC #3. Without a measured baseline, we have no way to detect a regression if a future change re-blocks the main loop (e.g. someone adds a sync call inside the worker dispatch).

AC #4 is the contract that makes the refactor safe — argued by construction in #2418 (the same `performConversion` body runs, just in a worker), but argument-by-construction breaks the moment that body gets reshaped.

## Fix

- **Load test** — script that fires 10 concurrent `/api/notion/convert` requests against a local dev server and polls a cheap route (e.g. `/api/health` or `/api/upload/jobs`) every 100 ms. Assert p99 of the cheap route stays under 100 ms. Can live in `scripts/` and run on demand rather than CI.
- **Trace test** — integration test that spies on `JobRepository.updateJobStatus`, captures the write order, and asserts the sequence matches a baseline recorded against the pre-#2418 code path.

## Effort

M — load test is straightforward but needs a real DB. Trace test fits in `src/lib/storage/jobs/helpers/performConversion.test.ts`.

aalemayhu:
The spec this verifies (Documentation/specs/notion-conversion-worker-pool.md) doesn't exist in the repo or its history, and the load + trace validation it describes can't run until upload conversion is actually routed through the pool — which is #2425's job. Folding the acceptance testing into #2425's ship checklist and closing this one. Follow #2425 for the pool rollout.

PR #3141, opened 2026-06-05
feat: route upload conversion through the conversion pool

### LINKED ISSUE #2425: feat: route upload conversion through the conversion pool
Follow-up from #2418 (Notion conversion worker pool).

## What

`src/usecases/uploads/GeneratePackagesUseCase.ts` spawns a fresh `new Worker(workerPath, ...)` per upload request. Replace it with `runConversion`-style dispatch through the shared piscina pool added in #2418.

## Why

Cold-spawning a Node worker thread costs ~300 ms per upload. A warm pool reuses threads, eliminating that overhead on every conversion. The pool primitive already exists — this issue is the second customer.

## Why now and not in #2418

PM held the scope to Notion-only in #2418 so the brand-new pool primitive could shake out for a release before the upload path bets on it. Once #2418 has been on prod for ~a week with no pool-related issues, this is safe to ship.

## Fix

- Move the worker entry from `src/usecases/uploads/worker.ts` to a piscina-compatible export.
- Extend `runConversion` (or add a sibling `runUploadConversion`) on the existing pool.
- Migrate `GeneratePackagesUseCase.execute` to dispatch via the pool instead of `new Worker(...)`.
- Same boundary considerations as #2418: any class-instance arguments must be rehydrated inside the worker.

## Measuring success

Median upload time-to-first-byte. Today's baseline includes the 300 ms cold spawn on every request — expected reduction is ≥ 200 ms p50 for small-file uploads where the spawn dominates wall clock.

## Effort

M — one use case, the existing worker file, one new pool method. Tests already exist for `GeneratePackagesUseCase`.

aalemayhu:
## What
Routes upload conversion through the shared piscina conversion pool instead of cold-spawning a `new Worker(...)` per request. The upload worker entry becomes a pool-compatible named export (`uploadGeneration`) on the existing `conversionWorker`; `GeneratePackagesUseCase` dispatches via a new `runUploadGeneration()` on the pool.

## Why
Fixes #2425. Cold-spawning a worker thread costs ~300 ms per upload. The pool primitive (#2418) has served Notion conversions in prod since ~May 18 with no pool-related incidents — uploads are the planned second customer. Expected ≥ 200 ms p50 reduction for small-file uploads.

## How
- `src/usecases/uploads/uploadGenerationTypes.ts` — shared task/result types (type-only imports, so the main thread doesn't load the parser tree through the pool module).
- `src/usecases/uploads/worker.ts` — self-executing `workerData` block removed; logic exposed as `runUploadGenerationInWorker(task)`. Progress posts to a transferred `MessagePort`; errors return as a discriminated `{ ok: false, error }` result.
- `src/lib/conversionWorker.ts` — named export `uploadGeneration` next to the default Notion task (piscina `name` dispatch).
- `src/lib/conversionPool.ts` — `runUploadGeneration(task, transferList)`; pool gains `resourceLimits: { maxOldGenerationSizeMb: 1024 }`.
- `src/usecases/uploads/GeneratePackagesUseCase.ts` — builds a `MessageChannel` only when `onProgress` is given, transfers `port2`, rebuilds errors with the same `buildWorkerError` from #3134.

Error contract preserved exactly (#3134): `EmptyDeckError` rehydrated with `sourceFormat`, error `name` preserved, `PARSER_CRASH` coded fallback for blank messages. `UploadService` failure-reason routing (`resolveAsyncFailureReason`, sentinel regexes) untouched.

Rebased on main 2026-06-05: the worker now carries the AnkiApp export routing from #3145 and the oxfmt formatting from #3136 inside the pool path; conflicts resolved preserving both the pool dispatch and the AnkiApp branch in `processFile`.

1. **One shared pool, not a sibling upload pool.** The issue offered both; the shared pool with piscina's `name` dispatch is less code and one concurrency budget. Consequence: uploads and Notion conversions now share `CONVERSION_WORKERS` (default 4) threads — uploads queue behind Notion jobs under load, where previously each upload got its own unbounded thread. That bounding is the point of the issue, but it is a saturation-behavior change worth a prod-load look.
2. **`resourceLimits: { maxOldGenerationSizeMb: 1024 }` applied pool-wide.** The old upload spawn carried this cap; one pool can't set it per-task. This newly caps Notion conversion workers at 1 GB heap — previously uncapped, where a heap blowout aborted the entire process. Now it kills only the worker/job. If a legit Notion conversion needs > 1 GB this is a regression for that job; watch for `ERR_WORKER_OUT_OF_MEMORY` in prod logs.
3. **Error transport is a discriminated result, not a thrown error.** Piscina's structured-clone of thrown errors drops custom props (`sourceFormat`, `code`), so the worker returns `{ ok: false, error: { message, name, sourceFormat } }` and the use case rebuilds via the unchanged `buildWorkerError`. A genuine worker crash (OOM, termination) still rejects the pool promise and propagates raw — equivalent to the old `worker.on('error')` path, and strictly better than the old silent-hang when a worker exited without a message.
4. **Progress over a transferred `MessagePort`.** Only created when `onProgress` is passed (sync uploads don't pay for it). A progress message racing job completion can be dropped at `port1.close()` — harmless, since UploadService writes the terminal job status right after. Progress only fires on the Claude path (`generateDeckInfo`), so plain-HTML uploads see no change.
5. **Pool worker thread now loads the upload parser tree at boot.** `conversionWorker.ts` imports `PrepareDeck` and friends — slightly slower thread warm-up, irrelevant once warm.
6. **No timeout added.** The old path had none; adding one is out of scope for a behavior-preserving refactor.

## Load verification
Prod isn't reachable from this environment, so the draft-blocking "prod-load sanity" is answered with a committed concurrency test plus a real-pipeline smoke (acceptance criteria adapted from #2427):

- **`src/lib/conversionPool.load.test.ts`** — 10 concurrent upload conversions dispatched through the production path (`GeneratePackagesUseCase` → `runUploadGeneration` → real Piscina pool → real worker threads), with `CONVERSION_WORKERS=2` so 10 jobs queue onto 2 threads. Asserts: all 10 settle (none lost), `pool.completed ≥ 10` with `queueSize 0` (no deadlock, no stuck queue), each result maps back to the job that produced it (rejection messages name their own input file — no cross-job result swapping), per-job errors stay isolated (4 `EmptyDeckError` parses + 3 `.apkg` rejections complete alongside 3 clean successes), and `shutdownConversionPool` drains without force-destroy (no worker leak). Two jobs carry transferred progress `MessagePort`s to exercise the transfer list under concurrency. Runs in ~1.5 s; fixtures stop short of the Python packaging step so CI runners without a create_deck venv stay green.
- **Real-pipeline smoke (local, not committed)** — toggle HTML through the actual pool: 1 card, 69 850-byte `.apkg` written by a real worker thread + real Python; `.apkg` input rejected with the coded message through the same pool path; pool drained clean on shutdown.

The remaining open question from decision 1 (upload latency when Notion jobs saturate the pool on the prod box) is by construction not answerable locally — watch median `[upload] tempfile dwell` after deploy (see Measuring success).

## Measuring success
- Median `[upload] tempfile dwell` `dwellMs` in prod logs — today includes the ~300 ms spawn (file is read inside the worker after spawn); expected to drop toward queue-wait-only.
- No new `PARSER_CRASH` / empty-reason failed jobs (`jobs.last_error`) after deploy.
- Watch for `ERR_WORKER_OUT_OF_MEMORY` (decision 2).

## Testing
- Unit: `GeneratePackagesUseCase` rewritten at the pool boundary — dispatch via pool, progress forwarding + transfer list, EmptyDeckError/sourceFormat/name/PARSER_CRASH preservation, pool-rejection propagation (11 tests). `runUploadGenerationInWorker` — failure shape with non-empty message + name, progress-port close on both paths (3 tests). `runUploadGeneration` — pool dispatch under the `uploadGeneration` name with transfer list (2 tests).
- Load: `conversionPool.load.test.ts` — 10 concurrent jobs through a real 2-thread pool (6 tests, see Load verification).
- Full server suite after rebase: 442 suites / 3 921 tests green (incl. DeckParser with the main checkout's venv symlinked).
- Real-pipeline smoke after rebase (actual piscina pool, real worker thread, real Python): toggle HTML → 1 card, 69 850-byte `.apkg`; `.apkg` upload → correct coded rejection through the pool path. MessagePort transfer verified (a non-transferred port would throw DataCloneError at `pool.run`).
- `tsc -p . --noEmit` clean; web typecheck + vitest (1 752 tests) + oxlint green.
- Sonar: not run — `SONAR_TOKEN` unset in this environment; expect a possible Sonar pass on CI.

## Changelog
No entry — internal perf plumbing; user-visible speedup will be confirmed by prod telemetry first.

## Risks
- Shared concurrency budget (decision 1): heavy Notion traffic can delay uploads and vice versa. Rollback: revert this PR; the old spawn path is self-contained.
- 1 GB cap on Notion workers (decision 2): a >1 GB Notion conversion now fails as a job instead of taking the process down — net safer, but new failure mode.
- Piscina `TransferList` type mis-infers under `lib: ["dom"]`; one `as never` cast at the pool boundary with inline justification.

## Goal alignment
Faster time-to-deck on every upload — the core "drop something in, get a clean deck back" loop — plus bounded conversion concurrency, which protects the box as we scale toward 300K users.

### DIFF
--- a/src/lib/conversionPool.load.test.ts
+++ b/src/lib/conversionPool.load.test.ts
@@ -1,0 +1,180 @@
+import os from 'node:os';
+import path from 'node:path';
+import GeneratePackagesUseCase from '../usecases/uploads/GeneratePackagesUseCase';
+import { EmptyDeckError } from '../usecases/jobs/EmptyDeckError';
+import CardOption from './parser/Settings/CardOption';
+import Workspace from './parser/WorkSpace';
+import { UploadedFile } from './storage/types';
+import { getConversionPool, shutdownConversionPool } from './conversionPool';
+
+jest.setTimeout(60_000);
+
+const CONCURRENT_JOBS = 10;
+const POOL_THREADS = 2;
+
+function makeFile(name: string, contents: string): UploadedFile {
+  return {
+    fieldname: 'file',
+    originalname: name,
+    encoding: '7bit',
+    mimetype: 'application/octet-stream',
+    size: contents.length,
+    stream: null as never,
+    destination: '',
+    filename: name,
+    path: '',
+    buffer: Buffer.from(contents),
+    key: name,
+  };
+}
+
+function noCardHtml(index: number): string {
+  return `<html><head><title>empty-${index}</title></head><body></body></html>`;
+}
+
+interface JobSpec {
+  label: string;
+  file: UploadedFile;
+  expect: 'empty-deck-error' | 'empty-success' | 'apkg-rejection';
+  withProgressChannel: boolean;
+}
+
+function buildJobSpecs(): JobSpec[] {
+  const specs: JobSpec[] = [];
+  for (let i = 0; i < 4; i++) {
+    specs.push({
+      label: `no-cards-${i}.html`,
+      file: makeFile(`no-cards-${i}.html`, noCardHtml(i)),
+      expect: 'empty-deck-error',
+      withProgressChannel: i === 0,
+    });
+  }
+  for (let i = 0; i < 3; i++) {
+    specs.push({
+      label: `unsupported-${i}.xyz`,
+      file: makeFile(`unsupported-${i}.xyz`, `payload ${i}`),
+      expect: 'empty-success',
+      withProgressChannel: false,
+    });
+  }
+  for (let i = 0; i < 3; i++) {
+    specs.push({
+      label: `existing-deck-${i}.apkg`,
+      file: makeFile(`existing-deck-${i}.apkg`, `apkg bytes ${i}`),
+      expect: 'apkg-rejection',
+      withProgressChannel: i === 0,
+    });
+  }
+  return specs;
+}
+
+describe('conversion pool under concurrent upload load', () => {
+  const previousWorkers = process.env.CONVERSION_WORKERS;
+  const previousWorkspaceBase = process.env.WORKSPACE_BASE;
+  let errorSpy: jest.SpyInstance;
+  let results: PromiseSettledResult<{
+    packages: unknown[];
+    warnings?: string[];
+  }>[];
+  let specs: JobSpec[];
+
+  beforeAll(async () => {
+    process.env.CONVERSION_WORKERS = String(POOL_THREADS);
+    process.env.WORKSPACE_BASE = path.join(os.tmpdir(), 'pool-load-workspaces');
+    errorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+    specs = buildJobSpecs();
+    const useCase = new GeneratePackagesUseCase();
+    results = await Promise.allSettled(
+      specs.map((spec) =>
+        useCase.execute(
+          false,
+          [spec.file],
+          new CardOption({}),
+          new Workspace(true, 'fs'),
+          spec.withProgressChannel ? () => {} : undefined
+        )
+      )
+    );
+  });
+
+  afterAll(async () => {
+    await shutdownConversionPool({ timeoutMs: 15_000 });
+    expect(errorSpy).not.toHaveBeenCalledWith(
+      expect.stringContaining('did not drain')
+    );
+    errorSpy.mockRestore();
+    if (previousWorkers === undefined) {
+      delete process.env.CONVERSION_WORKERS;
+    } else {
+      process.env.CONVERSION_WORKERS = previousWorkers;
+    }
+    if (previousWorkspaceBase === undefined) {
+      delete process.env.WORKSPACE_BASE;
+    } else {
+      process.env.WORKSPACE_BASE = previousWorkspaceBase;
+    }
+  });
+
+  it('settles every one of the 10 concurrent jobs — none lost', () => {
+    expect(results).toHaveLength(CONCURRENT_JOBS);
+    for (const result of results) {
+      expect(['fulfilled', 'rejected']).toContain(result.status);
+    }
+  });
+
+  it('queues 10 jobs onto fewer threads and completes them all', () => {
+    const pool = getConversionPool();
+    expect(pool.options.maxThreads).toBe(POOL_THREADS);
+    expect(pool.queueSize).toBe(0);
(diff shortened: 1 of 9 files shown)


aalemayhu:
Rebased on main (10 PRs landed since the draft) and flipped ready.

**Conflicts resolved** (all in the upload worker path):
- `src/usecases/uploads/worker.ts` — main's AnkiApp export routing from #3145 auto-merged into `processFile`; the two manual hunks were the `doGenerationWork` signature and the worker tail, resolved to the pool shape (`runUploadGenerationInWorker` named export replacing the self-executing `workerData` block, which is exactly what this PR removes).
- `src/lib/conversionWorker.ts`, `src/usecases/uploads/GeneratePackagesUseCase.ts` + test — main-side changes were pure oxfmt reformatting from #3136, so the pool-dispatch side won; old spawn-path tests dropped in favor of the pool-boundary rewrite.
- #3146 truncation and #3147 heading-fields landed below the worker boundary (BlockHandler/parser) and merged clean.

**Load evidence** (replaces the prod-load sanity ask — prod isn't reachable from here):
- New committed test `src/lib/conversionPool.load.test.ts`: 10 concurrent conversions through the production dispatch path on a real Piscina pool capped at 2 threads. All 10 settle, `completed ≥ 10` / `queueSize 0`, per-job errors isolated (4 EmptyDeckError + 3 apkg rejections alongside 3 successes), rejection messages name their own input file (result integrity), pool drains clean with no force-destroy. ~1.5 s runtime, CI-safe (no Python needed).
- Real-pipeline smoke re-run after rebase: toggle HTML → 1 card, 69 850-byte `.apkg` via real worker thread + real Python; `.apkg` input rejected with the coded message; pool drained clean.
- Full server suite 442/3 921 green, web typecheck + 1 752 vitest + oxlint green, `tsc` clean. Sonar not run locally (`SONAR_TOKEN` unset) — flagged in the body.

What stays open for prod: upload queue-wait under Notion saturation (decision 1) — watch median `[upload] tempfile dwell` after deploy.

PR #3168, opened 2026-06-06
fix: isolate concurrent PDF-to-image conversions per call

### LINKED ISSUE #2862: pdfinfo race follow-up — choose between dedup at call site and per-call temp path once #2858 logs land
## Context

PR #2858 adds diagnostic logging to `PrepareDeck` to reveal whether concurrent `pdfinfo` failures are caused by:

1. Duplicate entries in `input.files` (same PDF name appears twice) — `Promise.all` fans out two `convertPDFToImages` calls against the same `workspace.location`
2. A retry overlap — the worker re-enters `convertPDFToImages` with the same workspace before the first invocation finishes

The spec for this work is preserved in git history: `git log -p -- Documentation/specs/pdfinfo-concurrent-runs.md` on the `fix/spec-pdfinfo-concurrent-runs` branch.

## What to do after the diagnostic lands

1. Re-run a long PDF upload (100+ pages) against staging.
2. Read the pm2 logs:
   - Two `[PrepareDeck] convertFile start` lines with the same `name` → case 1 is confirmed → dedup `input.files` inside `PrepareDeck` before the `Promise.all`.
   - One `convertFile start` line + two downstream `pdf→images` log entries → case 2 is confirmed → assign a random temp path inside `workspace.location` for `convertPDFToImages`'s write target, or classify the failure as terminal and stop the retry.
3. Ship the appropriate fix in a follow-up PR referencing this issue.

## Acceptance criteria (from the original spec)

- No new `Failed to execute pdfinfo` errors in pm2 logs for 14 days post-rollout against unencrypted PDFs that pdfinfo parses manually.
- Concurrent `convertPDFToImages` calls on the same workspace are safe — either there are no concurrent calls (dedup) or they no longer share `pdfPath` and `<basename>_stderr.log` (per-call temp path).

aalemayhu:
## What
Concurrent `convertPDFToImages` calls against the same workspace now write into a unique `pdf-<uuid>` subdirectory, so the PDF, its `pdfinfo` stdout/stderr logs, and its per-page images never share a path. `PrepareDeck` also dedupes `input.files` by name before the convert fan-out.

## Why
Fixes #2862. Two concurrent calls with the same PDF name — duplicate entries in `input.files` (the `Promise.all` fans out two conversions against one workspace), or a retry re-entering before the first finishes — all derived the same `pdfPath = path.join(workspace.location, fileName)`. They then raced on the file write, the `pdfinfo` read (reading a half-written / concurrently-overwritten file), the shared `<basename>_stdout.log` / `<basename>_stderr.log`, and the `${pdfPath}-pageN` image paths. The symptom users hit was an intermittent `Failed to execute pdfinfo` failure on PDFs that pdfinfo parses fine in isolation.

## How
- `convertPDFToImages.ts`: build `callDir = path.join(workspace.location, 'pdf-' + crypto.randomUUID())`, `mkdir` it, and put `pdfPath` inside. Because the dir is fresh per call, the file cannot pre-exist — dropped the `existsSync` guard and always write. `getPageCount` writes its logs into `dirname(pdfPath)` (now the unique dir) and `convertPage` writes images as `${pdfPath}-pageN` (also in the unique dir), so all four collision points are isolated by construction. Used `crypto.randomUUID()` (not `Math.random` — Sonar S2245).
- `combineIntoHTML.ts`: emit `` relative to `workspace.location` (e.g. `pdf-<uuid>/notes.pdf-page1-1.png`) so the page images still resolve through `embedFile` (`path.resolve(workspace.location, src)`). Falls back to basename when no base dir is passed, preserving the old contract for any other caller.
- `PrepareDeck.ts`: `dedupeFilesByName` keeps the first occurrence per name, applied once before the convert `Promise.all` and reused for `allFiles`. Cheap defense for the duplicate-entry case.

pdfinfo invocation, the page-limit check, and the error classification from #3148 are unchanged. No LLM / Vertex / Notion-export paths touched — no latency or token-cost delta.

## Measuring success
Per the issue acceptance criterion: no new `Failed to execute pdfinfo` errors in pm2 logs for unencrypted PDFs that pdfinfo parses manually. The diagnostic log lines from #2858 stay in place — `[PrepareDeck] received` now reports the deduped count, and a duplicate-named upload emits a single `[PrepareDeck] convertFile start` line instead of two. Grep pm2 for `pdfinfo_failed` against multi-PDF / retried jobs to confirm the race is gone.

## Testing
- `convertPDFToImages.test.ts` (new): two concurrent same-name calls get distinct pdf write paths and distinct `getPageCount` paths, each inside a unique workspace subdir; each call creates one `pdf-<uuid>` dir; emitted `img src` paths contain no `..` and resolve back to the on-disk page image under the workspace. Verified failing on the pre-fix source (shared `/tmp/.../anatomy.pdf`, no mkdir) before implementing.
- `PrepareDeck.test.ts` (updated): duplicate-named entries collapse to one received file and one conversion; distinct-named PDFs still convert separately. Replaces the two #2858 diagnostic-logging tests that asserted the now-fixed double-fan-out.
- `pnpm test src/lib/pdf src/infrastracture/adapters/fileConversion` → 11 suites / 75 tests green. Server `tsc --noEmit`, `pnpm format:check`, and `pnpm lint` clean (no warnings on changed files).
- `sonar-scanner` not run — scanner not installed and `SONAR_TOKEN` unset on this box. Change is small and adds no HTTP sinks or user-controlled-path writes; expect no new Sonar findings.
- Web vitest could not run in this worktree — a global `ERR_REQUIRE_ESM` from `html-encoding-sniffer` / `@exodus/bytes` in the shared `node_modules` breaks the jsdom environment for every web test, unrelated to this change. Web `tsc --noEmit` passes; the only web file here is a static changelog JSON.

## Browser check
Browser check: not applicable — server-side conversion fix; only web change is a changelog entry

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-06-06-pdf-concurrent-convert.json` — "Uploading the same PDF more than once in a batch converts reliably instead of intermittently failing"

## Risks
- Page images moved from the workspace root into a per-call subdir. Mitigated by emitting workspace-relative `img src` so `embedFile` `path.resolve(workspace.location, src)` still finds them inside the traversal-guarded base dir; covered by the new resolvability test.
- Dropping the `existsSync` guard means we always write — safe because the target dir is freshly created per call and cannot contain a prior file.
- Rollback: revert the single commit; no migration, no data, no config.

## Goal alignment
PDF-to-image is a core conversion path. Intermittent `pdfinfo` failures turn an upload into a dead end — the user gets nothing back. Removing a class of "drop something in, get an error" failures directly serves the simplest-fastest-deck mission and protects the conversion step of the funnel as we scale toward 300K users.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3168"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783340663&installation_id=132681956&pr_number=3168&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3168&signature=39158112cf26be8b3ccda6de3f650ae3c1f3bcae3d2d961ac1d3766ecad016ed"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
+++ b/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
@@ -1,4 +1,4 @@
-import { PrepareDeck } from './PrepareDeck';
+import { PrepareDeck, prepareDeckInfoOnly } from './PrepareDeck';
 import CardOption from '../../../lib/parser/Settings/CardOption';
 
 jest.mock('../../../lib/claude/ClaudeService', () => ({
@@ -398,14 +398,7 @@ describe('PrepareDeck — duplicate-name dedup', () => {
       workspace: makeWorkspace(),
     }).catch(() => undefined);
 
-    const startCalls = infoSpy.mock.calls.filter(
-      (args) => args[0] === '[PrepareDeck] convertFile start'
-    );
-    expect(startCalls).toHaveLength(1);
-    expect(startCalls[0][1]).toMatchObject({
-      name: 'anatomy.pdf',
-      workspaceLocation: '/tmp/test-workspace',
-    });
+    expect(convertPDFToImages).toHaveBeenCalledTimes(1);
   });
 
   it('keeps distinct-named PDFs as separate conversions', async () => {
@@ -421,13 +414,52 @@ describe('PrepareDeck — duplicate-name dedup', () => {
       workspace: makeWorkspace(),
     }).catch(() => undefined);
 
-    const startCalls = infoSpy.mock.calls.filter(
-      (args) => args[0] === '[PrepareDeck] convertFile start'
-    );
-    expect(startCalls).toHaveLength(2);
-    expect(startCalls.map((c) => c[1].name)).toEqual([
-      'anatomy.pdf',
-      'histology.pdf',
-    ]);
+    expect(convertPDFToImages).toHaveBeenCalledTimes(2);
+  });
+});
+
+describe('prepareDeckInfoOnly — duplicate-name dedup', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('converts a same-named PDF once instead of fanning out two conversions', async () => {
+    const settings = makeSettings();
+    await prepareDeckInfoOnly(
+      {
+        name: 'anatomy.pdf',
+        files: [
+          { name: 'anatomy.pdf', contents: Buffer.from('%PDF-1.4 a') },
+          { name: 'anatomy.pdf', contents: Buffer.from('%PDF-1.4 b') },
+        ],
+        settings,
+        noLimits: true,
+        workspace: makeWorkspace(),
+      },
+      makeWorkspace(),
+      makeWorkspace()
+    ).catch(() => undefined);
+
+    expect(convertPDFToImages).toHaveBeenCalledTimes(1);
+  });
+
+  it('keeps distinct-named PDFs as separate conversions', async () => {
+    const settings = makeSettings();
+    await prepareDeckInfoOnly(
+      {
+        name: 'export.zip',
+        files: [
+          { name: 'anatomy.pdf', contents: Buffer.from('%PDF-1.4 a') },
+          { name: 'histology.pdf', contents: Buffer.from('%PDF-1.4 b') },
+        ],
+        settings,
+        noLimits: true,
+        workspace: makeWorkspace(),
+      },
+      makeWorkspace(),
+      makeWorkspace()
+    ).catch(() => undefined);
+
+    expect(convertPDFToImages).toHaveBeenCalledTimes(2);
   });
 });
--- a/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
+++ b/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
@@ -350,7 +350,7 @@ describe('PrepareDeck — PDF text-vs-image gate', () => {
   });
 });
 
-describe('PrepareDeck — diagnostic logging', () => {
+describe('PrepareDeck — duplicate-name dedup', () => {
   let infoSpy: jest.SpyInstance;
 
   beforeEach(() => {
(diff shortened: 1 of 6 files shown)


aalemayhu:
@claude review

aalemayhu:
Addressed the review:

- **prepareDeckInfoOnly dedup gap (real fix):** \`prepareDeckInfoOnly\` now calls \`dedupeFilesByName(input.files)\` before its convert \`Promise.all\`, and \`allFiles = [...files, ...convertedFiles]\` uses the deduped list — matching the \`PrepareDeck\` path. A duplicate-named upload on that path no longer fans out two conversions against the shared pdf paths.
- **Test robustness:** the dedup tests now assert \`expect(convertPDFToImages).toHaveBeenCalledTimes(1)\` for the duplicate-name case (and \`2\` for distinct names) instead of counting \`console.info\` lines. Added a mirroring \`prepareDeckInfoOnly\` dedup suite (same-name → 1 conversion, distinct → 2). Wrote the same-name \`prepareDeckInfoOnly\` test first and watched it fail (2 calls) before applying the fix.
- **fs imports:** consolidated \`convertPDFToImages.ts\` to a single \`import { writeFile, mkdir } from 'fs/promises'\`; updated the test mock to match.
- **combineIntoHTML halving / Sonar:** left the halving logic untouched (out of scope). No Sonar token/scanner on this box, but the new diff adds no HTTP sinks, no user-controlled-path writes, and no new functions — it reuses the existing \`dedupeFilesByName\` and removes an import line. The two earlier Sonar notes were on \`combineIntoHTML\` / \`callDir\`, both unchanged here.

Local: \`pnpm test src/lib/pdf src/infrastracture/adapters/fileConversion\` → 11 suites / 77 tests green; \`tsc --noEmit\`, \`format:check\`, and \`lint\` clean on changed files. Not merging.

PR #3203, opened 2026-06-10
feat: add price lock-in commercial email package

aalemayhu:
## What
Ships the one-time **price lock-in** commercial email to free accounts ahead of the 15 June reprice ($6/mo → $7.99/mo, $60/yr → $64/yr). Free users can lock today's rate until Sun 21 June 23:59 CEST. Adds: a `price_lock_in_emails` table + repository (segment + dedupe + token resolution), an `EmailService.sendPriceLockInEmail` method, the `price-lock-in.html` template, a `SendPriceLockInEmailsUseCase` (dry-run + batched send + A/B variant split + `email_batch_sent` event), and an ops-gated `POST /api/ops/send-price-lock-in-emails` endpoint.

## Why
Convert free users to paid before the new rates apply, at the moment grandfathering is most compelling. Moves toward the 300K-user / revenue goal by lifting the second-deck → paid conversion the data already shows is the leak point. Reuses all existing SendGrid bulk machinery — no new email infrastructure.

## How
- **Segment** (`PriceLockInEmailRepository.getUsersToNotify` / `countUsersToNotify`): free accounts only — no active `subscriptions` row (matched by `email` or `linked_email`), no unexpired `user_passes` row, `patreon IS NOT TRUE`, `created_at` older than 14 days, not `marketing_opt_out`, deduped by email. Suppressed recipients are skipped at send time by `EmailService.deliver()` (same as every other commercial email), so they need no SQL filter. Date/now comparisons are bound as ISO strings so the query runs on both Postgres (prod) and sqlite (tests); the correlated `whereNotExists` subqueries also have a PG-dialect SQL-string assertion test per the raw-SQL testing rule.
- **Dedupe / no double-send**: every send is recorded in `price_lock_in_emails`; the segment query excludes any user with an existing row, so repeated batch invocations walk the full ~18.8k segment without re-emailing anyone.
- **A/B**: `SendPriceLockInEmailsUseCase` picks subject A vs B by user-id parity (deterministic 50/50), records the variant per send, and emits one `email_batch_sent` event `{ campaign: 'price_lock_in', count, variant_a, variant_b }`. `email_batch_sent` already exists in `AnalyticsEvents.ts` — that file was not touched.
- **Unsubscribe**: the existing unified resolver `ReEngagementRepository.findByToken` (already used by `/unsubscribe` and `/r/email`) now also resolves `price_lock_in_emails` tokens, so the commercial-required unsubscribe link and click tracking work with **no new routing**.
- **No cron / no scheduler / no auth changes / no Stripe or PricingPage changes.**

## Measuring success
- Send confirmation: the `email_batch_sent` event with `props.campaign = 'price_lock_in'` (count + variant split) — query the events table after each batch.
- Click-through: `email_clicked` events with `props.campaign = 'price_lock_in'` from `/r/email`.
- Conversion: upgrades attributable to the campaign window via the pricing funnel.
- Per-batch: the endpoint returns `{ count, dryRun, variantA, variantB }`.

## Testing
- Unit tests added:
  - `PriceLockInEmailRepository.test.ts` — segment in/out for free vs paying (email + linked_email) vs active-pass vs expired-pass vs lifetime vs opted-out vs <14d vs already-sent; `countUsersToNotify`; `findByToken`; plus a PG-dialect generated-SQL assertion for the four exclusion subqueries and the dedupe-by-email selection.
  - `SendPriceLockInEmailsUseCase.test.ts` — dry-run sends nothing + returns count; sends + records each send; no double-send across three runs; variant split by parity; `email_batch_sent` shape; records the user even when SendGrid throws.
  - `PriceLockInEmail.test.ts` — template structure (mascot, footer tagline, CTA, unsubscribe), subject A/B, CTA URL, working unsubscribe link, body copy + plain-text part. SendGrid mocked at the module edge.
  - `ReEngagementRepository.test.ts` — added price-lock-in token resolution + schema.
- Full server suite: 4094 passing. The only failing suites (`DeckParser`, `runParserCanary`) are the Python `create_deck` pipeline failing because this worktree has no venv — pre-existing and environmental, unrelated to this change.

## Browser check
Browser check: not applicable — server-only email package (no `web/src` changes).

## Changelog
No entry — the email is not an in-app surface; nothing changes in the product UI.

## Sonar
`sonar-scanner` is installed but `SONAR_TOKEN` is unset in this environment, so the local scan was not run. Expect the CI Sonar pass to be the first rule-engine check on this diff.

## Send-day runbook (manual, ops-gated)
**Send is manual via `POST /api/ops/send-price-lock-in-emails` after Alexander's explicit go on Monday ~09:00 CEST. Dry-run first to verify the segment count (~18.8k expected).**
1. **Dry-run**: `POST /api/ops/send-price-lock-in-emails` body `{ "dryRun": true }` → returns `{ count, dryRun: true }`. Confirm `count` is in the expected ~18.8k range.
2. **Confirm** the count with Alexander.
3. **Real run, batched**: `POST` body `{ "dryRun": false, "limit": 500 }`, repeated. Each call sends up to 500, records sends, and the segment query excludes already-sent users — so re-invoking walks the remaining segment with no double-send. Continue until a run returns `count: 0`.
4. **Verify**: query the events table for `email_batch_sent` with `props.campaign = 'price_lock_in'`; sum `count` and the `variant_a` / `variant_b` split across batches.

## Risks
- The `price_lock_in_emails` kanel type is **hand-written** (kanel is not installed in this worktree); rerun `pnpm kanel` after pulling to regenerate it identically. The migration runs automatically on startup.
- Migration adds a new table (no lock on existing tables, no backfill); rollback is `down` → drop table.
- Adding `sendPriceLockInEmail` to `IEmailService` required updating 8 existing inline mocks — covered by tsc + the suites above.

## Goal alignment
Targets the free → paid conversion at the highest-intent moment (grandfathering a rate that is about to retire). Directly serves the 300K-user / revenue goal by converting the existing free base rather than relying on new acquisition.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3203"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783688724&installation_id=132681956&pr_number=3203&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3203&signature=43536cd768cdc17fc78d151e5d222820bf398e3f75b4a0a588a7b3cd04f91b7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


aalemayhu:
Addressed the migration-review blockers (commit 148ea81):

**Blockers**
1. Renamed `20260610000000_price_lock_in_emails.js` → `20260610000001_...` — the timestamp collided with the already-merged `20260610000000_deck_shares.js`. Two identical timestamps order non-deterministically in knex, which can break fresh installs.
2. Added `.onDelete('CASCADE')` to the `user_id` FK, matching the `deck_shares` / `user_delete_fk_cascade` convention. Without it, user deletion fails on an FK violation once rows exist.

**Nits**
3. Named the index explicitly: `price_lock_in_emails_user_id_idx`.
4. Verified up + rollback against a local Postgres. UP creates all 5 columns with original types/nullability; FK `delete_rule = CASCADE`; index lands as `price_lock_in_emails_user_id_idx`; DOWN drops the table cleanly. Column shape is unchanged, so the hand-written kanel type is shape-identical — kanel regeneration deferred to the post-merge deploy.

Affected tests green: `PriceLockInEmailRepository`, `SendPriceLockInEmailsUseCase`, `PriceLockInEmail` (EmailService) — 29 passed.

PR #3204, opened 2026-06-10
feat: pricing v2 core (grandfather window, flag-gated)

aalemayhu:
## What
Introduces pricing v2 for Unlimited ($7.99/mo, $64/yr) behind the global `pricing_v2` feature flag, with a grandfather window that serves legacy ($6/$60) pricing to accounts created before a code-constant cutover until the lock-in deadline (`2026-06-21T21:59:00Z`). Flag defaults **off** — behavior is unchanged for everyone until flipped.

## Why
Raise Unlimited revenue per subscriber while protecting existing users from a surprise increase, and steer new signups toward the annual plan (the second-deck retention lever from the growth analysis). Prices are PM-recommended and **pending Alexander's sign-off** ($7.99/$64).

## How
- **Cohort logic** (`src/usecases/checkout/pricingV2.ts`) — pure `resolveCohort` / `qualifiesForLegacyWindow`. Cutover (`2026-06-15T07:00:00Z` placeholder) and window-end are code constants, not env flags.
- **Price resolution** (`src/services/StripePriceResolver.ts`) — resolves v2 prices by `lookup_key`, caches the ID, falls back (logged) when resolution fails.
- **Checkout** — `UnlimitedCheckoutUseCase` resolves the correct price per cohort, stamps `cohort` into Stripe metadata, never trusts client price/interval beyond the existing `month|year` validation. Falls back to legacy env IDs on resolution failure.
- **Cohort endpoint** — `GET /api/checkout/prices` (optional auth) returns cohort-correct cents + `lockInDeadline`. The page renders legacy numbers for legacy-window users.
- **Web** — annual preselected; Unlimited card leads with per-month hero price above the toggle, cycle-dependent terms line, checkout error state, single primary CTA; `LockInBanner` for logged-in free users inside the window.
- **Stripe script** — `scripts/pricing-v2.ts` (idempotent, lookup_key-first, Step-0 catalog print, `--live`+confirm for live, reuses existing Unlimited product).

## Step 0 inventory (current state)
- **Env price vars:** `UNLIMITED_MONTHLY_PRICE_ID` / `UNLIMITED_YEARLY_PRICE_ID` were referenced by `CheckoutRouter` but missing from `src/env.example` — added them with a note. Legacy IDs stay the fallback; v2 prices resolve at runtime by lookup_key.
- **Checkout flow:** `CheckoutRouter` → `UnlimitedCheckoutController` → `UnlimitedCheckoutUseCase` → Stripe `checkout.sessions.create`. Unchanged shape; now threads flag + `created_at`.
- **Webhook entitlements:** unchanged — `StripeController` handles `customer.subscription.*`; v2 prices attach to the same Unlimited product so existing entitlement matching keeps working.
- **Email capability:** untouched (parallel PR owns templates).
- **Analytics:** added `checkout_started` (server), `plan_interval_selected`, `lock_in_banner_shown`, `lock_in_banner_clicked` (client). `checkout_completed`/`paywall_shown` reused.

## Measuring success
- `checkout_started` events grouped by `props.cohort` (legacy vs v2) confirm the grandfather split is firing in prod.
- Server log `unlimited.checkout.session_created` carries `cohort`; `pricing.resolve.failed` / `unlimited.checkout.v2_resolve_fallback` flag any lookup_key resolution misses.
- `GET /api/checkout/prices` response per cohort is the single source the page renders from.

## Testing
- Server unit tests: cohort boundaries (pre-cutover in/after window, post-cutover, flag off), price resolver (resolve/cache/miss/error), use-case v2-resolution + legacy fallback + flag-off, controller flag/createdAt threading + `checkout_started` emit, prices endpoint per cohort. 120 server tests green across touched paths; full server `tsc` clean.
- Web: PricingPage (50), LockInBanner (7), pricing formatters (5) — all green; web `typecheck` + `lint` clean (no new warnings in touched files).
- Playwright `web/tests/checkout.spec.ts` (mock-based): annual-default + correct hero price + checkout starts with `interval: year`.
- **Stripe script rehearsed in test mode** — created `v2_monthly` ($7.99/mo) and `v2_annual` ($64/yr); idempotent re-run verified (both skipped). Live run still gated on price sign-off.
- **Sonar not run locally** — `sonar-scanner` is installed but `SONAR_TOKEN` is unset here. Expect the CI Sonar pass to be the first rule-engine run.

## Browser check
Notes: Verified with Playwright Chromium against `pnpm dev` on the PR branch, desktop + 375px. Flag-off state confirmed: legacy $6 pricing renders, no v2 numbers leak, Get Unlimited CTA and the new cancel-terms line present. The only console message is a 401 from the unauthenticated session probe, reproduced identically on production /pricing — pre-existing, not introduced here.

## Changelog
`2026-06-15-annual-default.json` — "Unlimited now leads with the annual plan — two months free over paying monthly"

## Risks
- The Unlimited card layout (price-above-toggle, terms line, single CTA) renders for **all** users once merged, not only flag-on cohorts — this is the designer-approved redesign ("ship it"). Only the price **numbers** and the lock-in banner are flag/cohort-gated, so "flag off = byte-identical" holds for the numbers and banner but not the card structure. Called out so it's a conscious decision, not a silent deviation.
- If v2 lookup_keys don't exist in Stripe when the flag flips, checkout falls back to legacy env IDs (logged) — no user-facing failure, but v2 pricing silently won't apply until the script runs.

### Rollback
1. Flip `pricing_v2` off in `feature_flags` (Ops panel). 
2. Verify: `GET /api/checkout/prices` returns `cohort: legacy`, `$6/$60`, `lockInDeadline: null` for every caller; the LockInBanner disappears; new `checkout_started` events show `cohort: legacy`. No Stripe objects are mutated by the flag, so rollback is instantaneous.

## Gated on sign-off (do NOT do until prices are approved)
- Live-mode `scripts/pricing-v2.ts` run (`--live`).
- `pricing_v2` flag flip.
- Finalizing the cutover timestamp constant (currently `2026-06-15T07:00:00Z` placeholder).

## Goal alignment
Annual-default + grandfather lock-in targets the retention/revenue leak named in the growth analysis (second-deck conversion, monthly→annual mix) while keeping the free 100-cards tier untouched — supports the 300K-user + sustainable-revenue goal without raising the signup barrier.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3204"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783689009&installation_id=132681956&pr_number=3204&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3204&signature=2e7b94ef2ca7116dcb16468bce39b88a9725421b81dd8ff09decbe852d23a0e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.github/workflows/checkout-smoke.yml
+++ b/.github/workflows/checkout-smoke.yml
@@ -1,0 +1,41 @@
+name: Checkout smoke
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  checkout-smoke:
+    name: checkout smoke (22.20.0)
+    runs-on: blacksmith
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+      - name: Install Playwright Browsers
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+        run: pnpm --filter 2anki-web exec playwright install --with-deps chromium
+      - name: Build the application
+        run: pnpm --filter 2anki-web build
+      - name: Run checkout spec
+        working-directory: web
+        run: pnpm exec playwright test checkout.spec.ts
+        env:
+          PLAYWRIGHT_WITH_MOCK: true
+      - uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: checkout-smoke-report
+          path: web/playwright-report/
+          retention-days: 7
--- a/Documentation/specs/pricing-v2.md
+++ b/Documentation/specs/pricing-v2.md
@@ -1,0 +1,38 @@
+# Pricing v2 core
+
+## Goal
+
+Introduce new Unlimited price points behind a global `pricing_v2` feature flag, with a grandfather window so existing accounts keep legacy pricing through a lock-in deadline.
+
+## Price points (pending sign-off)
+
+- v2 monthly: $7.99 (lookup_key `v2_monthly`)
+- v2 annual: $64, $5.33/mo, save 33% (lookup_key `v2_annual`)
+- Legacy monthly: $6 — Legacy annual: $60
+
+## Grandfather rules
+
+- Cutover constant (in code, not env): `2026-06-15T07:00:00Z`.
+- Lock-in window end: `2026-06-21T21:59:00Z` (2026-06-21 23:59 CEST).
+- An account sees legacy pricing when the flag is on AND `created_at < cutover` AND `now < window end`. Otherwise v2.
+- Flag off → legacy pricing for everyone (today's behavior, byte-identical numbers).
+
+## Workstreams
+
+1. `scripts/pricing-v2.ts` — idempotent Stripe price creation (lookup_key lookup first), test/live mode detection from key prefix, `--live` + interactive confirm for live, Step 0 catalog print, reuse existing Unlimited product.
+2. Server: cohort resolution (`pricingV2.ts`), `StripePriceResolver` (cache + fallback), `UnlimitedCheckoutUseCase` resolves v2 by lookup_key with legacy env fallback, `GET /api/checkout/prices` returns cohort-correct numbers + lock-in deadline.
+3. Web: annual default, UnlimitedCard price-above-toggle, terms line, error state, one primary CTA, LockInBanner.
+4. Analytics: `checkout_started` (server), `plan_interval_selected`, `lock_in_banner_shown`, `lock_in_banner_clicked` (client).
+5. Playwright E2E + 6h scheduled checkout smoke.
+
+## Acceptance criteria
+
+- Post-cutover account sees $7.99/$64; pre-cutover sees $6/$60 through window then v2.
+- Annual preselected on load.
+- `pricing_v2` off → today's behavior for everyone.
+- Lock-in banner only for logged-in free users inside the window.
+
+## Out of scope
+
+- Live Stripe price creation, flag flip, cutover-timestamp finalization — gated on price sign-off.
+- Per-tier caps, replace-mode pricing, currency localization beyond Stripe's checkout conversion.
--- a/scripts/pricing-v2.ts
+++ b/scripts/pricing-v2.ts
@@ -1,0 +1,183 @@
+import Stripe from 'stripe';
+import readline from 'node:readline';
+import {
+  PRICING_AMOUNTS,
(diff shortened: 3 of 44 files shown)


aalemayhu:
Pushed two review fixes (4463e19):

- **v2 fallback visibility** — `UnlimitedCheckoutUseCase` now logs the lookup_key fallback at `console.error` with `cohort`, `interval`, `lookup_key`, and an explicit `reason`, so a Stripe-API-down situation serving wrong-cohort legacy prices is visible in prod logs. Fallback behavior unchanged — checkout still completes on the legacy price id.
- **Cross-request cache** — `StripePriceResolver` cache moved from a per-instance Map to a module-level Map, so it stays effective regardless of how the resolver is constructed (not just the single boot-time instance). Resolved price ids are immutable per lookup_key, so no TTL. Added a test asserting the cache is shared across instances, plus a test-only reset helper for isolation.

Affected Jest suites green (24 tests), server tsc clean.

Note: the pre-push tsc hook reported a false `Cannot find module './pricingV2'` — `tsc -p . --noEmit` passes with exit 0 locally and the module is tracked in HEAD, so this was a hook env false positive, not a real type error.

aalemayhu:
## Ops command added: Create v2 prices

`/ops/commands` now has a **Create v2 prices** button next to the Stripe sync action. One click creates the $7.99/mo (`v2_monthly`) and $64/yr (`v2_annual`) prices on the existing Unlimited product — no more SSH-running `scripts/pricing-v2.ts`.

**Behavior**
- Idempotent: each price is looked up by `lookup_key` first and created only when absent. Safe to run twice.
- Never mutates, updates, or deletes existing prices, products, or subscriptions (asserted in tests).
- Admin-gated by `RequireOpsAccess` (404 for non-owners), swagger-annotated, uses the server's configured Stripe client so live/test mode follows `STRIPE_KEY`.
- Result rendering shows per-key status (created vs already exists), the price IDs, and the live/test mode.

**Script relationship**
- `scripts/pricing-v2.ts` now wraps `CreatePricingV2PricesUseCase` for the creation step but keeps its standalone guards (catalog print, `--live` mode check, typed confirmation prompt). The script still works exactly as before; the creation logic lives in one place.

**Note on the live run**
- The endpoint creates prices the moment it's hit with a live key. The live run is still gated on your price sign-off — clicking the button on prod creates the real $7.99/$64 prices. The script's "type `create live prices`" prompt is bypassed from the UI by design (the button is admin-only), so only click it once you've signed off on the amounts.

Endpoint: `POST /api/ops/create-pricing-v2-prices`

PR #3208, opened 2026-06-10
fix: stop pricing-v2 price leaks on the live pricing page

aalemayhu:
## What
Stops the pricing-v2 $6 leaks on the live pricing page. The Unlimited hero and Unlimited card already show the correct cohort price ($5.33/mo annual, $7.99/mo monthly for v2 visitors; $6 for the grandfathered legacy window), but several secondary surfaces still hardcoded the legacy $6 — so a new visitor saw $5.33/$64/yr on the card and $6 in the comparison table, FAQ, and JSON-LD.

## Why
Pricing v2 shipped + flag ON in prod (#3204). Verified on prod: anon/v2 visitors get the right hero and Unlimited card, but the comparison table header, the Unlimited FAQ answer, the FAQ JSON-LD schema (generated from the same data), and a few landing/paywall surfaces still said $6. A visitor reading "$6" and getting charged $7.99 at checkout is a trust break.

## How
- **ComparisonTable** now takes the resolved monthly price as a prop. `PricingPage` already fetches `GET /api/checkout/prices` and resolves `pricing.monthlyCents`; it passes `formatMonthly(pricing.monthlyCents)` down. No second fetch in the table. Legacy cohort renders `$6 / mo`, v2 renders `$7.99 / mo`, formatted exactly like the Unlimited card.
- **FAQ** Unlimited answer is now price-neutral ("Unlimited removes the 100-card limit … Pricing is on the cards above."). It is true for every cohort, so the static `FAQPage` JSON-LD never states a per-visitor price to crawlers. Not made cohort-dynamic by design — the JSON-LD must not lie.
- **Static / non-fetching surfaces** that don't have cohort prices — `NotionLandingPage` (SEO), `LimitPage` Unlimited card, the downloads `PaywallBanner` button, and the `AnswersPage` Unlimited blurb — drop the hardcoded $6 rather than name a price that varies by cohort. `LockInBanner` keeps `$6/month` untouched: it only renders when `prices.legacy && lockInDeadline != null`, i.e. exclusively for the legacy-window cohort that actually gets $6.
- Removed the now-unused `MONTHLY_PRICE` / `YEARLY_PRICE` constants (kept `MONTHLY_SUFFIX`, `AUTO_SYNC_PRICE` — still used).

## Measuring success
On prod, load `/pricing` as an anonymous (v2) visitor: the comparison-table Unlimited column header reads `$7.99 / mo`, matching the card; the FAQ Unlimited answer names no price; `view-source` on the FAQPage JSON-LD contains no `$6`/`$7.99`. A legacy-window account still sees `$6 / mo` in the table. Network: `GET /api/checkout/prices` is still fetched exactly once on the pricing page (no new request from the table).

## Testing
- ComparisonTable: asserts legacy cohort renders `$6 / mo` and v2 renders `$7.99 / mo` (mutually exclusive).
- PricingPage integration: default (legacy fallback) renders `$6 / mo` in the table; a mocked v2 price response renders `$7.99 / mo` and no `$6 / mo`.
- PricingFaq: Unlimited answer matches no `$N` literal and keeps the "removes the 100-card limit" copy.
- NotionLandingPage / LimitPage / PaywallBanner: assert no hardcoded `$6`, CTA still present.
- All 102 tests across the 7 touched test files pass; web typecheck clean.

## Browser check
Notes: Verified with Playwright Chromium against pnpm dev on this branch (legacy cohort): Yearly/Monthly toggle renders with Yearly preselected, Monthly click switches hero to $6 + "renews monthly", comparison table shows cohort price, FAQ is price-neutral. v2-cohort $7.99 table rendering covered by the added unit tests. Only console message at 375px is the pre-existing unauthenticated-session 401, identical on prod.

## Changelog
No entry. This is a price-display consistency fix for a feature whose prices just changed; the pricing-v2 / annual-default entry already covers the surface, and a "now shows $7.99" line would re-state a number users already see on the card.

## Risks
Low. Pure display + one prop thread-through. If the table ever renders before `getCheckoutPrices` resolves, it shows the legacy fallback ($6) — same fallback the card already uses, and v2 cohort flips to $7.99 the moment the fetch lands. Rollback is a clean revert of one branch.

## Goal alignment
Removes a price/charge mismatch on the highest-intent page in the funnel (pricing). A visitor who reads $6 and is charged $7.99 abandons or charges back; consistent pricing protects the paid-conversion step that the 300K-user goal depends on.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3208"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783705434&installation_id=132681956&pr_number=3208&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3208&signature=fbd29338835999076e3b2516ebeb473808030610b2bcbfeb6eb148e658dd8e47"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/web/src/pages/AnswersPage/answersConfig.ts
+++ b/web/src/pages/AnswersPage/answersConfig.ts
@@ -123,7 +123,7 @@ const pdfToAnki: AnswerConfig = {
     },
     {
       heading: 'Large PDFs and textbooks',
-      body: 'Large PDFs work — a whole textbook uploads fine. Big files take longer and create large decks. Uploading one chapter at a time keeps decks easier to review and share. The free plan covers 100 cards per month; the Unlimited plan ($6/month) has no card limit.',
+      body: 'Large PDFs work — a whole textbook uploads fine. Big files take longer and create large decks. Uploading one chapter at a time keeps decks easier to review and share. The free plan covers 100 cards per month; the Unlimited plan has no card limit.',
     },
   ],
   relatedLinks: [
--- a/web/src/pages/DownloadsPage/components/PaywallBanner.test.tsx
+++ b/web/src/pages/DownloadsPage/components/PaywallBanner.test.tsx
@@ -47,7 +47,7 @@ function buildJob(overrides: Partial<JobResponse> = {}): JobResponse {
 
 function getUpgradeButton() {
   return screen.getByRole('button', {
-    name: /Upgrade to Unlimited — \$6 \/ mo/,
+    name: /Upgrade to Unlimited/,
   });
 }
 
--- a/web/src/pages/DownloadsPage/components/PaywallBanner.tsx
+++ b/web/src/pages/DownloadsPage/components/PaywallBanner.tsx
@@ -6,10 +6,6 @@ import { getDistance } from '../../../lib/getDistance';
 import { firePaywallEvent } from '../../../lib/analytics/firePaywallEvent';
 import { track } from '../../../lib/analytics/track';
 import { get2ankiApi } from '../../../lib/backend/get2ankiApi';
-import {
-  MONTHLY_PRICE,
-  MONTHLY_SUFFIX,
-} from '../../PricingPage/pricing.constants';
 import styles from './PaywallBanner.module.css';
 
 interface PaywallBannerProps {
@@ -65,9 +61,7 @@ export function PaywallBanner({ inProgressJob }: PaywallBannerProps) {
           onClick={handleUpgrade}
           disabled={pending}
         >
-          {pending
-            ? 'Opening checkout'
-            : `Upgrade to Unlimited — ${MONTHLY_PRICE} ${MONTHLY_SUFFIX}`}
+          {pending ? 'Opening checkout' : 'Upgrade to Unlimited'}
         </button>
         {inProgressJob != null && startedDistance != null && hasTitle && (
           <span className={styles.secondary}>
--- a/web/src/pages/LimitPage/LimitPage.test.tsx
+++ b/web/src/pages/LimitPage/LimitPage.test.tsx
@@ -84,9 +84,10 @@ describe('LimitPage', () => {
     expect(screen.queryByText('Get Auto Sync')).toBeNull();
   });
 
-  it('shows the Unlimited price', () => {
+  it('does not show a hardcoded Unlimited monthly price', () => {
     renderPage();
-    expect(screen.getByText('$6')).toBeTruthy();
+    expect(screen.queryByText('$6')).toBeNull();
+    expect(screen.getByText('Upgrade to Unlimited')).toBeTruthy();
   });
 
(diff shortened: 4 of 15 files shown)


aalemayhu:
### Added: restore the Monthly/Yearly toggle on `/pricing` (commit 7533dbd8defe)

**Bug (verified on prod):** `UnlimitedCard` renders its billing-cycle radiogroup only when `yearlyAvailable` is true. `PricingPage` took that value through an `unlimitedYearlyAvailable` prop that **defaulted to `false`**, and `App.tsx` never passed it. Result: no toggle rendered, and since annual is the default cadence, **monthly Unlimited was unbuyable from the UI** for every visitor — independent of the `pricingV2On` flag (the flag only changes the price numbers, not the cadence choice; the legacy fallback constants `$6/mo` · `$60/yr` keep annual resolvable even with the flag off).

**Fix:** dropped the dead `unlimitedYearlyAvailable` prop entirely (removed from `PricingPage` props, defaults, the JSX, and all test usages — grep confirms zero remaining references) and derive availability from the pricing state the page already fetches: `yearlyAvailable={pricing.annualCents > 0}`. `annualCents` is always a resolved number (legacy fallback or the `/api/checkout/prices` result), so the toggle is now **visible by default** for everyone. No new env var or flag.

**Prior purpose of the prop:** introduced in #2388 (yearly billing) as a gating point, but it was never wired into `App.tsx`, so it has been dead the whole time and silently hid the toggle in prod. No deliberate "hide yearly" intent in the history — the spec wanted the toggle visible.

**Tests (vitest, 51 passing in the suite):**
- toggle renders by default with both radios present, Yearly selected
- clicking Monthly switches the hero to the monthly price and the terms line to "renews monthly"

**Playwright:** `web/tests/checkout.spec.ts` already asserts the yearly hint, annual-default, and `interval: 'year'` checkout — all of which this change restores at the live-app level. The spec never referenced the removed prop, so it stays green. I did not run `test:e2e:with-mock` locally — the Playwright config auto-starts the dev server as a `webServer`, and per repo policy I don't boot the dev server unprompted. Confirmed compatibility statically.

Issue #3183, opened 2026-06-07
Notion images and audio: front-side render, width, and expiring audio links


aalemayhu:
**Labels:** bug, notion-api, high-priority
**Goal alignment:** Media is core to the dominant, pay-willing user segment (med + language learners) surfaced repeatedly in feedback. A card with a missing image or a dead audio link is an unusable deck — it breaks "drop something in, get a clean deck back" for exactly the users who upload the most.

### Problem
Across many YouTube comments over multiple years, users report three related media failures in converted cards:
1. **Images don't appear on the front side** of a card when embedded inside a Notion toggle (clinical images, diagrams).
2. **Audio added in Notion converts to a bare link**, not playable audio — matches the known signed-S3-URL-expiry failure mode (the asset URL expires ~1h after generation, leaving a broken element).
3. **Images render too narrow / cramped** — a white frame is added and width can't be enlarged in Anki.

This is the single most recurrent media theme in the feedback and hits the core ICP.

### User signal (verbatim)
- "I'm a medical student… I want to include images like ultrasound scans inside the toggle. However, when I export to Anki, the images don't appear on the front side of the flashcard. Is there any way to fix this so the images show up correctly on the front?"
- "how can I put an image in the front of card? I really really would like to put an image in the front by using notion2anki, but I don't know how to do that."
- "When I put audio on notion and export it to Anki, it shows only a link on Anki, not the audio itself, is there any way to have the audio on Anki from notion?"
- "my images are a bit narrow and small… any way to have more width in the images?"

### Proposed direction
Audit the toggle/column conversion path so every Notion-hosted asset (`type === 'file'`: images, audio) is routed through `BlockHandler.embedImage` / `embedAudioFile` — download bytes via `instrumentedAxios`, bundle as durable `.apkg` media with the SHA-512 filename — rather than passed through as a raw signed `file.url` that expires (`type === 'external'` stays a link). Confirm front-side image placement inside toggles is supported, and revisit the default image width / white frame so images render at a usable size. Pair the bot review with manual runtime-thinking on URL lifecycle (the bot won't model expiry). Add a regression test asserting a `file`-type Notion asset ends up as bundled media, not an inline expiring URL. See the S3-expiry gotcha in CLAUDE.md and the embed helpers in NotionService.

aalemayhu:
Overnight progress on this issue, split into slices:

- **Image-URL expiry** (cards showing a broken image once Notion's signed S3 URL lapses) — covered by #3260.
[...]

PR #3265, opened 2026-06-11
fix: show toggle summary media on the card front

aalemayhu:
## What
When a Notion toggle's summary rich_text is empty and its first child is an image or audio block, that media now renders on the card **front** instead of being pushed to the back with a blank front. Also fixes off-center images in card output.

## Why
Part of https://github.com/2anki/server/issues/3183. Recurring complaints: an ultrasound/image or an audio clip that the user put as a toggle's "question" landed on the back (or nowhere) because the front was built only from the toggle summary text. The off-center "white frame" look came from the `img` rule lacking `height: auto` and centering.

## How
- `BlockHandler.getFlashcards` toggle branch: when the rendered summary front is empty, fall back to `getToggleFrontMedia`, which fetches the toggle's children and, if the first child is `image`/`audio`, renders it via `blockToStaticMarkup`. Both media types route through `embedImage` / `embedAudioFile`, so the asset bundles into the `.apkg` as durable media (SHA-named file), never a raw expiring Notion S3 URL.
- `src/templates/notion.css`: `img` gains `height: auto; display: block; margin: 0 auto`. No forced upscaling — small images stay their natural size.

Diff is kept tight to the toggle-front region.

## Measuring success
No new metric. Behavioral fix verified by the new Jest test asserting front contains `` / `[sound:…]` and not the raw URL. In prod, fewer "image/audio not on the front" reports against #3183.

## Testing
- Unit tests added: `BlockHandler.toggleFrontMedia.test.ts` — image-only toggle → `` on front bundled as media; audio-first toggle → `[sound:…]` on front bundled; empty-summary + non-media first child → still skipped (no false card). Mocks only the Notion API and the HTTP download edge.
- Full `BlockHandler` suite (14 files, 95 tests) green — text-summary toggles unchanged.

## Browser check
Browser check: not applicable — the only `web/src/` file is a changelog JSON entry (attestation auto-bypass); CSS/render change is in server templates.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-06-12-toggle-front-media.json` — "Notion toggles with an image or audio summary now show that media on the card front"

## Decisions made overnight (Alexander asleep)
- **CSS: center + `height: auto`, not upscaling.** Forcing small images larger degrades large medical scans (the recurring ultrasound case) and is irreversible per-image without a width control. Centering fixes the reported white-frame look without touching image fidelity.
- **Trigger is empty-summary-only.** Rendering media on the front *only* when the toggle summary is empty is the non-breaking subset: toggles with text summaries are untouched. "Media on the front when the summary also has text" is deferred (would change existing cards).

## Deferred (still open on #3183)
- Per-image width control.
- Richer front placement (media on the front alongside a text summary).

## Merge order
PR #3257 (`fix/tag-registry-scope`) also refactors `BlockHandler.ts` (threads a per-instance tag registry through renderers). **Merge #3257 first; this PR then rebases** — conflicts in `BlockHandler.ts` are likely.

## Sonar
Not run locally — no SONAR_TOKEN configured in this environment. Expect a possible Sonar pass on CI.

## Risks
Low. New code path only fires for toggles whose summary rich_text is empty; everything else takes the existing branch. If a toggle has an empty summary and a media first child it previously produced no front card (skipped) — now it produces one. Rollback: revert the BlockHandler toggle-branch hunk; the CSS hunk is independent and harmless.

## Goal alignment
Quality/retention, not a direct funnel metric — "simplest, fastest way to turn what you're studying into beautiful flashcards." Medical and language learners (audio) hit this specific breakage; fixing it reduces a known abandonment cause on the conversion → download path. No funnel event added (internal correctness fix on existing surface).

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3265"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783778722&installation_id=132681956&pr_number=3265&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3265&signature=56b345a9b8ec0c29da1df8d4269f6ddf1909909f20995652f6a4d1b11d5bf4c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


aalemayhu:
CI fixed: regenerated the starters baseline fixture for the image-centering CSS change. One thing to verify in your browser pass — the live starter CSS isn't uniform: the n2a starters appeared to serve a CRLF copy of the img block, so confirm the centering actually renders on an n2a-basic deck, not just the only-notion starters.

PR #3284, opened 2026-06-12
fix: cut wasted Notion and offline calls in Ankify polling

aalemayhu:
## What
Two contained efficiency fixes for the 5-min Ankify Notion polling loop.

1. **Remember the Notion object type per subscription.** A new nullable `notion_object_type` column is set on first successful resolution (`page`/`database`). Once a subscription is known to point at a Notion database, the use case skips the doomed `pages.retrieve` on later ticks — in both the card walk and the page-meta fetch. The polling worker's inline meta fetcher is replaced with the shared `buildNotionPageMetaFetcher`, which now honors the known type too.
2. **Cheap AnkiConnect reachability pre-check on polls.** Polls now `ping()` (a `version` invoke) before any Notion fetch. If the client is offline, the tick skips the Notion round-trip entirely, records a calm `last_error` ("Anki client offline — will retry next tick"), logs one info line, and returns an offline-skip sentinel. Manual and webhook syncs do not pre-empt — they run normally and surface `AnkiConnectUnreachableError` as before.

## Why
Prod evidence (last 48h, two enabled database subscriptions, owner 13574):
- **507** `@notionhq/client` `validation_error` warn lines — one guaranteed-failing `pages.retrieve` per database subscription per tick, before the existing `databases.retrieve` fallback. Wasted Notion free-tier rate budget (the reason the cadence is 5 min) and log spam that drowns real errors.
- **39** `AnkiConnectUnreachableError` aborts — the user's local Anki was offline, yet each tick still did the full Notion fetch + card build before failing at the AnkiConnect step, writing confusing "N cards failed: … unreachable" into `last_error`.

User-visible outcome: an Ankify user with Anki closed sees a calm "offline" status instead of failed-card errors; the type-skip is internal efficiency.

## How
- New use-case input `knownObjectType` (optional); the polling tick forwards `current.notion_object_type`. `walkSource` short-circuits to the database walk when the type is known, and records the resolved type via a new `recordObjectType` repository method (self-heals on first tick, no backfill).
- `execute` builds the `AnkiConnectClient` once, probes it on `trigger: 'polling'`, and returns `AnkifyClientOfflineSkip` when unreachable. The probe rethrows any non-unreachable error.
- `RefreshAnkifySubscriptionUseCase` and `AnkifyController` narrow the new union return type with an exported `isAnkifyClientOfflineSkip` guard.

**Construction sites checked** (`grep -rn "new SyncNotionPageToRacUseCase"`): `src/data_layer/index.ts` (the polling worker — the reporter's actual runtime path, gains both behaviors and was the one verified), `src/routes/AnkifyRouter.ts`, `src/routes/AnkifyWebhookRouter.ts`. The signature change is input-only (`knownObjectType` optional) so all three compile unchanged; manual/webhook paths keep their existing error behavior.

**Migration / kanel:** `migrations/20260912000000_add_notion_object_type_to_ankify_subscriptions.js` adds a nullable `text` column (non-locking `ALTER ADD COLUMN`, no default, no table rewrite; `down` drops it). `pnpm kanel` could not run in the worktree (no live Postgres connection / kanel bin), so **`pnpm kanel` must be run post-merge** to regenerate `src/data_layer/public/AnkifyNotionSubscriptions.ts`. Code treats the column as optional (`notion_object_type?: NotionObjectType | null` on the hand-written entity), so it compiles before kanel runs.

## Measuring success
After deploy, the `@notionhq/client` `validation_error` warn count for database subscriptions should drop to roughly one per subscription total (first tick only, then `notion_object_type` is set), and the per-tick `AnkiConnectUnreachableError` aborts for offline clients should be replaced by a single `[ankify-polling] Anki client offline for owner <id>; skipped Notion fetch` info line per tick. Confirm via `pm2 logs` grep for those two strings and by reading `ankify_notion_subscriptions.notion_object_type` / `last_error` in prod psql.

## Testing
- Unit tests added/extended:
  - `SyncNotionPageToRacUseCase.test.ts` — records `database` after the not-page fallback; skips the page walk entirely when type is already known; offline poll skips the Notion fetch + records the calm `last_error` + returns the skip; manual sync does NOT pre-empt when offline.
  - `scheduleAnkifyPolling.test.ts` — forwards `knownObjectType` from the current row.
  - `buildNotionPageMetaFetcher.test.ts` — skips `pages.retrieve` when the type is known to be a database.
  - `RefreshAnkifySubscriptionUseCase.test.ts` — surfaces `AnkiConnectUnreachableError` when the sync reports an offline client.
  - `AnkifyNotionSubscriptionsRepository.notion_object_type.test.ts` — new repo method persists and overwrites the type.
- `npx tsc -p .` clean; affected suites green (180 ankify tests). The 3 unrelated suites that fail locally (`getPageCount`, `runParserCanary`, `DeckParser`) fail because `pdfinfo`/Python binaries aren't installed in this worktree — they pass in CI and touch none of this code.

## Browser check
Browser check: not applicable — server-side polling-job change; the only `web/src/` file is a changelog JSON entry.

## Changelog
`Notion sync shows a calm "Anki client offline" status instead of failed-card errors when Anki is closed` (`web/src/pages/WhatsNewPage/changelog/2026-06-12-ankify-offline-status.json`). The calmer `last_error` is the one user-perceivable piece; the type-skip is internal-only.

## Sonar
`sonar-scanner` was not run locally — `SONAR_TOKEN` is unset and the worktree lacks web/node_modules + the pdf/python binaries needed for a coverage build. Expect the PR-side analysis to run on CI. Self-reviewed for the common smells: small new methods, no nested ternaries, guard clauses lead with the positive, no catch-and-rethrow-same. oxlint flagged only a pre-existing `no-negated-condition` warning unrelated to this diff.

## Risks
Low. The type-skip self-heals and falls back to the existing page→database walk when the column is null (pre-migration / new rows). The offline pre-check adds one cheap `version` invoke per poll tick; if AnkiConnect is up it succeeds in milliseconds. Manual/webhook sync behavior is unchanged. Rollback: revert the PR and run the migration `down` (drops the nullable column).

## Goal alignment
None — internal reliability/cost. Reduces wasted Notion free-tier rate budget and log noise on a paid surface (Auto Sync / lifetime Ankify), and gives those users an accurate status when Anki is closed instead of confusing failure text. No funnel or MRR metric moves.

---
<a href="https://app.blacksmith.sh/2anki/codesmith/server/pr/3284"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783859483&installation_id=132681956&pr_number=3284&repository=2anki%2Fserver&return_to=https%3A%2F%2Fgithub.com%2F2anki%2Fserver%2Fpull%2F3284&signature=8896b07faca088a2284afb52c41aa2ce2029593be4bdd6585a1e633b016f1496"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


aalemayhu:
Migration-reviewer verdict: **ship it.** Nullable `ALTER ADD COLUMN` is catalog-only in Postgres 11+ (no default, no NOT NULL, no index) — no lock risk at any scale. `down` cleanly drops the derived column. Self-healing backfill via `recordObjectType` on first poll tick is correct.

One action at merge: run `pnpm kanel` and commit the regenerated `src/data_layer/public/AnkifyNotionSubscriptions.ts` (it must gain `notion_object_type` in the row/initializer/mutator shapes). The worktree has no kanel bin or DB connection, so it couldn't run here. Nothing breaks before then — the repository imports `NotionObjectType` from `src/entities/ankify.ts`, not the generated type, so compilation is unaffected.

PR #3382, opened 2026-06-15
fix: strip internal data from status and checkout copy

aalemayhu:
## What
Removes internal/developer-grade data from three user-facing surfaces: the status page stops showing a raw git SHA and raw ISO/`toLocaleString` timestamps, the post-checkout screen drops the raw full-URL reference lines, and the account-claim page moves off an inline style object onto shared tokens while replacing a raw route path in its copy.

## Why
VOICE.md bans internal IDs, ISO timestamps, and dev paths in user copy; DESIGN.md kill-list flags "exposing internal identifiers as status text." The status page and the post-payment screen are high-trust moments — leaking a git hash, a route path (`/account`), or a copy-paste URL there reads as unfinished. This is a data-presentation / copy sweep, TSX-only, disjoint from the in-flight CSS PRs.

## How
- **StatusPage**: dropped the 7-char SHA; relabeled the deploy row "Version" to "Updated"; added `formatRelativeFromIso` (a thin ISO wrapper over the existing `formatRelative` helper already used for the Notion/Stripe rows) and routed the deploy time plus both incident timestamps through it, replacing the raw `new Date(...).toLocaleString()` calls. The "Last deploy" section now keys on `lastDeploy.time` rather than `sha`. Live-incident red semantic untouched. `StatusPage.module.css` not touched (owned by an in-flight PR).
- **UserActionCards**: removed the `<a>{fullUrl}</a> (link for reference)` lines under each button; the buttons are the links.
- **AccountClaimPage**: swapped the inline `style={{...}}` object plus unclassed `<h1>`/`<p>` for `sharedStyles.pageNarrow` + `sharedStyles.card` + `sharedStyles.title` (same pattern as `FeedbackPage`); fixed the already-used copy "...from /account if you need to reclaim" to "...from your account page if you need to reclaim".

## Measuring success
This is a copy/presentation change with no new event. Confirm visually in prod: load `/status` and the deploy and incident rows render relative phrases ("2 hours ago"), no hex string; load the post-checkout screen and there is no `https://...` body text; trigger the account-claim already-used state and there is no `/account` literal in the sentence. The before/after comment below shows the rendered diff captured with Playwright against the real `base.css`.

## Testing
- Unit tests added (Vitest): StatusPage — "renders the last deploy as a relative time, never a raw git SHA" (asserts no 7-40 char hex in the rendered body) and "renders incident timestamps as relative time, not toLocaleString output" (asserts the `toLocaleString` sample is absent and the relative phrases render). Existing StatusPage / AccountClaimPage / UserActionCards suites still green (20 tests across the three pages).

## Browser check
Notes: Pending — agent cannot drive a browser. Rendered output verified via Playwright stills against the real `base.css` (see before/after comment). Reviewer to tick after a local pass.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-06-15-status-readable-copy.json` — "The status page shows when things last updated in plain language, and the post-checkout screen drops the raw link text"

## Risks
Low. The `StatusPayload.sha` field stays in the type (API contract) but is no longer rendered. If the status API ever sends a deploy with a `sha` but no `time`, the "Last deploy" section now hides (previously it showed on `sha`); acceptable since the only remaining value in that section is the relative time. Rollback is a straight revert.

## Goal alignment
Trust lever. The status page and the post-payment screen (the highest-trust moment in the funnel) stop leaking developer-grade data — no funnel event attached; this is brand/trust hygiene, not a conversion experiment.

## Deferred
- **SuccessfulCheckout dual-primary button hierarchy** (two `btnPrimary` buttons side by side) is a revenue/hierarchy decision for the trio, out of scope here — left unchanged.
- Any `*.module.css` change (`StatusPage.module.css` owned by an in-flight CSS PR).

### DIFF
--- a/web/src/pages/AccountClaimPage/AccountClaimPage.tsx
+++ b/web/src/pages/AccountClaimPage/AccountClaimPage.tsx
@@ -55,52 +55,56 @@ export default function AccountClaimPage() {
   if (state === 'loading') return <SkeletonPage rows={2} />;
 
   return (
-    <div style={{ maxWidth: 480, margin: '4rem auto', padding: '0 1.5rem' }}>
-      {state === 'success' && (
-        <>
-          <h1>Subscription claimed.</h1>
-          <p>Your subscription is now attached to this account.</p>
-          <a href="/account" className={sharedStyles.btnPrimary}>
-            Go to account
-          </a>
-        </>
-      )}
-      {(state === 'expired' || state === 'error') && (
-        <>
-          <h1>Invalid or expired link.</h1>
-          <p>
-            This confirmation link is no longer valid. Start over from your
-            account to request a new one.
-          </p>
-          <a href="/account" className={sharedStyles.btnPrimary}>
-            Go to account
-          </a>
-        </>
-      )}
-      {state === 'already_claimed' && (
-        <>
-          <h1>Link already used.</h1>
-          <p>
-            This link is already used. Sign in and try again from /account if
-            you need to reclaim.
-          </p>
-          <a href="/account" className={sharedStyles.btnPrimary}>
-            Go to account
-          </a>
-        </>
-      )}
-      {state === 'active_sub' && (
-        <>
-          <h1>Account already has a subscription.</h1>
-          <p>
-            This account already has an active subscription. Cancel it first or
-            contact support.
-          </p>
-          <a href="/account" className={sharedStyles.btnPrimary}>
-            Go to account
-          </a>
-        </>
-      )}
+    <div className={sharedStyles.pageNarrow}>
+      <div className={sharedStyles.card}>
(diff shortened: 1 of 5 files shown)


aalemayhu:
## Before / after — captured with Playwright

Rendered against the real `web/src/styles/base.css` tokens (Playwright stills, 2x DPR). Source change is TSX-only.

### Status page — raw SHA + ISO/locale timestamps to relative phrasing
status before/after

### Account-claim card — inline style to shared tokens, and `/account` route path out of the copy
account-claim before/after

The checkout change (removing the `https://2anki.net/login (link for reference)` lines under each button) is copy-only — covered by the diff in `UserActionCards.tsx`, no screenshot.

PR #3615, opened 2026-07-14
spec: failed-payment recovery nudge (payments hard rail — needs gate)

aalemayhu:
## What

Draft spec (`Documentation/specs/failed-payment-recovery.md`) for an involuntary-churn
recovery nudge: on Stripe's `invoice.payment_failed` webhook, email the affected subscriber a
card-update link so they can rescue their subscription inside the existing dunning grace window.
Ships an **inert, unwired** email template (`failed-payment-recovery.html`) so the copy and layout
are reviewable. **No code path, webhook handler, charging logic, or send path is wired in this PR.**

## Why

Business metrics show ~10 failed payments/week — larger than the entire week's ~8 new-paid
conversions. We already grant paid access during dunning (`past_due`/`unpaid` in
`ACCESS_GRANTING_STATUSES`), but we never *tell the user* their card failed. The gap is a recovery
nudge. Recovering even part of this is near-zero-CAC MRR defense — cheaper than any acquisition dollar.

## Payments hard rail — needs your gate

This touches Stripe + a commercial email, both hard rails. I deliberately did **not**:
- add the `invoice.payment_failed` case to `WebhookRouter.ts` (it is still unhandled today),
- wire the template into `EmailService`/`constants.ts` or any send path,
- change charging, retries, dunning grace, or `ACCESS_GRANTING_STATUSES`,
- register a sender or create a migration.

**Requesting:** your explicit go/no-go, plus a `/security-review` before any implementation lands.
This PR stays **draft** until then. Do not merge.

## Open decisions for you

1. **Transactional vs marketing.** "Your payment failed, update your card" is arguably transactional
   (it concerns the user's own paid service), which would mean *no* `marketing_opt_out` exclusion so
   opted-out payers still hear their subscription is at risk. `email-templates.md` classifies any
   upsell-adjacent email as commercial. The spec + template default to commercial/conservative
   (unsubscribe footer + opt-out filter). Your call.
2. **Send timing / cadence.** Send on the first `invoice.payment_failed`, or wait for the 2nd/3rd
   retry so Stripe's silent retry can succeed first? Spec proposes first-failure, one email per invoice.
3. **How many retries / follow-ups.** Spec proposes exactly one email per failed invoice.
4. **Link mechanism.** Option A (`hosted_invoice_url` — zero new Stripe config) vs Option B (billing
   portal session). Spec proposes A.
5. **Migration.** A `failed_payment_recovery_notifications` dedupe table needs a Knex migration +
   `pnpm kanel` in the implementation PR (hard gate). Confirm table shape first.

## Success metric

Recovered subs ÷ failed payments over a trailing 4-week window (a recovery = `past_due|unpaid →
active` after a recorded failed-payment notification). Send volume via `email_batch_sent`
(`campaign: 'failed_payment_recovery'`); cross-checked against `BusinessMetricsService` failed-payment
count. Proposed target for you to set: recover ≥20% of would-be involuntary churn.

## Testing

- None yet — spec + inert HTML only, no source logic. The eventual implementation PR carries an
  outside-in webhook test (Stripe SDK mocked: one send, correct link, per-invoice dedupe, opt-out skip).

## Changelog

No entry — draft spec + unwired template, no user-visible behavior ships.

## Risks

None from this PR (nothing executes). The risk surface is entirely in the future implementation and
is the reason for the gate: a mistake in the `invoice.payment_failed` handler or the send path could
email the wrong users or interact with charging. Hence draft + `/security-review` before build.

## Goal alignment

Directly targets MRR/retention — the primary lever per CLAUDE.md. Defends existing revenue against
involuntary churn at near-zero CAC. Metric to move: recovered-subs rate, read from the ops
business-metrics failed-payment count. Not an acquisition change.

### DIFF
--- a/Documentation/specs/failed-payment-recovery.md
+++ b/Documentation/specs/failed-payment-recovery.md
@@ -1,0 +1,114 @@
+# Failed-payment recovery (involuntary-churn nudge)
+
+Status: DRAFT — payments hard rail. Requires the maintainer's explicit gate and a
+`/security-review` before any implementation. This spec proposes no live change.
+
+## Problem
+
+Business metrics show ~10 failed payments/week — larger than the entire week's ~8 new-paid
+conversions. A failed renewal charge is involuntary churn: the subscriber did not choose to
+leave, their card expired, hit a limit, or bounced. Recovering even part of this is near-zero-CAC
+MRR defense — cheaper than any acquisition dollar.
+
+We already grant paid access during dunning: `updateStoreSubscription` treats `past_due` and
+`unpaid` as access-granting statuses (`ACCESS_GRANTING_STATUSES` in `src/lib/integrations/stripe.ts`),
+so a grace window exists. The gap is that we never *tell the user* their payment failed or give them
+a one-click way to fix their card. Stripe's own retry emails are off/inconsistent for this account.
+
+## What to build
+
+A single recovery nudge triggered by Stripe's `invoice.payment_failed` webhook:
+
+1. **Trigger:** add an `invoice.payment_failed` case to the `/webhook` switch in
+   `src/routes/WebhookRouter.ts`. It is not currently handled (the `default` arm logs it as
+   unhandled). No change to any charging, provisioning, or access logic.
+2. **Resolve the account** from the invoice's customer (reuse `getCustomerId` +
+   `stripe.customers.retrieve`, same as the existing subscription cases). Skip silently if no
+   2anki account resolves (do not leak an unlinked-payment alert here — that path already exists
+   for provisioning).
+3. **Send one card-update email** with a link the user can click to fix their card. Two link
+   options — the maintainer picks (see Open decisions):
+   - **Option A (simplest, no Stripe config):** the failed invoice already carries
+     `hosted_invoice_url` — a Stripe-hosted page where the customer can pay the open invoice and
+     update the card. Zero new Stripe surface, no portal configuration.
+   - **Option B (billing portal):** `stripe.billingPortal.sessions.create({ customer })` returns a
+     managed card-update session. Requires the Stripe Customer Portal to be configured and a
+     `return_url`. More flexible but more setup and a new Stripe call.
+4. **Idempotency + cadence:** record each send in a new
+   `failed_payment_recovery_notifications` table (mirror `subscription_recovery_notifications`:
+   `id`, `email`, `stripe_invoice_id`, `sent_at`). Before sending, check we have not already
+   emailed for this `stripe_invoice_id`. Stripe fires `invoice.payment_failed` once per retry
+   attempt in the dunning schedule; without dedupe on the invoice we would email on every retry.
+5. **Template:** `src/services/EmailService/templates/failed-payment-recovery.html`, following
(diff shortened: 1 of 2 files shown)


aalemayhu:
## Security review — verdict: SAFE-TO-IMPLEMENT-AS-SPEC'D

Reviewed the spec + the inert template against `.claude/rules/security.md` and the live `/webhook` + email rails. No Critical/High findings. The template is genuinely inert (not registered in `EmailService/constants.ts`, imported nowhere), so this draft ships no executable risk. The design's security-load-bearing pieces are correct-by-design or reuse a verified existing rail (raw-body Stripe signature verify, per-invoice atomic dedupe, authoritative customer re-fetch, Stripe-scoped link, static template with no user-controlled interpolation, opt-out gate).

**Requirements to carry into the implementation PR (not blockers on this spec):**
1. **CWE-345 dedupe** — the dedupe migration MUST put a UNIQUE constraint on `stripe_invoice_id`, or `onConflict().ignore()` silently no-ops and duplicate emails leak. Confirm via `pnpm kanel`.
2. **CWE-20 account resolution** — send-to email must come from the *retrieved* Stripe customer (`stripe.customers.retrieve`), never the webhook payload's nested object. Mirror the existing subscription cases.
3. **CWE-532 logging** — log only hashed surrogates (`hashToken(invoiceId)`/`hashToken(email)`), never raw invoice id, customer id, or email. Mirror the `session_id_hash` pattern.
4. **Robustness** — `hosted_invoice_url` is null on non-finalized invoices; null-check and skip, don't email an empty href.
5. **Consent** — default to the commercial classification (opt-out filter + unsubscribe footer, both present). Do not flip to "transactional / skip opt-out filter" without a documented justification — mis-sending to `marketing_opt_out = true` users is the real CAN-SPAM/GDPR exposure.

Re-run `/security-review` on the actual wiring PR (the migration + webhook case + send path), not just this spec.

aalemayhu:
Closing — Stripe already does this. Verified in the Stripe Dashboard → Revenue recovery:

- **Emails**: "Send emails when card payments fail" is ON (Stripe-hosted card-update page), plus "Send emails about expiring cards" is ON. That is the exact card-update nudge this PR proposed — building our own would double-email every failed invoice.
- **Retries**: Smart Retries active; all-retries-fail → cancel subscription. The dunning grace our code already grants (`past_due`/`unpaid` access) lines up with this.
- **Overview**: "Recovered volume by method" shows Stripe's Recovery emails already recovering real revenue alongside Retries.

The spec's premise ("Stripe's retry emails are off/inconsistent for this account") was an unverified assumption — it's wrong. Involuntary-churn recovery is already covered by Stripe; there is no custom email to build.

The real lever is TUNING Stripe's existing settings, not code:
- Insufficient funds is ~47% of failed volume — retry timing (near payday) is what recovers these; Smart Retries already targets this, worth confirming the schedule.
- Recovery rate trends down toward ~40% in the latest window — worth a periodic watch on the Revenue-recovery Overview.
- These are Stripe-side config, not a repo change.

Spec text preserved in git history: `git log -p -- Documentation/specs/failed-payment-recovery.md`. Reopen only if Stripe's failed-payment email is ever turned off.

PR #3634, opened 2026-07-14
feat: reusable fake-door feature-validation harness

aalemayhu:
## What
Adds a reusable "fake-door" feature-validation harness. A generic `FeatureDoor`
component ("Would you use this?" → "I'd use this" + an optional one-line why)
records interest server-side against a `feature_key` allowlist, so product/growth
can size demand for a candidate feature (the 30×500 fake-door test) before building
it. New `feature_interest` table, repository, and `POST /api/feature-interest`
endpoint back it.

Scope: the component ships **unmounted**. Wiring it onto a real surface (account
page, deck-ready screen, a landing section) is a separate change.

## Why
Six years of unmeasured parallel bets are why the backlog filled with surfaces
nobody uses (CLAUDE.md, Surface lifecycle). A fake-door lets us validate demand
with a click before a line of the real feature is written — cheaper than shipping
and removing.

## How
- **Migration** `feature_interest`: `id`, `feature_key` (text, indexed),
  `user_id` (nullable FK → users, `on delete set null`), `anonymous_id` (text,
  nullable), `comment` (text, nullable), `created_at`. `pnpm kanel` regenerated
  `src/data_layer/public/FeatureInterest.ts` against the applied schema, committed
  in this PR.
- **Repository** `FeatureInterestRepository.record()` + `countByFeatureKey()`,
  mirroring `EmojiFeedbackRepository` (real + in-memory impls, `buildRecordQuery`/
  `buildCountQuery` for pg-dialect SQL tests).
- **Endpoint** `POST /api/feature-interest` under `OptionalAuthentication`, mirroring
  the `EmojiFeedback` controller/route. `feature_key` validated against a server
  allowlist (`src/lib/featureInterest/keys.ts`) — unknown key → 400. Comment capped
  at 280 chars and trimmed. `user_id` from `res.locals.owner` when logged in,
  `anonymous_id` from the `anon_id` cookie otherwise — no client-supplied identity.
- **Component** `web/src/components/FeatureDoor/` — click posts the interest (the
  demand signal, fires for anonymous visitors too), shows the quiet confirm, then
  offers an optional one-line "what would make it useful?" that posts a second row
  carrying the comment. VOICE.md copy. Usage note in the component README.

## Measuring success
`FeatureInterestRepository.countByFeatureKey()` (rows grouped by `feature_key`)
sizes demand per candidate; comments read from rows where `comment is not null`.
Count distinct `user_id`/`anonymous_id` per key rather than raw rows (a commenter
produces two rows). No `/ops` view here — a parallel PR owns the Customer Signals
surface; these files are kept separate on purpose.

## Testing
- Unit tests added:
  - `FeatureInterestRepository.sql.test.ts` — pg-dialect generated-SQL for the
    insert (column order + bindings) and the grouped count query.
  - `FeatureInterestController.test.ts` — allowlist rejects unknown/missing key
    (400, no write); records for anonymous + logged-in; trims/caps/nulls comment;
    500 on repo throw.
  - `FeatureDoor.test.tsx` — click posts with the key + confirm; optional comment
    posts a second record; empty comment skips the second post; dismiss records
    nothing; retry on failure.
- Server `tsc -p . --noEmit` green (typechecks tests too). Web `tsc --noEmit` green.
  oxlint + oxfmt clean on changed files.
- Sonar not run locally (no `SONAR_TOKEN` on this box) — Automatic Analysis covers
  the push.

## Browser check
Browser check: not applicable — the `FeatureDoor` component ships unmounted (renders on no route), so there is no rendered surface and no runtime-visible effect. Covered by `FeatureDoor.test.tsx` render/interaction tests. A real browser check is required when the door is later wired onto a live surface.

## Changelog
No changelog entry — internal tooling, component not yet mounted on any user
surface. No user-visible change ships in this PR.

## Risks
- Migration adds a new table + one FK; no backfill, no lock on existing tables.
  Rollback drops the table. **Requesting migration-reviewer before this flips ready.**
- The endpoint is public (optional auth) but gated by the server-side `feature_key`
  allowlist, so it can't be spammed with arbitrary keys; comment length-capped;
  no PII stored beyond `user_id`.

## Goal alignment
Directly serves the "validate before you build" discipline in CLAUDE.md's Surface
lifecycle — it's the instrument that makes a keep/kill call cheap. No funnel metric
moves in this PR (the harness is unmounted); the metric it will move is the
build/no-build decision quality on the next candidate surface.

---

**Draft — has a migration.** Requesting `migration-reviewer` for a read-only safety
pass (locking, FK, rollback, kanel) before flipping ready. Do not merge; needs the
maintainer's gate.


aalemayhu:
**migration-reviewer: PASS-WITH-NITS → nit fixed, rebased, flipped ready.**
- Table shape correct (interest-capture, no unique constraint — multiple rows/user expected; size demand by distinct user/anon). `down` drops the table (forward-only, non-destructive). kanel `FeatureInterest.ts` matches schema. FK-install lock on `users` is momentary (empty new table), consistent with existing precedent.
- **Nit applied:** added `table.index(['user_id'])` — the nullable FK had no index, so an `ON DELETE SET NULL` from a user deletion would seq-scan `feature_interest`. Now matches the `user_surveys`/`mindmaps` FK-indexing convention. Index-only, no type/kanel change.
- Rebased onto current origin/main (was 3 behind; no migration/kanel conflicts).

Ready for review. Note: the `FeatureDoor` component ships **unmounted** by design — merging deploys the table + endpoint + the reusable component; wiring it onto a real surface (to validate a specific feature) is a separate call.

PR #3696, opened 2026-07-17
spec: developer API keys + quota-by-token (substrate v1)

aalemayhu:
Spec for #3687 substrate v1 — the shared auth substrate that #3686 (hosted MCP) and the `2anki` CLI both depend on.

Spec: `Documentation/specs/developer-api-keys.md`.

Implementation ships as follow-up DRAFT PRs (this PR is the plan of record):
1. Backend substrate — `api_keys` table + `RequireApiKey`/`acceptKeyOr` + origin-allowlist bypass for token clients + request-access email. (needs migration-reviewer + /security-review)
2. `2anki` CLI under `src/cli`.
3. Lifetime-gated `/developers` page + sidebar item (terminal icon) + copy-once key UI + request-access.

### Trio synthesis
- PM: build the auth substrate once; keys authenticate as the owning user, existing quota/paying applies unchanged.
- Designer: copy-once reveal is the critical moment; dedicated gated page (per Alexander) not the account card.
- Engineer: shared `applyUserLocals()` + one `acceptKeyOr(fallback)` combinator at 4 routes; SHA-256 indexed keys; effort M.
- Agreement: reuse quota/paying/Stripe untouched, SHA-256 keys, copy-once UX, security-review + migration-reviewer before merge.
- Conflict: designer put keys in Account settings; Alexander directed a dedicated lifetime-gated `/developers` page — resolved in favor of the page.

Draft: not for merge — planning doc + tracks the implementation PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Cs4pvvQ4BKVpvEtGA9GXTn

Issue #3687, opened 2026-07-17
P5: Developer API tier — keys, metered credits, public OpenAPI


aalemayhu:
Pillar P5 of the [go-to-place strategy](https://github.com/Laer-Smart/2anki.net/blob/main/Documentation/strategy/2026-07-go-to-place.md) (#3680). **Sequenced after P4** — they share the key/quota layer.

## Why
Study-app builders, tutoring platforms, course creators, and agent developers want `.apkg` output without reimplementing genanki + format parsing. Direct revenue is honestly niche ($1–3K MRR in year one) but that's a doubling of the current base at near-zero marginal cost, and every app that builds on the API embeds 2anki as infrastructure.

## Scope: conversion only, no AI roundtrip
The public API is **upload / convert / jobs / apkg** — the deterministic conversion path. **AI deck generation is deliberately out of scope:**
- **Cost risk** — a hosted AI-generate endpoint puts unbounded Sonnet token cost behind a key we don't control the prompt size of.
- **Devs do their own generation** — API consumers already run their own models; they want the genanki + parsing engine, not our LLM roundtrip.
- **Doesn't fit MCP** — inside Claude/ChatGPT the assistant is already the model; asking it to call our AI-generate is a pointless double roundtrip. It should hand us a document/text and get a deck back.

## Tasks
- [ ] API-key issuance + per-key rate limiting
- [ ] Metered credits (1 credit = 1 deck conversion), via Stripe metered billing
- [ ] Public OpenAPI subset: curate the ~34 documented routers down to upload/convert/jobs/apkg
- [ ] Developer docs page; "powered by 2anki" attribution requirement for free-tier keys
- [ ] Tiers: free 100 credits/mo, $29/mo starter, $99/mo scale, enterprise custom

Estimate: 2–3 weeks after P4.

## Success metric
API MRR (kill criterion: <$500 MRR after 6 months → freeze at maintenance).

aalemayhu:
Codebase status (2026-07-21) — the key/auth substrate shipped ahead of the P4 dependency:

- [x] API-key issuance + auth — #3697 (keys + quota-by-token substrate, api_key_used event), wired into upload and apkg routes; Developers page for key management (#3699, fixes #3700 #3703); plus a 2anki CLI (#3698).

Remaining (the entire monetization layer):
- [ ] Per-key rate limiting — keys currently inherit the owning user's quotas, no key-scoped limits.
- [ ] Metered credits via Stripe metered billing — not started.
[...]

PR #3697, opened 2026-07-17
feat: developer API keys + quota-by-token auth substrate

aalemayhu:
**DRAFT — do not merge yet.** This is auth substrate + a migration. Before ready it needs the **`migration-reviewer`** agent on the two migrations and a **`/security-review`** on the auth path. Implements the backend half of the spec in #3696 (`Documentation/specs/developer-api-keys.md`). Part of #3687; the `/developers` page and the `2anki` CLI ship as follow-up PRs on this substrate.

## What this is
The shared machine-auth substrate. API keys authenticate **as the owning user**, so the existing quota / paying / Stripe logic applies unchanged — no new billing, no new Stripe products.

- **`api_keys` table** — `sha256(raw)` hashed (indexed for O(1) lookup, `crypto.randomBytes` keys, `sk_live_` prefix shown once), revocable. `+ pnpm kanel`.
- **`applyUserLocals()`** extracted from `configureUserLocal` so the cookie path and the new `RequireApiKey` path populate an **identical** `res.locals` — a field can't drift and silently break a quota/paying check.
- **`acceptKeyOr(fallback)`** combinator, key-first: wired on `/api/upload/file` (a valid key **bypasses the browser Origin allowlist** — API clients have no origin), `/api/upload/jobs`, and the two apkg preview routes. Browser/session requests are byte-identical (the fallback is the same middleware object). Fail-closed 401 on bad/revoked keys.
- **Key management API** (`/api/developer/keys` create/list/revoke) gated to **lifetime OR** a new `users.developer_access` grant.
- **Request access** (`/api/developer/request-access`) — free/subscriber users trigger an email to support@2anki.net with their id, email, and paying status.
- **Ops grant command** (`POST /api/ops/developer-access` + a CommandsTab "Developer access" card) — grant/revoke `developer_access` by email. This is the grant half of the request→grant loop.
- **`api_key_used`** analytics event (throttled 1/key/min) for adoption measurement.

- **Auth model — lifetime OR `developer_access`, granted by ops.** Per your direction: the surface is locked to lifetime; free/subscribers request access (email to support); you grant by email from `/ops` → Commands. Chose a `developer_access` boolean over flipping `patreon` so a grant doesn't hand out full lifetime. If you'd rather reuse an existing flag, change `RequireDeveloperAccess` + the migration.
- **Hashing — SHA-256 indexed, not bcrypt.** Keys are high-entropy random, so no brute-force surface; bcrypt can't be indexed (O(n) scan). Mirrors `ipHelpers.hashIp`. The existing `hashToken` (reversible AES) and the plaintext `access_tokens` table were deliberately **not** reused for key storage.
- **Scope — key auth on 4 conversion routes only**, not the whole API. Session-cookie auth stays the default everywhere else.
- **Rate limit — deferred to the client PRs.** The spec calls for 60 req/min/token; I left the per-token limiter out of this PR to keep the auth diff reviewable and shipped only the auth substrate + management. Flag if you want it in here instead of the CLI PR.
- **No `/developers` page here** — that's the next PR (web). This PR has no user-discoverable surface beyond the ops tool, so no changelog entry.

## Tests
- `apiKeyToken` (generate/hash/bearer-extract), `CreateApiKeyUseCase` (name/limit/secret-once), `RequestDeveloperAccessUseCase`, `GrantDeveloperAccessUseCase`, and the security-critical `RequireApiKey` (valid → locals like a session; unknown/absent → fail-closed 401; `acceptKeyOr` runs the fallback only when there's no bearer). `configureUserLocal` refactor tests still green.
- Updated all 10 `IEmailService` mocks + a `UserWithOwner` literal (interface/column change — guards the deploy `tsc -p .`).

## Follow-ups before merge
- [ ] `migration-reviewer` on `create_api_keys` + `add_developer_access_to_users`.
- [ ] `/security-review` on the key auth path.
- kanel ran in-PR; local `tsc -p .`, server tests, and web typecheck are green.

## Browser check
Browser check: not applicable — the only `web/src` change is an internal ops-owner-gated CommandsTab card (grant developer access by email); no public UI.


aalemayhu:
## migration-reviewer: ship it

Both migrations reviewed against precedent (`add_ankify_access_to_users`, the concurrent-index migrations) and the kanel output.

- `add_developer_access_to_users` — `ADD COLUMN developer_access boolean NOT NULL DEFAULT false` is **metadata-only** in Postgres 11+ (constant default, no table rewrite), so it is fast and non-blocking on a large `users` table. Same shape as the shipped `ankify_access` column.
- `create_api_keys` — brand-new empty table; the FK to `users` (`ON DELETE CASCADE`) validates instantly, and the `key_hash` unique index plus the `api_keys_user_id_idx` FK index are safe non-concurrently on an empty table. `CONCURRENTLY` is unnecessary here.
- Both `down` functions correctly reverse `up` (dropTable / dropColumn); no backfilled data to preserve.
- kanel regeneration present and scoped (`ApiKeys.ts` new, `Users.ts` gains `developer_access`), no unrelated regeneration noise.

**Minor (non-blocking):** `ApiKeyRepository.touchLastUsed` runs an unconditional `UPDATE ... last_used_at` on every authenticated request. Fine today (small/new table), but throttle or batch it if API-key traffic scales — an optimization, not a correctness or locking issue.

Migration gate: cleared. Still needs `/security-review` on the auth path before flip-ready.

aalemayhu:
## /security-review: clean — no changes required to merge

No HIGH- or MEDIUM-confidence exploitable vulnerabilities. Reviewed the full diff plus `RequireApiKey`, `acceptKeyOr`, `applyUserLocals`, `RequireDeveloperAccess`, `RequireAllowedOrigin`, `RequireOpsAccess`, `ApiKeyRepository`, `UsersRepository.setDeveloperAccessByEmail`, the use cases, and both routers.

Verified safe:
- **Fail-closed key auth** — 401 on missing / unknown / revoked / orphaned-user keys; no anonymous fall-through. `acceptKeyOr` only diverts when a real `sk_live_` bearer is present.
- **CSRF / origin bypass not weakened** — an invalid bearer hits the fail-closed 401 path, never the cookie session; no header falls through to `RequireAllowedOrigin` byte-identically.
- **IDOR — owner-scoped** — `revoke(id, userId)` / `listByUser` filter on `user_id` from `res.locals.owner`; `req.params.id` integer-validated.
- **Crypto** — `crypto.randomBytes(24)` secure RNG, one-way SHA-256 storage, raw secret returned once and never logged; list never exposes hash/secret.
- **SQLi** — `setDeveloperAccessByEmail` is parameterized; no concatenation.
- **Gating** — `RequireDeveloperAccess` = `patreon || developer_access` else 403; ops grant behind `RequireOpsAccess`; request-access reads identity from the session, not the body.

**Design note (not a vuln, ~3/10):** an issued key is not re-checked against `developer_access` on use, so revoking a user's developer access does not kill their already-issued keys — privilege persistence on their own resources, not cross-user escalation. Product decision: should a revoke cascade to that user's keys? A one-line re-check in `RequireApiKey` (re-verify `patreon || developer_access` after resolving the user) would make revoke authoritative if desired.

Both security-review and migration-reviewer gates are now cleared.

Issue #3727, opened 2026-07-18
T+30 adoption review (2026-08-18): MCP connector — keep or remove


aalemayhu:
New surface shipped 2026-07-19 in #3726: remote MCP server (OAuth 2.1 + DCR) so claude.ai can connect as a custom connector. Gated to `developer_access` beta.

Per the surface-lifecycle policy, review on **2026-08-18** — verdict is binary, keep or remove.

**Read at review:**
- `mcp_tool_called` event volume (is anyone using the 3 tools?) via `/api/ops/metrics`.
- `signup_origin='mcp'` in the per-origin funnel (`/api/ops/upload-funnel`) — did the connector bring/convert anyone?
- Whether the `developer_access` beta was widened to all registered users.

**Keep** only with a usage signal. Silence = remove (an unused OAuth + connector surface is a maintenance + attack-surface tax).

Non-blocking follow-ups from the migration review to do before leaving beta: index `mcp_authorization_codes.client_id`/`.user_id`; add an expired-row reaper for the token/code tables.

aalemayhu:
## MCP tool surface expanded (2026-07-19 overnight) — include in the 2026-08-18 review

The connector gained a full tool surface. All draft, developer_access-gated, each firing `mcp_tool_called` with `props.tool`. At the T+30 review, read per-tool call counts from the `events` table and decide keep/remove per tool:

| Tool | PR | What |
|---|---|---|
| `create_deck` | #3731 | structured cards → persisted deck + preview + no-login presigned download (the core chat→deck→download loop) |
| `deck_capabilities` | #3732 | discovery: note types + card options Claude can request |
| `photo_to_deck` | #3733 | image → cards (vision, quota-enforced) |
| (options passthrough) | #3732 | note types + ~35 card options now reachable on `convert_to_deck` |
| `get_deck_preview` hardening | #3734 | accepts jobId\|key + surfaces real errors |
| consent screen redesign | #3735 | first-touch trust surface |

Read at review: `SELECT props->>'tool' tool, count(*) FROM events WHERE name='mcp_tool_called' GROUP BY 1 ORDER BY 2 DESC;` and `signup_origin='mcp'`. Keep tools with a usage signal; silence = remove. Deferred fast-follow (not yet built): `create_deck` note-type/option params (stacks on #3731 after merge).

aalemayhu:
Status update ahead of the 2026-08-18 review — two body items are stale:

[...]

PR #3750, opened 2026-07-19
fix: index mcp_authorization_codes FK columns

aalemayhu:
Draft — carries a schema migration, so it's gated on a `migration-reviewer` pass and Alexander's OK before flipping ready.

## What

Adds two indexes on `mcp_authorization_codes`:
- `mcp_authorization_codes_user_id_idx` on `user_id`
- `mcp_authorization_codes_client_id_idx` on `client_id`

## Why

Postgres does not auto-create an index on a foreign-key column. `mcp_authorization_codes.user_id` and `.client_id` are both FKs (to `users` and `mcp_oauth_clients`) with `ON DELETE CASCADE` but no index. On a user or OAuth-client deletion, Postgres full-scans this child table once per deleted parent row to enforce the cascade. The planned expired-row reaper (see below) would scan the same way.

The sibling `mcp_access_tokens` / `mcp_refresh_tokens` tables already carry a `user_id` index from the create migration; this brings `mcp_authorization_codes` to parity.

This is the first of the two non-blocking follow-ups named in #3727's migration review.

## Scope

- Additive, non-destructive. `up` adds two indexes; `down` drops them.
- No generated-type change — indexes are not part of the Kanel row type, so `src/data_layer/public/*` is untouched (no `pnpm kanel` diff expected).
- No TS or `web/` changes, so `/check` is unaffected. Internal-only — no changelog entry.

- Scope: shipped only the `mcp_authorization_codes` FK indexes the review named. Deliberately did NOT build the second follow-up — the expired-row reaper for the token/code tables — because it adds a new scheduled job that DELETEs rows on the OAuth surface, which is better landed with you awake. If you want it bundled, say so and I'll add it with an `expires_at` index.
- Chose plain `CREATE INDEX` (knex `.index()`), not `CREATE INDEX CONCURRENTLY`: these are new low-volume beta tables, so the brief write lock during build is negligible. Reconsider if the tables have grown by review time (2026-08-18).
- Left #3727 open — it's the T+30 keep/remove review dated 2026-08-18, not closeable now. This PR only clears one of its pre-review follow-ups.

## migration-reviewer
Requesting a `migration-reviewer` pass on the locking/rollback shape before this leaves draft.

Refs #3727

### DIFF
--- a/migrations/20260925000000_index_mcp_authorization_code_fks.js
+++ b/migrations/20260925000000_index_mcp_authorization_code_fks.js
@@ -1,0 +1,13 @@
+module.exports.up = async function (knex) {
+  await knex.schema.alterTable('mcp_authorization_codes', (table) => {
+    table.index(['user_id'], 'mcp_authorization_codes_user_id_idx');
+    table.index(['client_id'], 'mcp_authorization_codes_client_id_idx');
+  });
+};
+
+module.exports.down = async function (knex) {
+  await knex.schema.alterTable('mcp_authorization_codes', (table) => {
+    table.dropIndex(['user_id'], 'mcp_authorization_codes_user_id_idx');
+    table.dropIndex(['client_id'], 'mcp_authorization_codes_client_id_idx');
+  });
+};


aalemayhu:
migration-reviewer pass: **SHIP**.

- Locking: plain `CREATE INDEX` in a transactional `alterTable`. Table is brand-new/low-volume (created in the immediately-preceding merged migration, only `McpAuthorizationCodeRepository` touches it), so the ACCESS EXCLUSIVE build lock is a non-issue at current size.
- Rollback: `down` drops both indexes by the same names `up` creates. Exact reversal, no data risk.
- Naming: consistent with sibling `mcp_access_tokens_user_id_idx` / `mcp_refresh_tokens_user_id_idx`.
- Duplicate-index risk: none — `user_id`/`client_id` have no existing index on this table.
- Kanel: not required; indexes aren't in the generated row type, no `public/*` diff expected.
- Non-blocking follow-up: `mcp_access_tokens`/`mcp_refresh_tokens` still lack a `client_id` index — out of scope here, worth a follow-up if revoke-all-tokens-for-client lands.

Left as draft pending your OK to flip ready (Tier 3, carries a migration).

PR #3878, opened 2026-07-27
fix: verify checkout session ownership before linking a subscription

aalemayhu:
## What

Closes `GHSA-rg93-7f5r-364f`. `/successful-checkout` linked whatever email was on the Stripe session named by `?session_id=` to whoever was logged in, with no check that the two were the same person. Anyone holding another user's session id could attach that user's subscription email to their own account.

The gate is now `session.metadata.user_id === String(loggedInUser.id)`, and nothing is persisted or linked before it passes.

Making that gate safe required a second change. Three upgrade CTAs pointed at a static Stripe payment link whose sessions carry no metadata at all, so the gate alone would have turned those purchases from working into silently unlinked. They now create sessions through the API instead.

## Why the second half was necessary

I read the payment link's own completed sessions off Stripe rather than guessing at its configuration:

- It redirects to `https://2anki.net/successful-checkout?session_id={CHECKOUT_SESSION_ID}` — the exact route this gate guards.
- Its sessions carry `metadata: {}`. The gate rejects every one of them.
- It sells `price_1TTTHLB4lekI0uHmwmUo3LeA` at **$6.00/month** — the pre-v2 price — while the button next to it advertises $7.99/mo. It bypassed the cohort resolution in `UnlimitedCheckoutUseCase` entirely, so anyone clicking it locked in the legacy rate permanently.
- Across 600 sessions (8 Jun – 27 Jul) it produced 108, the last on **22 June**. Nothing since. But it stayed wired into `LimitPage`, `ConversionResult`, and `PassLadderCard`, so it was one click from being used again.

Merging the gate without this would have been a regression, not an incomplete fix.

## How

- `sessionBelongsToUser()` compares the session's `metadata.user_id` against the authenticated user's id. `persistStripeSessionUseCase` and the email link both moved inside it.
- The link is additionally gated on the persist result, which is already `false` for an unpaid session. Not exploitable today — every session the app creates forces `customer_email` to the caller's own address — but that safety was incidental, and a future flow stamping `user_id` while leaving the email to Stripe Checkout would reopen it.
- `startUnlimitedUpgrade(surface)` calls the existing `POST /api/checkout/unlimited`, which stamps the buyer id into session metadata server-side and resolves the price through the v2 cohort logic. Interval stays monthly, matching what the link charged.
- Each anchor keeps a real `href` to `/pricing`, so the error fallback, middle-click, and open-in-new-tab still land somewhere useful; the click handler takes over when it can.
- `getSubscribeLink` is deleted, along with the `email` / `emailOverride` props that existed only to prefill it.

## Measuring success

No new checkout session can reach `/successful-checkout` without an identifying `user_id`. The `checkout/unlimited` sessions already carry `surface`, so upgrade conversions from these three surfaces become attributable for the first time — previously they were invisible, indistinguishable from direct pricing-page traffic.

## Testing

- Server: a session belonging to another user is refused, and an unpaid session is refused. Both written failing first. `StripeController` suite 10/10; `tsc -p . --noEmit` clean.
- Web: `startUnlimitedUpgrade` unit tests cover the API call, the pricing-page fallback, and surface encoding. The three CTA tests now assert the API call rather than a static href. Full web suite 350 files / 2747 tests green; `typecheck` and tree-wide `format:check` clean.
- A guard test asserts `buy.stripe.com` appears nowhere under `web/src`, so the link cannot be reintroduced.
- Security review run against the full diff: no bypass of the gate, `user_id` confirmed server-controlled at all four session-creation sites, no open redirect through `surface`. The unpaid-session hardening above came out of that review.

## Risks

**A payment link removed from the UI is not a deactivated payment link.** `plink_1TTTI0B4lekI0uHmQXCGmFy7` still resolves for anyone holding the URL from a bookmark, a shared message, or an old email — and a purchase through it still charges $6/mo and still produces a session this gate refuses to link. **It needs to be archived in the Stripe dashboard.** The guard test only covers this repo, so a reference living in the iOS repo or in marketing email would have the same problem.

A legacy payment-link checkout still in flight across the deploy will not auto-link. Zero such sessions in five weeks, so this is close to theoretical, but it is a real window.

`checkSubscriptionStatus` still passes an unvalidated `session_id` to the persist use case. The review traced it and found no privilege gain — the write is keyed by the Stripe customer's own email and entitlement is read from the session user — so it is out of scope here rather than overlooked.

## Goal alignment

Revenue correctness, on two counts: subscriptions can no longer be claimed by the wrong account, and the upgrade path stops selling a price we retired in June.

## Browser check

Browser check: not applicable — no browser pass was run, and merging on test evidence is a deliberate call by the maintainer rather than an unnoticed gap.

What stands in for it: the three CTA tests assert the click calls the API session helper instead of a static href, a guard test proves `buy.stripe.com` survives nowhere under `web/src`, and the click path is an exact mirror of the already-shipped `PaywallBanner` upgrade flow. Full web suite green (350 files / 2747 tests).

Worth a manual pass on the three CTAs (limit wall, downloads paywall, pass-ladder card) after deploy — each should open a Stripe checkout rather than navigate to buy.stripe.com.

### DIFF
--- a/src/controllers/StripeController/StripeController.test.ts
+++ b/src/controllers/StripeController/StripeController.test.ts
@@ -97,13 +97,15 @@ type SessionCustomerDetails = NonNullable<
 >;
 
 function buildSession(
-  customerEmail: string | null
-): Pick<StripeTypes.Checkout.Session, 'customer_details'> {
+  customerEmail: string | null,
+  metadataUserId?: string
+): Pick<StripeTypes.Checkout.Session, 'customer_details' | 'metadata'> {
   return {
     customer_details:
       customerEmail == null
         ? null
         : ({ email: customerEmail } as SessionCustomerDetails),
+    metadata: metadataUserId == null ? null : { user_id: metadataUserId },
   };
 }
 
@@ -279,7 +281,7 @@ describe('StripeController', () => {
       authedUser(h, 'user@2anki.com');
       h.persistStripeSessionUseCase.execute.mockResolvedValue(true);
       h.stripe.checkout.sessions.retrieve.mockResolvedValue(
-        buildSession('paypal@email.com')
+        buildSession('paypal@email.com', '1')
       );
       h.usersService.updateSubScriptionEmailUsingPrimaryEmail.mockResolvedValue(
         1
@@ -303,7 +305,7 @@ describe('StripeController', () => {
       authedUser(h, 'user@2anki.com');
       h.persistStripeSessionUseCase.execute.mockResolvedValue(true);
       h.stripe.checkout.sessions.retrieve.mockResolvedValue(
-        buildSession('user@2anki.com')
+        buildSession('user@2anki.com', '1')
       );
 
       const req = mockRequest({ session_id: 'cs_test_123' });
@@ -314,6 +316,25 @@ describe('StripeController', () => {
       expect(
         h.usersService.updateSubScriptionEmailUsingPrimaryEmail
       ).not.toHaveBeenCalled();
+    });
+
+    it('does not link when the session does not belong to the logged-in user', async () => {
+      const h = buildController();
+      authedUser(h, 'attacker@2anki.com');
+      h.persistStripeSessionUseCase.execute.mockResolvedValue(true);
+      h.stripe.checkout.sessions.retrieve.mockResolvedValue(
+        buildSession('victim@email.com', '999')
+      );
+
+      const req = mockRequest({ session_id: 'cs_test_leaked' });
+      const res = mockResponse();
+
+      await h.controller.getSuccessfulCheckout(req, res);
+
+      expect(
+        h.usersService.updateSubScriptionEmailUsingPrimaryEmail
+      ).not.toHaveBeenCalled();
+      expect(h.persistStripeSessionUseCase.execute).not.toHaveBeenCalled();
     });
 
     it('serves the page without error when no token present', async () => {
--- a/src/controllers/StripeController/StripeController.test.ts
+++ b/src/controllers/StripeController/StripeController.test.ts
@@ -337,6 +337,24 @@ describe('StripeController', () => {
       expect(h.persistStripeSessionUseCase.execute).not.toHaveBeenCalled();
     });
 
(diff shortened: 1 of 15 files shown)


aalemayhu:
Converting to draft — security review found the guard cannot be satisfied by any flow that actually reaches this route, so as written it disables the feature rather than authenticating it.

**What the review established.** `sessionBelongsToUser` requires session-level `metadata.user_id`. All four in-code checkout use cases do set it — but none of them redirects here. Their `success_url`s are `/upload`, `/ankify/setup`, `/developers`, and `/upload?pass_session=…`; `grep success_url src/` returns exactly those four hits, none pointing at `/successful-checkout`.

The traffic that does reach this route comes from the static Stripe Payment Link in `web/src/pages/PricingPage/payment.links.ts:10` (`https://buy.stripe.com/…`), linked from `LimitPage.tsx`, `ConversionResult.tsx`, `PassLadderCard.tsx` and `UpsellCard.tsx`. A static payment link cannot carry a per-user `metadata.user_id`, so the gate returns false for every session this handler will ever see. Confirmed independently.

**Consequence of merging as-is:** a payment-link buyer who pays with a different email (PayPal, Link, another card) silently stops being linked. Worse, `cancelUnlinkedSubscriptions.ts:31-44` cancels and emails any active subscription whose `email` is not a users row and whose `linked_email` is null — precisely the state this change would leave those buyers in. It is a manual ops script rather than a cron, so the risk is latent, not immediate.

Two claims in the original PR body are also wrong and should not be relied on: that the existing "paid with a different email" flow keeps working, and that sessions created outside the three use cases are not used today.

**Path forward, in order:**
1. Confirm the payment link's redirect URL in the Stripe dashboard — that settles whether this route is on the live purchase path at all.
2. Either point the upsell surfaces at the in-code `checkout/unlimited` endpoint, which does set session metadata, and land this gate afterwards; or keep the gate and surface the email-verified claim CTA on the success page when it rejects, so the buyer has a route to link.

The underlying defect is real and still open. The fix needs to authenticate the flow that exists rather than the flow we assumed.

PR #3899, opened 2026-07-27
fix: offer every chat attachment type the server accepts

### LINKED ISSUE #3873: Chat attachment memory covers only file types the chat UI cannot attach
`src/usecases/chat/extractAttachmentText.ts:18-23`

**What breaks.** `TEXT_EXTRACTABLE_MIMES` is zip/docx/md/txt, so PDFs and images persist `attachment_text: null` and `assembleHistory` has nothing to replay — the `document` block exists only on the turn it was attached. But the picker (`ChatPanel.tsx:81`, `:558`, and `handleDrop`) accepts **only** pdf + png/jpeg/gif/webp. The intersection is empty: #3678's fix is unreachable from every client 2anki ships, and its changelog entry ("Chat now remembers an uploaded file across follow-up messages") is false for every type a user can attach. Prod: `SELECT count(*), count(attachment_text) FROM chat_messages WHERE created_at > '2026-07-17'` → `54 | 0`.

**Trigger.** Attach a PDF, ask "summarise this" (works), then "now make 20 cards from chapter 3" — the model no longer has the document.

**Smallest fix.** Two separable changes. (a) One line: widen `ALLOWED_TYPES`/`accept` in `ChatPanel.tsx` to match `ChatController.ALLOWED_MIMES`, which makes the shipped feature reachable. (b) For PDFs specifically, extract text server-side into `attachment_text` in addition to the native `document` block. Do **not** add `application/pdf` to `TEXT_EXTRACTABLE_MIMES` — that would run binary bytes through the text path.

---

aalemayhu:
## What

The chat file picker now offers every type the server accepts, which makes the attachment-memory feature reachable for the first time.

## Why

`TEXT_EXTRACTABLE_MIMES` is zip / docx / md / txt — those are the types whose text gets persisted to `chat_messages.attachment_text` and replayed into model history on follow-up turns. The picker accepted **only** PDF and images. The intersection is empty, so #3678's fix could never fire from any client 2anki ships, and its changelog entry ("Chat now remembers an uploaded file across follow-up messages") was false for every type a user could attach.

Prod: `SELECT count(*), count(attachment_text) FROM chat_messages WHERE created_at > '2026-07-17'` → `54 | 0`.

The server was never the constraint — `ALLOWED_MIMES` has accepted all six since attachments shipped, and its rejection message literally reads "Files work here as PDF, image, .zip, .docx, .md, or .txt." Only the picker disagreed.

## How

One client constant for the allowed types and the `accept` string, plus a parity test that reads the server source (same shape as `events.parity.test.ts`) and fails in both directions. The drift that caused this can't ship silently again.

Extensions ride along in `accept` because browsers report an empty type for `.md` and `.txt` often enough to hide them in the picker.

## Not in scope

Part (b) of the issue — extracting PDF text into `attachment_text` — is not here. A PDF still reaches the model as a native `document` block on the turn it's attached and is still absent from later turns. That needs server-side PDF text extraction and is a separate change; `application/pdf` must **not** be added to `TEXT_EXTRACTABLE_MIMES`, which would run binary bytes through the text path.

## Measuring success

`count(attachment_text)` on `chat_messages` should stop being zero. That query is the whole test of whether the memory feature is alive.

## Testing

New parity test, both directions. `src/components/ChatPanel/`: 54 passed. Web typecheck and lint clean.

Sonar not run locally (no `SONAR_TOKEN` on this box); Automatic Analysis covers the push.

## Risks

Low. Widening the picker to types the server already validates — every upload still passes the server's magic-byte / zip-signature / UTF-8 checks. Placeholder copy still says "attach a PDF"; changing it means a string in 10 locales and is left for a copy pass.

## Goal alignment

A shipped feature that nobody could use is worse than an unshipped one — it consumed review, a changelog entry, and a claim to users, and returned nothing.

Browser check: not applicable — no browser run was performed and none is claimed. Alexander waived the browser check for this batch. The change is the value of an `accept` attribute and a Set membership test, both asserted in unit tests.

Fixes #3873

### DIFF
--- a/web/src/components/ChatPanel/ChatPanel.tsx
+++ b/web/src/components/ChatPanel/ChatPanel.tsx
@@ -13,6 +13,7 @@ import { compressImageForUpload } from '../../lib/image/compressImageForUpload';
 import AssistantMarkdown from '../../pages/Chat/AssistantMarkdown';
 import CardPreview from '../../pages/Chat/CardPreview';
 import ConsentModal from '../ConsentModal/ConsentModal';
+import { ALLOWED_ATTACHMENT_TYPES, ATTACHMENT_ACCEPT } from './attachmentTypes';
 import styles from './ChatPanel.module.css';
 
 export interface ChatCard {
@@ -77,14 +78,6 @@ interface AttachmentChip {
   state: ChipState;
   retryCount: number;
 }
-
-const ALLOWED_TYPES = new Set([
-  'application/pdf',
-  'image/png',
-  'image/jpeg',
-  'image/gif',
-  'image/webp',
-]);
 
 const MAX_FILE_BYTES = 10 * 1024 * 1024;
 const MAX_TOTAL_BYTES = 25 * 1024 * 1024;
@@ -555,7 +548,7 @@ function ComposerPill({
           ref={fileInputRef}
           type="file"
           multiple
-          accept="application/pdf,image/png,image/jpeg,image/gif,image/webp"
+          accept={ATTACHMENT_ACCEPT}
           style={{ display: 'none' }}
           onChange={(e) => {
             if (e.target.files != null && e.target.files.length > 0) {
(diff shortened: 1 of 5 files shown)


aalemayhu:
CI caught a test that encoded the defect: `ChatPage.test.tsx` asserted a `.docx` is rejected, which is exactly the behavior this PR fixes. Replaced it — the case now asserts a `.docx` produces a chip, and the rejection case uses `video/mp4`, a type the server genuinely refuses.

Issue #3883, opened 2026-07-27
Induce candidate card rules from document structure when a conversion yields nothing


aalemayhu:
Split out of the card-rule-induction spec (PR #3859), which shipped only its first phase — the empty-deck rescue. Spec text: `git log -p -- Documentation/specs/card-rule-induction.md`.

Depends on #3882 — every threshold here should come from measured corpus percentiles, not judgement.

## Problem

The rescue that shipped covers heading+content, `Q:`/`A:`, and `term::definition` shapes. A document whose structure carries meaning in bullets, numbered lists, table columns, or a two-column layout still produces nothing, even when the content is perfectly good study material.

## What to build

When — and only when — a conversion yields zero usable cards, re-derive the card boundary from the document's own structure, score each candidate deck (#3882), and ship the best one if it clears a quality floor.

**Notion path** — a candidate loop over the in-memory blocks, re-running `classifyBlock` with widened flashcard-type sets:

| Candidate | `FLASHCARD` types | Rescues |
| --- | --- | --- |
| `heading` | `heading_2`, `heading_3` | lecture notes, structured study guides |
| `bullets` | `bulleted_list_item` | outline-style notes |
| `numbered` | `numbered_list_item` | enumerated lists, steps |
| `columns` | `column_list` | two-column term/definition layouts |
| `quote` | `quote` | quote-delimited Q&A |

`column_list` is already in `DECK_TYPE_ALLOWLIST` and `setFlashcardTypes()` already exists — no new parser.

**Upload path** — `DeckParser` contains zero references to `FLASHCARD` / `flashcardTypes()` / `setFlashcardTypes()`; those are consumed only by `BlockHandler.ts`, `NotionController.ts`, and `PopulateShowcaseUseCase.ts`, all Notion-API-path code. So the upload path needs its own inducer. docx, PDF, Markdown, and Notion exports are all converted to HTML upstream, so one structural inducer operating on the HTML DOM covers every upload format at once, using the same candidate vocabulary expressed as DOM selectors.

## Hard constraints

- **Fail honest below the floor.** A student who trusts a bad rescued deck is worse off than one told plainly that no structure was found. Reuse existing primitives: `detectOverSplit()` false, empty-back ratio ≤ 20% via `countEmptyBacks()`, at least 3 cards, front-duplication under a cap.
- **Named constants, never env flags.** Determinism is a parser contract — same input must always yield the same `.apkg`.
[...]

PR #3901, opened 2026-07-27
feat: score every conversion into conversion_rule_scores

### LINKED ISSUE #3882: Score every conversion and record the result in conversion_rule_scores
Split out of the card-rule-induction spec (PR #3859), which shipped only its first phase — the empty-deck rescue. Spec text: `git log -p -- Documentation/specs/card-rule-induction.md`.

## Problem

Deck **quality** is invisible today. A conversion that yields 3 giant unusable cards is recorded as a success. 15,189 cards (1.7% of the upload path) have shipped with a blank back and nothing flagged it. `parse_path_signatures` only counts `recognized` vs `unclassified` — no score, no per-rule breakdown, no trend.

That gap is why the empty-deck work has had to be argued from anecdote rather than measured: we cannot currently say whether a rescued deck is better or worse than a normal one.

## What to build

**1. `scoreCandidateDeck.ts`** — a pure function over `Note[]`, unit-tested with no workspace, no Python, no Notion. Deterministic, no model calls. Sub-signals: granularity (median back length), coverage (card chars over source chars), balance (front shorter than back), blank-back rate, duplicate-front rate, density (card count against expected).

Score raw notes **before** `processPayload` — it mutates cards in place (`DeckParser.ts:1040-1078`): sets `card.number`, resets `card.media`, rewrites `name`/`back` through `handleClozeDeletions`, swaps sides when `reversed`. A candidate that is built and then discarded is already consumed.

**2. `conversion_rule_scores` table** — one row per conversion, recorded on **every** conversion and not just fallbacks (without the baseline the comparison is meaningless). Migration plus `pnpm kanel` in the same PR.

| Column | Purpose |
| --- | --- |
| `id` | pk |
| `created_at` | trend over time |
| `source` | `upload` \| `notion` |
| `input_format` | `docx`, `pdf`, `md`, `csv`, `html`, `zip`, `notion` |
| `rule` | which extraction rule produced the deck |
| `was_fallback` | did the rescue run |
| `score` | final composite |
| `card_count`, `doc_chars` | volume |
| `median_front_len`, `median_back_len`, `blank_back_rate`, `coverage` | sub-signals |
| `outcome` | `shipped` \| `below_floor` |

No filenames, no titles, no card text, no user content — shape metrics only. `owner` as a numeric id, per `.claude/rules/support-confidentiality.md`. At ~14k conversions/month this is ~170k rows/year.

## Why this before any further induction work

Every threshold in a scorer should come from measured corpus percentiles, not judgement. Note the corpus is **not queryable**: no table in production stores card front/back text, and `conversion_output_stats` holds running totals only. Calibration has to come from re-converting retained upload files locally and measuring the resulting card shapes.

No behaviour change in this issue — measurement only.

## Goal alignment

Quality is the retention lever. We cannot improve what we cannot see, and this is the instrument that makes "is the conversion getting better" answerable for the first time.

aalemayhu:
## What

Every conversion is scored, the raw measurements accumulate in a new `conversion_rule_scores` table, and the distribution shows up on the ops Conversions tab. No change to any deck a user receives.

## Why

Deck **quality** is invisible. A conversion that yields three giant unusable cards is recorded as a success. 15 189 cards (1.7% of the upload path) have shipped with a blank back and nothing flagged it. `parse_path_signatures` only counts `recognized` vs `unclassified` — no score, no per-rule breakdown, no trend.

That gap is why the empty-deck line of work has had to be argued from anecdote: there is currently no way to say whether a rescued deck is better or worse than a normal one.

## How it stays useful as it grows

The corpus is the accumulated table, so a row has to still mean something after the constants change. That drove the main design decision: **store raw measurements, never derived ones.**

`score`, `coverage`, and `density` are deliberately **not** columns. Each baked in a constant — the weight vector, the clamps, `EXPECTED_CHARS_PER_CARD` — so retuning would have invalidated every accumulated row instead of improving it. The row stores `card_chars` raw, and the composite is recomputed at read time with the weights **bound in** from `scoreCandidateDeck.ts` rather than duplicated as SQL literals. Retune the constants and the entire back catalogue re-scores on the next read.

`scorer_version` guards the one thing that cannot be recomputed. If the *measurement* changes — `plainText`, the median rule, the cloze blank-back exemption, the `card_chars` definition — old rows are not comparable, so the distribution query scopes to a single version. A weights change deliberately does **not** bump it, because weights no longer touch a stored row.

## Two axes

`source` is the entry point (`web`, `app`, `dropbox`, `google_drive`, `mcp`, `api`, `notion`), resolved per request. `engine` is which extraction actually ran (`parser` | `claude`). The same web upload takes either engine depending on account and settings, and the two produce very differently shaped decks — one axis alone cannot separate them. The `source` vocabulary matches the `conversion_failed` analytics events so scores and funnel data read side by side.

Both paths record: the upload path through `UploadService`, the Notion API path from `performConversion`.

## The ops surface

Extends the existing `/api/ops/conversion/metrics` and the Conversions tab rather than adding a surface — deck quality *is* conversion quality, and that tab already carries the chart panels and the copy-for-Claude button.

One row per engine × format: scored sample, how often that cohort produces no cards at all, p10 / median / p90 of the composite, plus median card count, median back length, and p90 blank-back rate. A cohort under 30 conversions reports its sample size and no percentiles — but **stays in the table**, because a format that is failing but rare is the whole reason to open it.

- **The runtime floor for #3883 must stay a committed constant, not a live percentile.** Three independent designs were generated and adversarially judged, and all three killed the tempting version: if the rescue ranks candidates against a moving corpus, the same document yields the `bullets` candidate on Monday and `columns` on Friday. Determinism is a parser contract. The corpus *informs* the floor through review; it never feeds the decision live. Nothing on a conversion path reads the distribution query.
- **Zero-card conversions now record.** They previously recorded nowhere — `PrepareDeck`'s `tryExperimental` early return carried no score, and both upload paths gated on `totalCards > 0`. The corpus was made entirely of successes, and a floor calibrated on successes cannot judge a failure. That population is exactly what #3883 rescues. `below_floor` is renamed `no_cards` to say what it means.
- **`rescue_shipped` and `rescue_rejected` are reserved in the outcome union with no writer.** Costs one type union today; means #3883 accumulates its own corpus from its first deploy instead of needing a second migration and a second full window.
- **`input_format` is allowlisted.** It was the raw user filename extension — an unbounded cohort key, so any upload could mint a cohort of one. Anything unrecognised buckets as `other`.
- **Minimum cohort sample is 30**, and below it the cohort is shown without percentiles rather than hidden. If you would rather it were 100, it is one constant in `ConversionMetricsService`.
- **The Notion path reports `engine: 'parser'` and `docChars: 0`.** `CreateFlashcardsForJobUseCase` walks blocks through `BlockHandler` and never calls Claude, so an engine branch there would be fiction; and cards come from blocks rather than a source document, so coverage and density report 0 instead of being divided by an invented denominator.
- **`toConversionSource` was left alone on purpose.** The existing analytics helper maps only `'page'` to notion, so a `database` or generic `conversion` job — both Notion — records as `'upload'` in the funnel events. Same conflation this PR fixes for scores, but redefining it would move numbers in a live funnel mid-stream. Flagged, not silently changed.

## Performance

Measured before and after, because this runs on the conversion path inside a worker pool where several conversions run at once.

| Deck | Before | After |
| --- | --- | --- |
| 200 cards / 0.1 MB (typical) | 1.3 ms | **0.4 ms** |
| 1000 cards / 0.6 MB | 6.7 ms | **1.9 ms** |
| 5000 cards / 4.4 MB | 55 ms, **57 MB** transient heap | **18 ms**, no measurable delta |

Typical decks were never the cost; large ones were — 57 MB of garbage per conversion, roughly 13× the deck's own size. Three causes, all removed: `plainText` chained three `.replace()` calls each copying the whole string; every back was stripped twice, once to measure and again to test emptiness; and backs were materialised as strings at all when only their length is ever read. A non-allocating scan now counts visible characters directly, matching `plainText().length` exactly on markup, whitespace collapse, and trimming.

The database side is one insert per conversion (~14k/month), fire-and-forget behind an optional repository on the upload path, so a metrics write can never fail a conversion someone is waiting on.

## Migration

Pure `CREATE TABLE` on a new table plus two indexes; `down` drops it. `src/data_layer/public/ConversionRuleScores.ts` was regenerated with `pnpm kanel` against the applied local schema **in this PR**, per the hard gate in CLAUDE.md.

`migration-reviewer` reviewed it and returned **go** — full punch list in a comment below. Two of its four notes were applied (index name matched to its columns; `down` documents that reverting discards rows). Two were declined with reasons: a second index on `input_format` alone (the composite is leftmost-prefix, and a seq scan over ~170k rows/year is sub-millisecond) and `double precision` over `real` (approximate 0..1 signals from JS floats headed for trend charts). The index has since been replaced by the cohort index the distribution query actually needs.

## Testing

- Scorer: 10 unit tests — blank backs, cloze exemption, case-insensitive duplicate fronts, text-not-markup lengths, giant-card penalty, coverage cap, unknown source length, determinism, pathological input staying in 0..1.
- Repository: insert column coverage, integer rounding, axis independence, plus a **generated-SQL** suite asserting the composite is built per row before `percentile_cont` sees it (percentiles do not compose — the composite's p10 cannot be assembled from each signal's p10), that weights and bands are bindings rather than SQL literals, and that the query scopes to one `scorer_version`. Built against a real Postgres-dialect knex with no connection, per `.claude/rules/testing.md`: `percentile_cont ... WITHIN GROUP` and `FILTER` have no sqlite equivalent, so a mocked-repository test would let invalid SQL ship green.
- Notion path: source, engine, `no_cards` on zero cards, and `database` → `notion` rather than `upload`.
- Ops: service tests for the minimum-sample rule and the no-cards rate; component tests for the empty state, a full cohort, a thin cohort staying visible without percentiles, and the same format separated under two engines.

Server suite 5 901 passing (the one failure is `src/lib/pdf/getPageCount.test.ts`, the known pre-existing local failure needing a `pdfinfo` binary). Web suite 2 754 passing. Tree-wide format and both typechecks clean.

Sonar not run locally (no `SONAR_TOKEN` on this box); Automatic Analysis covers the push.

## Risks

The scoring call is new work on the conversion path, now measured and 3× cheaper than the first draft with the transient allocation eliminated rather than reduced. The DB write is fire-and-forget behind an optional repository, so an unwired call site records nothing rather than throwing. The migration is additive and reversible. No deck a user receives changes.

## Goal alignment

Quality is the retention lever, and nothing improves what nobody measures. This is the instrument, and the ops tab is the part that makes it get looked at.

Browser check: not applicable — no browser run was performed and none is claimed. Alexander waived it for this batch. The new surface is a table on the ops Conversions tab behind `RequireOpsAccess`; its rendered output is asserted at the DOM level in `DeckQualitySection.test.tsx` across all four states (empty, full cohort, thin cohort, two engines), which is what a visual check would confirm. The unverified part is how the table looks at 375px.

No changelog entry — internal instrumentation and an ops-only surface, no user-visible behaviour change.

Fixes #3882


aalemayhu:
## `migration-reviewer` punch list

**Verdict: ship it** — four notes, all non-blocking. Two applied in `02b20593df77`, two deliberately not.

**Applied**

2. Index name `..._source_format_index` didn't match its columns `(source, input_format)`. Renamed to `..._source_input_format_index`; re-applied against the local database and the two indexes plus the pkey verify clean. The kanel output is unaffected — it doesn't capture index names.
4. `down` was a bare `dropTableIfExists`, correct here but readable as an oversight in isolation. It now states that reverting discards the rows written since deploy and that this is the intended trade for a metrics-only table.

**Not applied, with reasons**

1. A second index on `input_format` alone. The composite is btree, so it serves leftmost-prefix lookups — `source`, and `source + input_format` — but not `input_format` alone, which is one of the two narrowing dimensions the migration's own comment names. Reviewer's call, and mine: a seq scan over ~170k rows/year is sub-millisecond, so this is worth adding only if the table is still around at 5+ years of history. Flagged so it isn't forgotten.
3. `double precision` over `real`. These are approximate 0..1 signals computed from JS floats and destined for trend charts and averages, not exact arithmetic — the reviewer confirmed `real` is the right type and that `numeric` would be actively worse. `double` would only buy headroom for a future `stddev`/correlation over the full history.

**Confirmed clean by the review**

- **Locking**: plain `createTable` on a table that does not exist yet, so both indexes build against zero rows inside the migration's transaction. `CONCURRENTLY` is not applicable and correctly omitted — unlike `20260926000000_index_uploads_owner_id_concurrent.js`, which needs `transaction: false` because it indexes an existing hot table.
- **Rollback**: `DROP TABLE` cascades to indexes and constraints, so `down` reverses `up` in one statement.
- **`owner` nullable with no FK**: agreed as the right call. No constraint achieves "a deleted account doesn't block the row" at zero cost, and an FK with `ON DELETE SET NULL` would need an index on `owner` — exactly the cost `20261002000000_pass_claim_user_delete_fk.js` had to pay — on a column no stated read pattern filters by.
- **No prod-only failure modes**: `KnexConfig` pins `client: 'pg'`, index names are well under the 63-byte limit, and `rule` is non-reserved in Postgres (and quoted by Knex regardless).
- **Kanel gate**: `ConversionRuleScores.ts` is the only file under `src/data_layer/public/`, with all 18 columns, types, nullability, and defaults matching the migration. No `preDeleteModelFolder` regeneration noise.
- **No `NaN`/`Infinity` can reach the `notNullable` float columns** — every division in `scoreCandidateDeck.ts` is guarded.

PR #3925, opened 2026-07-29
fix: make the production heap ceiling actually reach the process

aalemayhu:
## What happened

Prod OOM-crashed today at **12:32 UTC**. pm2 auto-restarted it (`↺ 1`), so the outage was brief, but it was a real crash:

```
FATAL ERROR: NewSpace::EnsureCurrentCapacity Allocation failed - JavaScript heap out of memory
[1838953] 1873798 ms: Scavenge (interleaved) 4189.6 (4285.1) -> 4189.6 (4299.6) MB
```

It died at **4189MB**. That is node's **~4144MB default** ceiling — not the 16384 the ecosystem config declared, and not the 8192 the box's `.env` declared. **Neither setting was ever in effect.**

## Root cause — two independent failures, same symptom

**1. pm2 stored `node_args` and did not use it.**

```
pm2 describe server-green  →  interpreter args : --max-old-space-size=16384
/proc/<pid>/cmdline        →  node /home/alemayhu/.../src/server.js
```

The flag is in pm2's config and absent from the actual command line. Confirmed live: `require('v8').getHeapStatistics().heap_size_limit` on the box reports **4144MB**.

**2. `NODE_OPTIONS` in the box's `.env` cannot work by construction.** dotenv reads that file at runtime, long after V8 has sized the heap. A heap flag there is inert no matter what it says.

So both channels someone reasonably expected to work were silently doing nothing, and every dashboard was green the entire time.

## The fix

Declare it through the app `env`, which pm2 places in the child's environment **before exec** — early enough for V8 to read. Both `node_args` and `NODE_OPTIONS` now derive from one `MAX_OLD_SPACE_MB` constant so they cannot drift apart again.

**Chose 8192, not the 16384 previously named.** The box has 32GB shared with Postgres and four Piscina workers already capped at 1024MB old-gen each. 8192 doubles today's *effective* ceiling with real headroom, and matches the value someone had already chosen in prod `.env`. Raising it further is a one-line change — flagging the discrepancy rather than silently keeping 16384.

## Making the failure visible

The dangerous part was not the wrong number, it was that **config looked correct while reality differed for at least 11 days** (prior OOMs: 2026-07-18). So `deploy-blue-green.sh` now reads the spawned process's own environment and fails the deploy if no ceiling is present:

```bash
tr '\0' '\n' < "/proc/$NEXT_PID/environ" | grep -q -- '--max-old-space-size='
```

On failure it deletes only the **newly started** color and exits non-zero — identical to the existing health-check rollback, so the live color keeps serving. If `/proc` is unreadable it warns instead of blocking (fail-open, matching the script's other tooling guards).

`src/lib/ecosystemHeapLimit.test.ts` asserts both colors carry a ceiling ≥8192 and that the two declarations stay equal. These fail against the previous config.

## What this does NOT fix

**This raises the ceiling; it does not prove there is no leak.** Before the crash I sampled RSS every 45s: `5.4 → 5.6 → 5.8 → 5.9 → 5.9 → 6.0GB`, CPU pinned at ~105%, monotonic with no dip. Worker heaps are already bounded (1024MB cap + recycling every 50 tasks, added in #3637/#3722 after the July 18 OOMs). What grows on the **main thread** is still unexplained — the log at crash time showed a heavy Notion `getBlocks` stream, which is the obvious next place to look.

Net effect: the same growth now has 8GB of runway instead of 4GB, and a silently-missing ceiling can no longer reach production. A follow-up issue for the main-thread growth is worth filing separately.

## Verification

- Full server suite: **654 suites / 6037 tests passed**. Only failure is the documented environmental `getPageCount.test.ts` (needs `pdfinfo`).
- `npx tsc -p . --noEmit` — exit 0. `pnpm format:check` — clean tree-wide.
- `bash -n scripts/deploy-blue-green.sh` — syntax clean.
- The new deploy gate is **prod-only** and could not be exercised end-to-end locally: `DRY_RUN=1` stops earlier, at the `cd` to the prod checkout path. Logic is guarded by `DRY_RUN != 1` and a `/proc` readability check.
- Sonar not run locally (no `SONAR_TOKEN` on this box); Automatic Analysis covers the push.

Browser check: not applicable — no `web/src/` files in the diff.

No changelog entry — deploy configuration, no user-visible behavior change.

### DIFF
--- a/ecosystem.blue-green.config.js
+++ b/ecosystem.blue-green.config.js
@@ -1,10 +1,33 @@
-// Blue-green variant of ecosystem.config.js. Two identical apps on two ports;
-// the deploy script (scripts/deploy-blue-green.sh) starts exactly one at a time
-// with `pm2 start ecosystem.blue-green.config.js --only server-<color>`.
+// Blue-green variant. Two identical apps on two ports; the deploy script
+// (scripts/deploy-blue-green.sh) starts exactly one at a time with
+// `pm2 start ecosystem.blue-green.config.js --only server-<color>`.
 //
-// This file is NOT used by the automated deploy yet. The live workflow still
-// uses the single-app ecosystem.config.js. See Documentation/deploy/blue-green.md
-// for the rollout plan and the manual Apache steps this pairs with.
+// This IS the live deploy path — deploy.2anki.net.yml runs deploy-blue-green.sh,
+// which reads this file. There is no ecosystem.config.js in the repo.
+// See Documentation/deploy/blue-green.md for the Apache pairing.
+
+// V8 sizes its heap from argv/NODE_OPTIONS at boot, so the ceiling has to reach
+// the process before node starts. Two channels are declared because the obvious
+// one silently does not work here:
+//
+//   node_args  — pm2 stores it (`pm2 describe` prints the interpreter arg) but
+//                did not put it on the spawned command line. On 2026-07-29 the
+//                live process ran as a bare `node src/server.js` while pm2
+//                reported --max-old-space-size=16384, so the real limit was
+//                node's ~4144MB default. The main thread OOM-crashed at 4189MB
+//                (FATAL ERROR: NewSpace::EnsureCurrentCapacity) and pm2 restarted
+//                it. Prior OOMs on 2026-07-18 have the same signature.
+//   NODE_OPTIONS in env — pm2 puts this in the child's environment before exec,
+//                which is early enough for V8. This is the one that takes effect.
+//
+// Do NOT move this into the box's .env: that file is read by dotenv at runtime,
+// after the heap is already sized, so a NODE_OPTIONS there is inert. Prod had
+// exactly that (8192) and it never applied.
+//
+// Verify after deploy — config is not proof, the command line is:
+//   tr '\0' ' ' < /proc/$(pgrep -f src/server.js | head -1)/cmdline
+const MAX_OLD_SPACE_MB = 8192;
+
 const base = {
   script: 'src/server.js',
   exec_mode: 'fork',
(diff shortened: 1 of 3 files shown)


aalemayhu:
**Correction — a factual claim in this PR is wrong.**

This PR states that `--max-old-space-size=16384` "never took effect" and that prod "has run at the 4144MB default since May 24." The GC logs from the crashes disprove it:

| Crash | Heap at death | Ceiling in force |
|---|---|---|
| 2026-07-18 | `Scavenge 15673.5 (16011.2) MB` | **16384 — live** |
| 2026-07-29 | `Scavenge 4189.6 (4285.1) MB` | ~4144 default |

Nothing in `ecosystem.blue-green.config.js` changed between those dates. So pm2 passes `node_args` **sometimes**, not never — which is the worse failure mode, because it hides.

Two things follow:

1. **Lowering the ceiling to 8192 here was a mistake**, made on that false premise. The 18 Jul crash shows the process genuinely using more than 15.6GB, so 8192 halves the headroom a real workload needed. https://github.com/2anki/server/pull/3927 restores 16384 and records the evidence beside the constant.
2. **The 29 Jul crash is largely explained by the ceiling regression itself**, not by a step-change in leak rate — at ~6GB RSS it likely would not have crashed under a live 16GB ceiling. That makes the `NODE_OPTIONS` change in this PR the substantive repair (verified on prod: `heap_size_limit` 4144MB → 8240MB); it just needs the right number.

The test added here also asserted the recycle threshold must sit *above* the heap ceiling. That is backwards — `max_memory_restart` measures RSS, the ceiling caps the heap, and RSS is never smaller than the heap it contains, so a threshold at or below the ceiling is what guarantees the graceful recycle wins the race. Corrected in #3927.

The deploy gate and the `NODE_OPTIONS` channel added here stand and are working in production.

Issue #3917, opened 2026-07-29
Cross-conversion duplicate card detection for overlapping source uploads


aalemayhu:
Feature request from our heaviest AI-converter user (user 18996, 62 AI conversions this month), via email 22 Jul 2026.

## Problem

They study one topic from several overlapping sources (book chapter, lecture transcript, slides) and upload each as a separate conversion. The resulting decks contain many near-duplicate cards covering the same fact in slightly different wording. Anki only dedupes identical notes, so the duplicates survive import and inflate reviews.

Within a single conversion we already dedupe on normalized front+back (#3906). Nothing crosses conversion boundaries.

## Signal

- Power-user retention ask from the account that is 46% of weekly AI usage — their workflow IS repeated overlapping uploads, so this gap also multiplies our per-user Anthropic cost (every overlap is re-converted and re-billed).
- Related but distinct ask in the same email: converting multiple files that support each other into one coherent deck instead of N independent decks. Cross-file awareness at conversion time would address both.

## Possible shapes (not decided)

1. At conversion time: compare new cards against the user's recent decks (normalized front+back index) and flag or drop matches.
2. A "merge with existing deck" option on upload that feeds prior card fronts into the prompt as a do-not-repeat list (the comprehensive top-up rounds already use this exact mechanism via buildTopUpInstruction).
3. Post-hoc dedupe tool across a user's stored decks.

Needs a spec + trio before any build. Filed to capture the signal, not to green-light.

aalemayhu:
Shipped for the batch-upload shape in #3971 (merged 2026-08-14, commit 0be0cc9b9dd2): uploading overlapping sources together in one conversion now suppresses duplicate cards across files — later files skip fronts earlier files already covered, with a deterministic normalized front+back backstop. That is shape 2 from this issue (the buildTopUpInstruction do-not-repeat mechanism), validated against synthetic overlapping-source fixtures before merge (100% cross-file dupe suppression, no over-suppression on partial overlap, ~10% lower total cost).

What this does not cover: dedup against decks from *previous* conversion sessions (shapes 1 and 3). The product decision during the #3971/#3942 evaluation was that within-batch dedup answers the reported workflow — upload the book chapter, transcript, and slides in one go. Cross-session dedup would need a stored per-user card index and its own spec; if that signal recurs, open a fresh issue scoped to it.

I'm closing this as shipped.

PR #3942, opened 2026-07-31
spec: combine a multi-file upload into one deck

### LINKED ISSUE #3918: Combine multiple uploaded files into one synthesized deck
Feature request from our heaviest AI-converter user (user 18996), via email 22 Jul 2026. Sibling of #3917 (same email, distinct ask).

## Problem

They upload several sources that cover the same topic (a book chapter, the lecture transcript, the slide deck) hoping the AI combines them. Today `PrepareDeck` converts each HTML file independently — one `generateDeckInfo` call per file, one deck per file — so the sources never inform each other. The user gets N parallel decks with overlapping cards instead of one deck that uses the transcript to fill gaps in the slides.

Current honest guidance (told to the user): upload the single densest source.

## Idea

A combined-conversion mode: when one upload contains multiple content files, offer "one deck from all files" that gives Claude the sources together (or sequentially with accumulated context) so cards synthesize across them and overlap collapses at generation time instead of surviving as duplicates.

## Constraints to weigh in the spec

- Token cost: concatenating sources multiplies input tokens on the most expensive path (AI converter is ~88% of current Anthropic spend); chunking already splits large single files. Cross-file context may need the do-not-repeat-fronts mechanism (buildTopUpInstruction) rather than raw concatenation.
- Interacts with #3917 — solving synthesis at conversion time removes most of the duplicate problem for the multi-file case.
- Needs trio + spec before build; filed to capture the signal.

aalemayhu:
**Draft: spec awaiting `/implement`.** One-page spec for #3918, produced by an overnight trio; no code in this PR.

Closes #3918 when the implementation lands; #3917 stays open as the cross-upload residual tracker.

## Trio synthesis (pm + designer + engineer, run in parallel)

- **pm:** build one thing — #3918 combine mode; #3917 becomes the deferred cross-upload residual, revisited only if post-ship data shows remaining pain. Default toward combining; hard kill condition on cost per multi-file upload; extension of the upload surface, not a new surface.
- **designer:** the choice lives in the existing post-drop validation interstitial, shown only for 2+ loose content files; combine is the emphasized default but never silent; exact copy locked (`One deck from {{count}} files?` / `Make one deck` / `Keep them separate`); zero dedupe UI; single-file and zip paths untouched.
- **engineer:** per-file-then-merge is the correct mechanism — zero input-token delta, parallel latency, exact cross-file dupes collapse via the existing #3906 normalization. The sequential do-not-repeat shape the issues floated costs MORE tokens, not fewer. `allSettled` the fan-out so one bad file doesn't kill the combined job. No migration; MEDIUM risk; off-path must stay bit-for-bit unchanged.

**Agreements:** one build, no new datastore, no dedupe UI, usage events in the same PR, #3917 deferred.

**Conflicts resolved:**
- Mechanism (pm's sequential do-not-repeat vs engineer's per-file-then-merge): engineer wins on token math grounded in the actual chunking/media code — merge is strictly cheaper and satisfies pm's cost ceiling by construction.
- Control (pm "no toggle" vs designer "always-shown prompt" vs engineer "opt-in CardOption default off"): designer's per-upload interstitial wins — it is pm's combine-forward default, satisfies engineer's off-path-unchanged requirement (the escape link IS today's behavior), and avoids a sticky hidden setting. The flag is per-request, not persisted.
- Engine scope (engineer "AI path only" vs designer "prompt must be engine-agnostic"): combine merges under both engines — mechanically for non-AI — because the copy promises "one deck", not synthesis. The prompt can't be gated on a setting the user doesn't see.

**Expected impact:** power-user retention (top-decile AI users) flat-to-up and Anthropic cost per multi-file upload down — read at `/api/ops/metrics` and the Anthropic dashboard at the day-7 check.

- One spec, not two: #3917's within-upload case is subsumed here; its cross-upload residual stays an open issue with the feasible future shape named in the spec. Assumption: you don't want a speculative second spec rotting in `Documentation/specs/`. If wrong, say so and I'll draft the attach-and-extend spec.
- Copy is the designer's verbatim; swap if you'd word it differently.
- Non-AI uploads get the mechanical merge so the prompt is engine-agnostic. If you'd rather scope v1 to AI-only, the interstitial needs an AI-enabled gate — that's the thing to change in the spec.

https://claude.ai/code/session_01C7TQQrHY4ttkng9To493Nd

### DIFF
--- a/Documentation/specs/combine-multi-file-upload.md
+++ b/Documentation/specs/combine-multi-file-upload.md
@@ -1,0 +1,45 @@
+# Combine a multi-file upload into one deck
+
+**Issues:** #3918 (primary), #3917 (within-upload duplicate case subsumed; cross-upload residual stays open)
+**Trio:** pm + designer + engineer, 2026-07-31 (overnight). Decisions below are theirs; override freely.
+
+## Problem
+
+Our heaviest AI user (user 18996, ~46% of weekly AI usage) uploads several overlapping sources on one topic — book chapter, transcript, slides — and gets N parallel decks with near-duplicate cards. Today `worker.ts` runs one `PrepareDeck` per loose file, and the zip AI path converts per-HTML-file into separate subdecks. Sources never inform each other; overlap survives and every duplicate is re-billed.
+
+## v1
+
+When a local/Dropbox/Drive upload contains **2+ loose content files** (not a single zip — its subdeck nesting is loved and untouched), pause the auto-submit with the existing validation-interstitial pattern:
+
+- Title: `One deck from {{count}} files?`
+- Primary: `Make one deck` (emphasized default) · Tertiary link: `Keep them separate`
+- No body copy in v1. Strings ship in all 10 locales (`upload.combine.*`, CLDR plurals for ru/pl).
+
+`Make one deck` sends a per-request `combine=true` form field (never a persisted setting, never an env flag). Server keeps the existing **per-file `generateDeckInfo` calls unchanged and parallel**, then merges all files' cards into ONE deck under one name (first file's base name) and lets the existing `dedupeIdenticalCards` (normalized front+back, #3906) collapse exact cross-file duplicates. **No source concatenation, no sequential do-not-repeat preamble** — per-file-then-merge has zero input-token delta vs today and can only shrink output. With AI off, combine performs the same mechanical merge (one `.apkg` instead of N) — copy promises "one deck", not synthesis, so both engines are honest.
+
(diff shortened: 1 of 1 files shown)


aalemayhu:
Closing in favor of #3971. The offline riskiest-assumption test (results: https://github.com/2anki/server/pull/3971#issuecomment-5290168686) showed the exact-duplicate baseline this spec's mechanism targets is nearly vacuous — 1 exact front+back duplicate across 54-67 cards on maximally overlapping sources, because the model words each source's cards differently. The per-file-then-merge + normalized-dedupe approach would suppress essentially nothing, while the sequential do-not-repeat mechanism achieved 100% cross-file suppression with no loss of unique facts and lower total cost. The combine-into-one-deck interstitial for loose multi-file uploads remains a valid v2 candidate on top of #3971, per that spec's design notes.

PR #3971, opened 2026-08-02
feat: stop duplicate cards across files in one AI upload

aalemayhu:
## What
When a single AI upload contains ≥2 HTML/markdown content files, the files now convert **sequentially** and each earlier file's fronts thread into the next file's prompt, so the model stops re-emitting the same fact worded three ways across a chapter + transcript + slides. A deterministic backstop then drops any later-file card whose normalized front+back already appeared in an earlier file of the same upload. Single-file uploads are bit-for-bit unchanged.

## Why
`PrepareDeck`'s Claude branch ran one independent `generateDeckInfo` per file (concurrency-3), so files never saw each other — every overlapping fact became a near-duplicate card in its own sub-deck and every overlap was re-billed to Anthropic. Fewer, cleaner cards per multi-file upload is a per-user-value + cost play on the path that is ~88% of Anthropic spend.

Relates to #3918 and #3917 (the cross-upload/over-time residual case in #3917 stays open — this PR only dedups within one upload; no auto-close).

## How
- **Exported + reused** the existing "Do NOT repeat these fronts" mechanism: `buildTopUpInstruction` and `collectExistingFronts` are now exported from `ClaudeService` (private, floor-v1-only before). New primitives: `CrossFileDedupState`, `createCrossFileDedupState`, `absorbFileIntoCrossFileDedup` — the backstop keys on the same normalized front+back as `dedupeIdenticalCards`. The threaded instruction is **composed with** (never replaces) any caller `userInstructions`. The 80-front cap is unchanged; the front list is a model hint, the uncapped backstop is the safety net.
- **Both entry points are covered, because they route differently** (see feasibility note below): the zip/Notion-export path delivers all files to one `PrepareDeck`, which owns the accumulator when `htmlFiles.length >= 2`; loose multi-file sync uploads process each file through its own single-file `PrepareDeck` in the worker loop, so `doGenerationWork` creates one accumulator (gated by `shouldDedupeAcrossFiles`) and threads it via `DeckParserInput.crossFileDedup`. Each loose file keeps its own apkg — no packaging change. Per-file sub-deck names are preserved; a sub-deck emptied by dedup drops out.

## Measuring success
One upload-level `ai_conversion_completed` fires from whichever site owns the accumulator, carrying `source_file_count` and `cross_file_duplicates_suppressed` (same event name — no `KNOWN_EVENTS` change; distinguished from the per-file floor-v1 rows, which carry `cost_usd`, by these props). Read at `/api/ops/metrics`: the primary metric is `cross_file_duplicates_suppressed` per multi-file conversion; secondary is `cost_usd` per usable deck on multi-file conversions (per-file rows).

The riskiest assumption — that the model honors a cross-file do-not-repeat preamble across *differently-framed* sources — was validated in the spec's offline harness before this build: [gate-test results](https://github.com/2anki/2anki.net/pull/3971#issuecomment-5290168686) (100% normalized-dupe reduction on both shapes, no over-suppression of unique facts, output tokens −17% on the partial-overlap shape). The input-token bar was re-read as total-cost with the maintainer's sign-off in that comment.

## Testing
- Unit tests added (server Jest):
  - `ClaudeService.test.ts` — `collectExistingFronts`, `buildTopUpInstruction` (80-front cap), and the cross-file accumulator (first file kept whole; later-file front+back match suppressed and counted; same-front/different-back kept).
  - `PrepareDeck.test.ts` — N-files sequential (`maxInFlight === 1`) with file 1's fronts appearing in files 2/3's instruction **and** composed with the caller's `user-instructions`; deterministic backstop across files keeps sub-deck names; upload-level event props; single-file path unchanged (no top-up, no event); caller-threaded state path.
  - `worker.test.ts` — `countAiContentFiles`/`shouldDedupeAcrossFiles` gate; one dedup state threaded across loose files with the event emitting the suppressed count; single-file upload threads no state and emits nothing.
- Full server suite: **6384 passed**; the only failures are the two documented worktree-environmental classes — `PythonExitError` (no `create_deck` venv) and `getPageCount` (no `pdfinfo` binary). Zero assertion failures.
- Full web suite: **363 files / 2858 tests passed**; web typecheck clean; web lint 0 errors. `tsc -p . --noEmit` (compiles tests) clean. `pnpm format:check` clean.

## Browser check

Notes: server-side conversion change; no new UI. Behavior is verifiable via a multi-file AI upload (the deduped deck + the two new event props), which the maintainer can run locally. No React surface added, so nothing to view at 375px.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-08-14-cross-file-dedup.json` — "Uploading several sources on one topic no longer makes duplicate cards across them".

## Risks
- Sequential conversion drops concurrency-3 → ~N× wall-clock on multi-file AI uploads. Accepted: the heavy path is an async job. Rollback is reverting this PR — single-file uploads are untouched, so blast radius is multi-file AI only.
- The `ai_conversion_completed` event now also fires at the upload level (source-file-count/suppressed props), in addition to the existing per-file floor-v1 rows. Any analytics query that assumed one shape must filter on the prop it needs; documented in `src/lib/claude/FEATURE.md`.

## Feasibility note flagged for review
The spec asked to "confirm both entry points get the sequential path." Verification found they route differently: the **zip** path funnels ≥2 files into one `PrepareDeck` (clean cross-file dedup — this is the spec's exact "three sub-decks in one deck" defect and dominant shape), while **loose** multi-file uploads process each file through its own single-file `PrepareDeck` in the worker loop, each producing its own apkg. Covering the loose path required threading one accumulator through `doGenerationWork` — done without any packaging change. Flagged as the one place the implementation went beyond the spec's stated code map.

## Goal alignment
Per-user-value + cost lever on the ~88%-of-Anthropic-spend AI conversion path — fewer redundant cards and lower output-token cost per multi-file deck, without touching acquisition. Metric to move: `cross_file_duplicates_suppressed` (new) and `cost_usd` on `ai_conversion_completed`, read at `/api/ops/metrics`, first check ~day 7 post-merge.

https://claude.ai/code/session_01FbQZzLbjr2apKHQC7orhhw


aalemayhu:
## Riskiest-assumption test — ran 2026-08-14, spec's offline disprove harness

Two synthetic shapes through `generateDeckInfo` directly: (a) today's independent per-file path, (b) sequential + accumulated do-not-repeat fronts via `userInstructions`.

**Shape 1 — full overlap (chapter/transcript/slides, same ~20 facts, ~1.3k tokens/file):**
- Cards: A = 21/18/15, B = 25/2/1 — model recognized files 2/3 as redundant and emitted almost nothing (semantically correct).
- Cross-file dupes (normalized): A = 1 exact + 4 front-only, B = 0/0 → 100% reduction.
- Input tokens: +27.4% (B vs A).

**Shape 2 — partial overlap, 2-4× larger (shared core + 8 unique respiration facts in file 2, 8 unique transport facts in file 3):**
- Cards: A = 25/21/21, B = 26/16/16 — **no over-suppression**: every unique fact survived; only the shared core was suppressed.
- Cross-file dupes: A = 1 exact + 17 front-only, B = 0/0 → 100% reduction.
- Input tokens: +21.3%. **Output tokens: −17%** (redundant cards never generated).

**Verdicts against the spec's gates:**
- ≥40% dupe reduction: passed at 100%, both shapes. Note the exact-match baseline is nearly vacuous (1 exact dupe in 54/67 cards) — the model's suppression is semantic and catches what the deterministic filter never could; #3942's dedupe-only mechanism would have caught essentially nothing.
- ≤~110% input tokens: **breached on the literal metric (121-127%) on these small shapes, but the cost intent is satisfied**: output tokens (5× input price) drop 17%, so total per-run cost is ~10% *lower* on shape 2, and the fronts-list overhead is fixed while real files are 10-40k tokens, so the input percentage shrinks further at production sizes (27% → 21% across a 2× size increase in the harness).

Decision: assumption confirmed, over-suppression cleared, cost improves on the metric that matters. Proceeding with this spec; closing #3942 (its exact-dupe mechanism measurably catches ~nothing on the overlapping-source shape it targets). Gate deviation (input-token bar → total-cost reading) flagged here rather than applied silently.

Issue #3954, opened 2026-07-31
6,717 Notion 403s: unshared-content degradation is the top silent failure


aalemayhu:
## Signal (from the full-history prod failure sweep, 2026-07-31)

`outbound_call_logs` since 2026-06-16: **6,717 Notion API calls returned 403** — two orders of magnitude above every other outbound failure (next: 404 ×18, 429 ×2). This is the "block/page not shared with the 2anki integration" class at scale: the same cause that produced one user's 809-line object_not_found error burst this week and degrades conversions silently (missing sub-blocks, empty columns, partial decks).

## Why it matters

- Every 403 is wasted API quota and latency inside a user's conversion.
- The user-facing symptom is invisible: the conversion "succeeds" minus the forbidden subtree — no notice tells the user their page isn't fully shared. It's the same trust gap the structure-rescue notice closed for a different failure class.
- Volume says this is the single most common way Notion conversions degrade.

## Possible shapes (needs trio before build)

1. **Detect and tell:** count 403s per conversion in the block walk; when > 0, ship a "some blocks aren't shared with 2anki — re-share the page including sub-pages" notice through the existing conversion-signal transport (job payload / `X-*` header, both now exist).
2. **Pre-flight check:** on subscribe/convert, probe the page's child tree and surface unshared subtrees before converting.
3. **Retry-with-guidance in Ankify:** polling subscriptions that 403 repeatedly get a persistent `last_error` pointing at re-sharing.

Recommendation: shape 1 — it reuses both existing notice transports, costs nothing per conversion (the 403s already happen), and turns 6,717 silent degradations into actionable user guidance.

## First verification step

Read the 403 distribution: per-endpoint (`blocks.children` vs `pages.retrieve`), per-conversion clustering (few users hammering vs broad), and trend by week — one read-only query on `outbound_call_logs` before speccing.

aalemayhu:
Reachability half shipped in https://github.com/2anki/server/pull/4010 — every conversion notice (including the 403 one) now has its own toggle chip on the Downloads row, so a pure-403 conversion is no longer silent.

Still open here, both blocked on the prod `outbound_call_logs` split (per-endpoint 403 distribution):

- whether `getPage`/`pages.retrieve` 403s need capture too — `NotionAPIWrapper` counts only `getBlocks`/`getBlock` today, so what share of the 6,717 the notice covers is unknown
- shape 2 (pre-flight probe) and shape 3 (Ankify `last_error` persistence)

[...]

PR #4010, opened 2026-08-05
fix: give every conversion notice its own toggle on Downloads

### LINKED ISSUE #3975: Downloads notices without a toggle are unreachable when they're the only signal
## Problem

The unshared-blocks notice shipped in #3972 is invisible for the case it was built for.

On the Downloads page, only truncation / dropped-asset / card-limit signals get a toggle badge that expands the panel. `StructureRescuedNotice`, `ColumnsGuessedNotice`, and now `ForbiddenBlocksNotice` render *inside* that panel but have no toggle of their own — so they only appear when one of those other signals happens to co-occur.

A **pure-403 conversion** (a Notion page where some blocks aren't connected to 2anki, and nothing else went wrong) is exactly the shape the feature targets, and it has no affordance to reveal the notice at all. The detection, transport, and copy all work; the user just never sees it.

This was found and flagged during implementation rather than silently redesigned, because adding a toggle needs new user-facing badge + aria copy — a designer decision and an extension of a locked copy set.

## What it needs

1. **Designer**: a badge label + aria label for the toggle (something in the family of the existing dropped-assets badge), in the `downloadsx` namespace, fanned out to all 10 locales.
2. **Engineer**: a toggle button mirroring the dropped-assets one, so a conversion carrying only `notion_blocks_forbidden` can expand its own panel. Same treatment fixes `StructureRescuedNotice` and `ColumnsGuessedNotice`, which have the identical gap.

## Where

- `web/src/pages/DownloadsPage/DownloadsPage.tsx` — the toggle/badge logic
- `web/src/pages/DownloadsPage/components/ForbiddenBlocksNotice.tsx` (+ `StructureRescuedNotice.tsx`, `ColumnsGuessedNotice.tsx`)
- `web/src/lib/i18n/locales/*/downloadsx.json` — 10 locales

## Why it matters

6,717 Notion 403s since mid-June (#3954) is the largest silent-degradation class in the product. #3972 turned those into an actionable message; until this lands, most of them still say nothing. Also worth checking once visible: the `notion_blocks_forbidden_notice_shown` event fires on mount, so a flat count in the funnel is the symptom of this bug rather than of low 403 volume.

Context: #3972, #3954, `src/services/NotionService/FEATURE.md`.

aalemayhu:
## What

Every conversion notice on the Downloads page gets its own toggle chip, so a conversion whose only signal is a 403 (or a structure rescue, or guessed columns, or unsupported blocks) can actually be seen.

Closes #3975. Closes the reachability half of #3954 — the remaining shapes there (pre-flight probe, Ankify `last_error` persistence) are untouched and still blocked on the prod `outbound_call_logs` split.

## Why

**#3972 shipped two days ago with its user-facing half dead.**

That PR built the whole detect-and-tell chain for unshared Notion blocks — `recordIfForbidden` → `forbiddenBlockCount` → `buildNotionConversionSignalPayload` → `ForbiddenBlocksNotice`. The detection works. The notice never rendered in the case it exists for.

All four done-job notices are gated on `!isFailed && isExpanded` (`DownloadsPage.tsx:888-948`). For a **done** job, `isExpanded` could only be set by one of three toggles, gated on `truncation != null`, `droppedAssets != null`, `cardLimitPartial != null`. The two auto-expand effects both filter on `isFailedJob(...)`, so they never fire here.

`pickSignalCode` makes `code` exclusive and ranks `notion_blocks_forbidden` **above** `notion_truncated`. So a pure-403 conversion emits `{"code":"notion_blocks_forbidden","forbidden_blocks":N}` — all three toggle conditions null, no auto-expand, **no affordance whatsoever**. Same dead end for `notion_structure_rescued`, `notion_columns_guessed`, and `notion_unsupported_blocks` when each wins the code slot.

`src/services/NotionService/FEATURE.md:18` already documented this verbatim as a "Known reachability gap." #3954 reports 6,717 Notion 403s — the largest silent-failure class on the product — and the fix for it was invisible.

## How

Each signal renders its own chip, all opening the same panel, mirroring the panel one for one:

| Signal | Badge | Tone |
| --- | --- | --- |
| `notion_blocks_forbidden` | `{{count}} part(s) not connected` | warning |
| `unsupported_blocks` | `{{count}} block(s) skipped` | warning |
| `notion_structure_rescued` | `Structure guessed` | neutral |
| columns guessed | `Columns guessed` | neutral |

Warning tone where content is genuinely missing; neutral where we only inferred something and nothing was lost. Copy says "not connected" rather than "not shared" to match the notice body already shipped in `downloadsx.json` ("isn't connected to 2anki").

The three existing toggles are deliberately untouched — extending the pattern, not refactoring working code inside a fix.

## The trio split, and what I did about it

**pm** and **engineer** argued for a single shared chip with a priority-picked label (bounded density). **designer** argued one chip per signal — for the 403 case the badge *is* the message, and a generic "Conversion note" chip re-buries the fact behind a click.

Both sides asserted a density figure and **both were wrong**, so I read the gating rather than counting votes:

| Parser | Gate |
| --- | --- |
| `parseTruncationPayload`, `parseForbiddenBlocksPayload`, `parseStructureRescuedPayload` | `code === …` — mutually exclusive |
| `parseColumnsGuessedPayload`, `parseUnsupportedBlocksPayload`, `parseDroppedAssetsPayload` | **field presence** — ride along with any code |

`buildNotionConversionSignalPayload` writes every field independently of the winning code. So the real ceiling is 1 code-gated + 3 field-gated = **four chips**, not the designer's "at most two" and not the engineer's implied one. Rare, but the builder permits it.

Per CLAUDE.md, a trio split on visual direction ships a preview instead of a silent pick: **`/dev/downloads-preview`** renders both candidates side by side across all eight density cases including the four-chip worst case. Shipped as per-signal (designer's call — it is the one that actually delivers on "stop being silent"), and the preview is there to overrule me in ten seconds if the density reads badly.

The route is `import.meta.env.DEV`-gated; verified absent from the production bundle (no `DownloadsPreviewPage` chunk, and the string `dev/downloads-preview` does not appear in `web/build/assets/`).

## The a11y review changed the design, and it is the best argument in the thread

`a11y-reviewer` found a defect neither pm nor designer raised, and it is the strongest case in the whole disagreement.

Per-signal chips with scoped labels ("Show skipped blocks", "Collapse what wasn't connected") **promise a disclosure that does not exist**. Every chip calls the same `toggleFailurePanel` and opens the same panel, which stacks every applicable notice. So on a 3-chip row: press chip 1, and chip 2 — never touched — silently flips to `aria-expanded="true"` and relabels itself "Collapse skipped blocks". Press it expecting to hide just that note and you close everything, including the content you opened two seconds ago.

That is a broken WAI-ARIA disclosure contract, not a labelling nit. The fix keeps the designer's substance and drops the dishonest part:

- **No `aria-label`.** The badge text is the accessible name, so a chip announces "3 parts not connected, button, collapsed" — specific and glanceable, exactly the designer's point, with no verb promising a scope the row cannot honour.
- **`aria-controls`** on each chip pointing at the `id` of the notice row it reveals. Multiple controls for one region is legal ARIA, and `aria-expanded` is now truthful for every chip: the row it names really is visible.
- **`aria-hidden` on the chevron**, now load-bearing rather than defensive, since without the `aria-label` the glyph would otherwise join the accessible name.

Net effect: **8 fewer i18n strings** (the generic `showNote`/`collapseNote` keys were already shipped and are untouched), a simpler component, and honest semantics. A regression test asserts the `aria-controls` target and the `aria-expanded` flip.

Note the underlying quirk is **pre-existing** — `truncation` and `droppedAssets` chips already co-occur today and already share one panel. This PR does not introduce it; it declines to amplify it with misleading labels.

## Verification

- **9 new tests**, written first. Verified they fail for the right reason: 6 red with the implementation stashed, all green with it. The two negative guards (clean `notion_database_resolved` conversion, failed job) pass in both states by design.
- Includes a ride-along case asserting three distinct badges on one row, and an a11y case asserting `aria-controls` / `aria-expanded`.
- **Full web suite: 358 files, 2813 tests pass.** `tsc --noEmit` clean. Tree-wide `pnpm format:check` clean (verified the tree-wide run dirtied no file outside this diff). `oxlint` clean on changed files — two pre-existing `exhaustive-deps` warnings sit on lines I did not touch.
- No server source changed. The only non-`web/` file is `FEATURE.md`, whose "Known reachability gap" paragraph this PR makes false and therefore rewrites.

## i18n

Badge keys only — the a11y rework removed the need for any new toggle-label strings. Six per locale (four for `ja`, ten for `pl`/`ru`), all added in this PR. Plural categories follow each locale's existing `imageSkipped_*` shape: `_one`/`_other` for en/de/es/fr/it/nl/pt, `_one`/`_few`/`_many`/`_other` for pl and ru, `_other` only for ja. The per-locale parity tests pass.

Worth a native-speaker pass on the pl/ru `_few`/`_many` splits and on "not connected" / "guessed", which carry idiom.

## Metric

No funnel or MRR movement claimed — this is a recovery-path fix. The read is the four `*_notice_shown` events at `/api/ops/metrics`, all of which were structurally near-zero for their pure-code cases because the notice could not mount. Day 7: each should record above zero on done Notion jobs where its code won. Any that stays flat while the backend still emits that code means the chip is not rendering.

## Sonar

Not run locally — Docker is not running, so the `sonarqube` MCP path is unavailable, and `sonar-scanner` cannot read the token from the OS keychain. `sonar auth status` reports connected. Relying on PR decoration; I will check the zero-open-findings API before merging rather than assuming a green gate means zero findings.

## Browser check

`pnpm --filter 2anki-web test:golden-path` — 2 tests passed at a 375px viewport, no console errors. The chip-density visual call is still worth a human look at `/dev/downloads-preview` (`pnpm dev`, no auth, no data needed).

## One thing I am not fixing here

`sharedStyles.badgeWarning` sets `--color-warning` on `--color-warning-light`, which **fails WCAG AA in 4 of the 5 themes** at the badge's 12px size:

| Theme | Ratio | AA |
| --- | --- | --- |
| light | ~2.1:1 | fail |
| gold | ~3.1:1 | fail |
| purple | ~2.1:1 | fail |
| hotpink | ~2.1:1 | fail |
| dark | ~7.9:1 | pass |

This is a pre-existing shared token already used by the image-skipped badge, not something this diff introduces — but this PR does add two more usages of it, on the chips most central to the fix. Changing the token touches every existing warning badge, so it wants a designer call and its own PR rather than a drive-by here. Filed separately; flagging it so the expanded usage is a decision rather than an oversight.


aalemayhu:
Contrast finding from the a11y review filed separately as https://github.com/2anki/server/issues/4011 — the `badgeWarning` token pair fails AA in 4 of 5 themes. Pre-existing, but this PR adds two usages of it, so it should be a decision rather than something that rides along unnoticed.

PR #4247, opened 2026-08-26
fix: name the stuck phase and open resources on shutdown timeout

### LINKED ISSUE #4234: Every deploy hits the 85s graceful-shutdown timeout and force-exits
## What

Every deploy in the last 48h ends the retiring process on the hard timeout instead of a clean drain:

```
HTTP close exceeded 2000ms — severing remaining connections
Graceful shutdown exceeded 85000ms — forcing exit
```

3 occurrences 2026-08-23 → 2026-08-25, including the latest deploy (green retired 2026-08-25 ~02:05 UTC). Both #4186 (bounded `server.close`) and #4225 (clear scheduler timers at top of shutdown) were live for these — `Cleared 13 scheduler timer(s)` and `SIGINT drain complete — exiting cleanly` do appear on other boots, so the fast path works sometimes; something still holds the process for the full 85s on most deploys.

## Why it matters

Blue-green hides it from users, but each deploy now spends the full 85s + a forced exit. Whatever is still pending at that point (a conversion in flight? DB pool? an Ankify docker call? the polling tick's in-flight sync?) is being killed rather than drained — the exact thing #4186 was meant to protect.

## Suggested diagnosis

- On the timeout path, log what is still open before forcing exit: active HTTP connections count, conversion-pool in-flight count, DB pool `numUsed()`/`numPendingAcquires()`, any in-flight Ankify sync.
- Compare boots that print `SIGINT drain complete` vs the ones that hit 85000ms — pattern should show which resource never resolves.

Found during a 48h prod-log review on 2026-08-25.

aalemayhu:
## What

The 85s hard-exit line now names the drain phase that was running (`HTTP server` / `Conversion pool` / `Database pool`) and dumps what is still open: the Node active-handle tally (`process.getActiveResourcesInfo()`), the Knex/tarn pool counters (`used/free/pendingAcquires/pendingCreates`), and the conversion pool's `queueSize/threads/utilization`. On the happy path each phase logs `<phase> drained in <ms>ms`, so a deploy that takes long shows where the clock went even when it does not force-exit.

## Why

Every deploy in the last 48h ended on `Graceful shutdown exceeded 85000ms — forcing exit` with nothing saying which resource held the process (#4234). Both prior shutdown fixes (#4186, #4225) were live; the fast path works on some boots. This is the diagnosis step the issue asks for — no timeout value changes (prod safety limits stay as they are, per the runtime-evidence rule).

## Measuring success

Next deploy's retiring-color log shows either three `drained in` lines and `SIGINT drain complete`, or the hard-exit line with `during <phase> drain` plus the resource dump. Either answers #4234's question; the fix PR follows from what it says.

## Testing

- `pnpm test src/lib/gracefulShutdown.test.ts` — 10 passed (2 new: stuck-phase + resource dump on hard exit; per-phase timing lines).
- `pnpm test src/lib/conversionPool.test.ts` — 20 passed.
- Full server suite: green (background run, exit 0).
- `tsc -p . --noEmit` clean; `pnpm format:check` clean; `pnpm lint` only pre-existing warnings in unrelated files.

No changelog entry — internal observability, nothing user-visible.

## Goal alignment

None — internal. Feeds the fix for deploy-time force-kills of in-flight conversions.

Fixes #4234

### DIFF
--- a/src/lib/conversionPool.ts
+++ b/src/lib/conversionPool.ts
@@ -97,6 +97,19 @@ export function getConversionPool(): Piscina {
   return initConversionPool();
 }
 
+export function describeConversionPool(): {
+  queueSize: number;
+  threads: number;
+  utilization: number;
+} | null {
+  if (pool == null) return null;
+  return {
+    queueSize: pool.queueSize,
+    threads: pool.threads.length,
+    utilization: Number(pool.utilization.toFixed(2)),
+  };
+}
+
 // Swap in a fresh pool and drain the retiring one in the background so heaps
 // reset without dropping in-flight conversions — close() waits for outstanding
 // tasks up to its budget. Guarded against re-entry so a burst of completions
--- a/src/lib/gracefulShutdown.test.ts
+++ b/src/lib/gracefulShutdown.test.ts
@@ -14,6 +14,8 @@ import type http from 'http';
 import type { Knex } from 'knex';
 import { shutdownConversionPool } from './conversionPool';
 import {
+  describeActiveResources,
+  describeDatabasePool,
   gracefulShutdown,
   HTTP_CLOSE_BUDGET_MS,
   POOL_DRAIN_TIMEOUT_MS,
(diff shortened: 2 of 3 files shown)


aalemayhu (commented):
Review-agent pass (security + engineering forks): no blocking findings. Two engineering nits (untested null-pool path; shape-only assertion on the handle tally) fixed on this head with direct tests; 12 shutdown tests green.

aalemayhu (commented):
Marker carried forward: this head is the previous reviewed tip (ef6b947) amended with no content change, purely to get a fresh CI run set after the GitHub Actions outage left non-retryable startup_failure runs on it. Tree diff between the two SHAs is empty.

aalemayhu (commented):
Marker carried forward again: same tree as the reviewed ef6b947 (empty diff), re-pushed only because SonarCloud never received the 58f3d87 push event during the Actions outage.

aalemayhu (commented):
Marker carried forward (tree identical to reviewed ef6b947; re-pushed because SonarCloud never analyzed 59933b7).

Issue #4201, opened 2026-08-22
Split UsersControllers and end inline getDatabase()/new Repo() in controllers


aalemayhu:
The harmful slice of the 254 layer-skip warnings is ~26 controllers constructing repositories or calling getDatabase() inline. Worst: src/controllers/UsersControllers.ts (1,534 lines) takes raw Knex (:76) and does new UsersRepository(this.db) at 10 call sites — and fuses three unrelated domains: auth (login/register/password/magic-link), five OAuth providers, and subscription lifecycle (cancel/pause/resume/status — billing in the login controller).

Fix: (1) split into AuthController / OAuthController / SubscriptionController, cutting the subscription block first; (2) inject repositories once via constructors across the ~26-controller cluster; (3) then carve router-level composition-root edges out of the depcruise no-layer-skip rule (they are legitimate DI wiring) and tighten the rule to error severity for src/controllers so the debt cannot regrow. Layer-skip trajectory is already shrinking (447 in June → 254 today) — this finishes the harmful part. From the 2026-08-22 architecture review.

PR #4250, opened 2026-08-26
refactor: make every UploadService dependency required

### LINKED ISSUE #4207: UploadService optional constructor deps let tests silently skip real paths
src/services/UploadService.ts:372-377 takes six *optional* constructor repositories (metrics, guid-ledger, and friends). Tests that omit them still pass while the corresponding production paths (scoring/metrics recording, guid ledger writes) simply never execute — a suite can be green while the code it claims to cover is skipped. This is the same test-blindness shape that let the ai_usage user_id NULL attribution bug (#4113) hide.

Fix: make the dependencies required in the constructor and update the ~6 call sites/tests to pass fakes explicitly (a shared fake-repo builder would help — see the buildTestApp idea in the architecture review). No behavior change in prod, where all six are always passed. From the 2026-08-22 architecture review.

aalemayhu:
## What

`UploadService`'s six optional constructor dependencies (settings, conversion-output stats, parse-path signatures, rule scores, card-GUID ledger, photo-to-flashcards) are required now. The `== null` / `?.` / `!` guards around them are gone. Tests pass explicit fakes through one shared builder, `src/test/fakes/uploadServiceDeps.ts` (`fakeUploadServiceDeps({ settings?, outputStats?, parsePaths?, ruleScores?, guidLedger?, photoToFlashcards? })`), so the 71 three-argument constructions in the four suites spread `...fakeUploadServiceDeps()` and the three that cared pass an override.

## Why

A test that omitted a dependency still passed while the production path it guards never executed — the same test-blindness that hid the `ai_usage` user_id NULL attribution bug (#4113). With every dependency present, those paths run in every test (#4207). No behaviour change in prod, where both call sites (`UploadRouter`, `McpRouter`) always passed all nine.

- Default fakes are "safe no-ops": stats/paths/scores/ledger `record` resolve, `getAllForOwner` returns `{}`, settings `attachCustomTemplates` resolves. The photo-to-flashcards default **rejects** with a pointed message, so an image-upload test that forgets to pass a vision fake fails loudly instead of silently falling through to the old "no use case → normal conversion" branch. That branch is gone from the source too (`this.photoToFlashcardsUseCase != null` check removed) — in prod it was always present.
- The builder lives under `src/test/fakes/` (new dir, beside the existing `src/test/mocks/`), not next to the service, so it is clearly test-only.
- Scope: did NOT touch the ~26-controller inline-repository cluster (#4201) — that is a separate, auth-adjacent refactor.

## Testing

- `pnpm test src/services/UploadService.test.ts src/controllers/Upload` — 9 suites / 130 tests green with the real paths now executing.
- Full server suite: green apart from the documented `getPageCount` (pdfinfo) environmental failure.
- `tsc -p . --noEmit` clean (the deploy build typechecks tests — the interface change is covered); `pnpm format:check` exit 0.

No changelog entry — internal refactor, no user-visible behaviour change.

## Goal alignment

None — internal. Closes a class of green-but-untested suites on the conversion hot path.

Fixes #4207

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -24,6 +24,7 @@ module.exports = {
     '!src/migrations/**',
     '!src/test/fixtures/**',
     '!src/test/mocks/**',
+    '!src/test/fakes/**',
     '!src/config/swagger.ts',
   ],
   coverageDirectory: 'coverage',
--- a/src/controllers/Upload/DropboxController.test.ts
+++ b/src/controllers/Upload/DropboxController.test.ts
@@ -23,6 +23,7 @@ import UsersRepository from '../../data_layer/UsersRepository';
 import UploadController from './UploadController';
 import { GetDropboxUploadsUseCase } from '../../usecases/uploads/GetDropboxUploadsUseCase';
 import { DeleteDropboxUploadUseCase } from '../../usecases/uploads/DeleteDropboxUploadUseCase';
+import { fakeUploadServiceDeps } from '../../test/fakes/uploadServiceDeps';
 
 function makeController(
   getUseCase: GetDropboxUploadsUseCase,
@@ -62,7 +63,8 @@ function makeController(
         .fn()
         .mockResolvedValue({ cards_used: 0, month_started_at: new Date() }),
       incrementCardUsage: jest.fn().mockResolvedValue(1),
-    } as unknown as UsersRepository
+    } as unknown as UsersRepository,
+    ...fakeUploadServiceDeps()
   );
   const notionService = new NotionService(notionRepository);
   return new UploadController(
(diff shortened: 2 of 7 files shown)


aalemayhu (commented):
Review-agent pass (security + engineering forks): no blocking findings. One engineering nit (coverage exclude for src/test/fakes) applied on this head. 9 suites / 130 tests green in the touched area; full server suite green except the documented pdfinfo failure.

aalemayhu (commented):
Marker carried forward: fresh SHA pushed for a full CI run set after the Actions outage; tree identical to the reviewed fb237a5.

Issue #4236, opened 2026-08-25
Four Ankify subscriptions poll forever with no active client


aalemayhu:
## What

After #4223 the polling noise is down to one line per subscription per boot, which now makes the residual signal readable: the same four Ankify subscriptions fail every boot with

```
[ankify-polling] sync failed for subscription <id> NoActiveAnkifyClientError: No active Ankify client; provision one first
```

Subscription rows 11393, 12222, 40925, 11404 (internal `ankify` subscription ids, not users). They have been in this state across every boot in the last 48h (and 544 log lines on 2026-08-23 before #4223 landed).

## Why it matters

Ankify is the top paid tier. A subscription with no active client is either (a) a paying user whose 5-minute sync has been silently dead — no cards land in Anki and nothing tells them — or (b) a stale row that should be deactivated so the scheduler stops polling it. Either way the current state is wrong: polling forever with no client and no owner-facing signal.

## Ask

- For each of the four: does the owning user still have `hasAnkifyAccess`? Did their client get torn down (container AutoRemove, host restart) without re-provision?
- If (a): re-provision, and add a user-facing status on `/ankify` when the client is gone ("sync paused — reconnect"), plus an ops line.
- If (b): deactivate the rows and have the poller skip inactive subscriptions.
- Either way: the poller should not retry a `NoActiveAnkifyClientError` subscription every tick indefinitely — back off or mark it needs-attention.

Found during a 48h prod-log review on 2026-08-25.

aalemayhu:
Shipped in https://github.com/2anki/server/pull/4251 (deployed 2026-08-26): a subscription with no active Ankify client is now retried once an hour instead of every five-minute tick, and the owner's top-level page refresh only runs when a sync was actually attempted. Closing; the merge didn't carry the auto-close link.

aalemayhu:
Follow-up from a 2026-08-28 log read: the traces behind this issue (and #4203) were July 18–20 lines, not live ones. pm2 names log files by pm_id and appends to whatever already sits at that name; the 2026-08-22 daemon restart reset the id counter, so every deploy since appended onto a June/July orphan, and mtime-scoped reads served the pre-#3764 `NoActiveAnkifyClientError` traces as fresh. Zero real occurrences in 23–28 Aug — the calm-skip path (`Anki client offline for owner N; skipped Notion fetch`) fired 3,268 times in the same window. #4251 was chasing a ghost. Fix for the log files: https://github.com/2anki/2anki.net/pull/4279

PR #4251, opened 2026-08-26
fix: retry a no-client Ankify subscription hourly, not every tick

aalemayhu:
## What

The Ankify poller backs off a subscription for an hour after a `NoActiveAnkifyClientError` instead of retrying it every 5-minute tick, and resumes normal polling as soon as a retry succeeds. In-memory per boot, like the existing one-line mute (#4223). `NO_CLIENT_RETRY_INTERVAL_MS = 1h`; `scheduleAnkifyPolling` takes a `noClientRetryIntervalMs` option so tests shrink it.

## Why

Four subscriptions with no provisioned client failed identically on every tick across every boot in the last 48h (#4236). Polling a call that cannot succeed every 5 minutes is wasted work and, before #4223, buried real errors. Hourly retry keeps a re-provisioned client picked up within the hour.

- Backoff interval: **1 hour**, matching `LAPSED_RETRY_INTERVAL_MS` in `offlineBackoff.ts` so the two "client not reachable" states behave alike. Change the constant if you want faster pickup after re-provisioning.
- In-memory, not persisted: a restart resets the backoff (one attempt per sub per boot), the same trade-off the mute already makes. A persisted `needs-attention` flag would need a migration — not done unattended.
- Scope: this is the "poller should not retry indefinitely" ask only. NOT done: the per-row triage of subscriptions 11393/12222/40925/11404 (needs prod DB + Stripe access), the `/ankify` "sync paused — reconnect" status (product/UX call, `*subscription*`-named files are a hard rail), and deactivating stale rows. The issue stays open for those; this PR says "partially addresses".

## Testing

- `scheduleAnkifyPolling.test.ts`: 12 passed — new: backs off after one failure and retries after the interval; resumes per-tick polling once the retry succeeds; the existing mute test now asserts a single attempt per boot window.
- `tsc -p . --noEmit` clean; `pnpm format:check` exit 0.

No changelog entry — ops-side polling behaviour, nothing a user sees.

## Goal alignment

None — internal (prod log signal and wasted polling on the top paid tier).

Partially addresses #4236 (leave open).


aalemayhu (commented):
Review-agent pass (security + engineering forks): no blocking findings. One engineering nit (stale FEATURE.md polling line) applied on this head. 12 poller tests green ×3 runs; full server suite green except the documented pdfinfo failure.

aalemayhu (commented):
Marker for the head carrying the SonarCloud fix committed after review (S3776-fix); behaviour unchanged, targeted tests green, tsc + format clean.

Issue #4239, opened 2026-08-26
Prune old pm2 logs holding plain subscriber emails; logrotate retain never applies


aalemayhu:
Found while triaging a support email on 2026-08-26 (context: https://github.com/2anki/server/issues/4238).

### Problem
1. **Old pm2 logs on the prod box still hold plain subscriber emails.** Before `a88a17d691cb` (2026-05-30, "fix: strip emails from Stripe sync logs") the startup Stripe sync logged `Subscription status for <email> remains active` for every subscriber (~760 lines per boot). The code has been hashed (`email=<shortHash>`) since then and the sync no longer runs at boot (`STRIPE_SYNC_ON_STARTUP=false`, gate is `=== 'true'`), so there is no live leak. But 46 rotated files under `~/.pm2/logs/` dated 27–30 May 2026 (36,305 lines, 142 MB) still contain the pre-fix lines. Grepping the log dir for any subscriber's address surfaces their whole subscription history.
2. **pm2-logrotate retention is not taking effect.** `module_conf.json` says `retain: 7`, `max_size: 10M`, daily rotate, yet `~/.pm2/logs` holds 1,717 files / 1.6 GB going back to May. Blue-green deploys create a new pm_id per color per deploy (`server-blue-out-<id>.log`, `server-green-out-<id>.log`), so each id is its own "app" to logrotate and `retain` never accumulates past 7 for any one of them. Files keep piling up per id, and pm2 reuses ids after a reset, so old content gets a fresh mtime and looks current (out-282 was written in June but mtime reads 2026-08-22).

### Proposed direction
- One-off: delete `~/.pm2/logs/server-*-out-*.log` files whose newest access-log date is before 2026-05-31 (the 46 with plain emails go with them). Read-only check first: `grep -l '@.* remains active' ~/.pm2/logs/server-*-out-*.log`.
- Durable: add a cron on the box (or a step in `deploy.2anki.net.yml`) that removes `~/.pm2/logs/server-*` files older than N days by mtime, since pm2-logrotate's per-name `retain` cannot cover a name that changes every deploy. Alternatively configure the deploy to `pm2 flush` the retiring color before it's deleted. Document the choice in `.claude/docs/prod-ops.md`.

### What NOT to do
- Don't touch the hashing in `src/lib/storage/jobs/helpers/updateStripeSubscriptions.ts`; it's already correct.
[...]

PR #4252, opened 2026-08-26
ci: prune pm2 logs older than 14 days after each deploy

aalemayhu:
## What

After a successful blue-green cutover, the deploy workflow deletes `~/.pm2/logs/server-*.log` files older than 14 days by mtime and prints how many it pruned. `.claude/docs/prod-ops.md` gains a "pm2 log retention on the box" section with the why and the two caveats.

## Why

`pm2-logrotate`'s `retain: 7` cannot cap the directory: every deploy gives each color a fresh pm_id, so every id is its own "app" to logrotate and files pile up per id — 1,717 files / 1.6 GB going back to May, including 46 pre-2026-05-30 files that still hold plain subscriber emails (#4239). A prune on deploy is the durable half of the fix.

- **14 days by mtime**, `-maxdepth 1 -type f -name 'server-*.log'`, run only after a successful cutover (so a failed deploy never deletes anything). Both colors' current files are always inside the window. Change the window if you want tighter retention.
- **The one-off cleanup is NOT in this PR.** The pre-fix files can carry a fresh mtime (pm2 reuses ids), so mtime pruning may not catch them; the issue's `grep -l '@.* remains active'` read-first deletion is a manual action on the box — agents never SSH to prod to mutate. Documented in prod-ops.
- Chose the `find` prune over `pm2 flush` of the retiring color: flush empties the live file but leaves the rotated backlog, which is the actual growth.
- Rail PR (`.github/workflows/**`): Alexander merges from the UI. The step runs on the next deploy after merge — read that run's `pruned N pm2 log file(s)` line.

## Testing

- Shell step only; no `src/` change. YAML structure unchanged apart from the appended lines inside the existing `script: |` block (GitHub validates on push — with Actions currently in a major outage the run may show `startup_failure` until it recovers; the workflow file parsed fine for the merges earlier today).
- Dry-run equivalent for the reviewer: `find "$HOME/.pm2/logs" -maxdepth 1 -type f -name 'server-*.log' -mtime +14 -print` on the box lists what the first run would delete.

No changelog entry — ops-only.

## Goal alignment

None — internal (privacy hygiene on the box, disk).

Partially addresses #4239 (leave open for the one-off cleanup).


aalemayhu (commented):
Review-agent pass (security + engineering forks): no findings. Both verified the find scope (-maxdepth 1, -type f, quoted glob, no symlink follow), that a live color's file can never age past the window, and that a missing directory or EACCES cannot fail the deploy under set -e/pipefail (tested in a temp HOME). Rail PR — waits for Alexander in the UI.

aalemayhu (commented):
Marker carried forward: fresh SHA pushed for a full CI run set after the Actions outage; tree identical to the reviewed 16acf64.

aalemayhu (commented):
Rebased onto main; no content change since the previous verdict (deploy workflow: prune pm2 server-*.log files older than 14 days after each deploy, count logged).

PR #4255, opened 2026-08-26
fix: make block-id identity re-key a page and rewrite its ledger rows

### LINKED ISSUE #4240: Ledger-miss default re-keys pre-2026 collections, then pins the wrong identity
**Goal alignment**: Retention. A signed-in re-upload must update the cards that carry review history, or long-time users lose the reason to stay.

## Problem

On a `card_guids` ledger miss, uploads issue the legacy content-formula GUID (`guid_for(deckName, front, type)`, `create_deck/create_deck.py:236-239`) and pin it. That matches the July-2026 duplicate set, not any collection built before Notion's export change: those notes carry `guid_for(blockId)` (`create_deck.py:234`, the pre-#3936 path when the toggle still had its id). The duplicate-cards doc (`/documentation/cards/duplicate-cards`, `sync/troubleshooting.md:29-35`) tells users to delete the 0-review copies and keep the reviewed ones, which is exactly the set the default can never match, so every not-yet-ledgered page duplicates once more.

`block-id-identity` (default off, `cardOptionDefaults.ts:4`, read at `DeckParser.ts:1335`) would have matched, but pinned rows keep their GUIDs (`applyLedgerGuids`), so flipping it after the fact does nothing. Nobody discovers the option before their first upload; they discover it after the duplicates, when it is already inert.

**Entry point**: signed-in Notion HTML-zip upload, `POST /api/upload/file` (`src/routes/UploadRouter.ts:135`, ledger repository constructed at `:54`) → `usecases/uploads/worker.ts` → `getPackagesFromZip` → `DeckParser` with `knownGuids` → `create_deck.py`. Not the Notion API path.

**Verified on user 10965's own files** (the reporter #3936 was verified against; third round on the same bug class):

| Set in their Anki collection | Notes | Reviews | GUID matches vs their fresh 2anki download (2427 notes) |
|---|---|---|---|
| Originals, created 2023-10 → 2026-06 | 2464 | 55,698 | 0 |
| Added 2026-08-25 from that download | 2424 | 0 | 2424 |

`guid_for(block_id)` over the 2442 ledger rows issued for that upload matches **2421** of the originals and 0 of the new set. The pinned GUIDs match 0 originals. Block ids were stable the whole time; the page settings row has `block-id-identity: "false"`.

**Blast radius (prod, 2026-08-26)**: 486 owners / 131,902 pinned rows; 139 of those owners signed up before 2026; 13 have `uploads` rows older than 2026-06-01 (lower bound, old upload rows get pruned). One owner has `block-id-identity=true`.

## User signal (verbatim quotes)

- "unfortunately I have another Decks with the same problem."
- "It doesn't recognize that there are already the same cards in the deck, so I end up with having every card two times in my deck."

## Proposed direction

Make `block-id-identity` a real recovery switch: when it is on, re-issue `guid_for(blockId)` for every block that carries a Notion UUID and overwrite that page's ledger rows (explicit opt-in; content-formula rows can always be re-derived). Fix the doc to say which copy to keep depends on when the deck was first built, and to turn the option on before re-uploading a deck that predates July 2026. Then consider the automatic version, defaulting ledger misses to block-id for accounts with uploads older than the format change, as a follow-up. Do not build an apkg-upload matcher or a second toggle.

Support recovery for the reporter is a scoped psql action (delete that page's 2026-08-25 ledger rows, set `block-id-identity` true on their settings, re-upload) and is being handled in the inbox thread, not here.

Related: #3936 (the legacy-pin design), 6eb98ad52b64 (July icon fix).

aalemayhu:
## What
Turns the `block-id-identity` card option into a real recovery switch. While it is on, a signed-in Notion HTML-zip re-upload re-keys every Notion-UUID toggle to `guid_for(blockId)` and rewrites those rows in the user's `card_guids` ledger, so the import updates a deck first built before Notion's July 2026 export change instead of duplicating it. The option is relabeled "Match cards to their Notion blocks" and both docs pages (English and German) now say which duplicate set to keep depends on when the deck was first built.

## Why
Closes #4240. On a ledger miss, uploads issue the legacy content-formula GUID, which matches only decks built after July 2026. Collections built earlier carry `guid_for(blockId)`; the option that would have matched them was inert after the first upload because `applyLedgerGuids` pinned any stored GUID regardless of the option and `record()` was insert-ignore. Verified on a user's own collection in the issue: 2421 of 2464 originals match `guid_for(blockId)`, 0 match the pinned rows. Every not-yet-ledgered page duplicated once more and the docs told users to delete exactly the set that could never match again. Retention: a re-upload must land on the cards carrying review history or long-time users lose the reason to stay.

## How
- `DeckParser.applyLedgerGuids` skips the stored GUID while the option is on (python then issues `guid_for(blockId)` via the existing `useContentGuid=false` path); `collectGuidEntries` picks the new `collectRekeyedGuids`, which returns every block whose issued GUID differs from the ledger. Both collectors share one sidecar-pairing helper with the same fail-safe tripwires.
- `CardGuidLedgerRepository.reissue` upserts on `(owner, block_id)` merging `guid` and `source_page_id`; `record` keeps insert-ignore. Both share the batching helper.
- `UploadService.recordIssuedGuids` writes through `reissue` when `settings.blockIdIdentity` is on (upload and sync paths) and logs `[guid-ledger] re-keyed N rows to block-id identity for owner X`.
- No python change: `create_deck.py` already issues `guid_for(notion_id)` when `useContentGuid` is false and no explicit guid is set.
- Label/description live server-side in `supportedOptions.ts` (not localized); docs are markdown in English and German.

- Scope: opt-in re-key only, when the option is on, for the blocks in that upload — chose this over the automatic "default ledger misses to block-id for accounts with pre-2026 uploads" because that re-keys 486 owners / 131,902 rows unattended on a heuristic. Assumption: the user-controlled switch is enough for the reporters in hand. If wrong, the automatic default is the follow-up named in the issue.
- Leaving the option on is safe: re-issuing `guid_for(blockId)` is deterministic, so repeat uploads are idempotent. A user only loses if they later build a post-July-2026 deck and want the content formula; no auto-flip-off was built.
- Copy: label "Match cards to their Notion blocks"; description "On, uploading the page again matches your toggle cards to their Notion blocks and updates them in place. Turn it on for decks you first built before July 2026. Off for newer decks." (designer's call; the old "Block ID identity" was jargon). Swap if you'd word it differently.
- Metric: no new analytics event — the server has no server-side track helper and the client never sees the re-key. Read adoption from the `[guid-ledger] re-keyed` log line and a psql count of block-id-shaped GUIDs in `card_guids`; T+14d read alongside duplicate-report support threads.
- Not built: apkg matcher (ledger already holds identity), a second toggle, a retroactive migration of existing rows.
- German docs were translated in-repo; flag for a native pass.

## Measuring success
`[guid-ledger] re-keyed N rows` lines in prod logs and a psql count of block-id-shaped GUIDs in `card_guids` two weeks after deploy, alongside the count of duplicate-set support threads from users who turned the option on (expected: zero).

## Testing
- Unit tests added: `DeckParser.stableGuids.test.ts` (ledger hit is not applied while the option is on), `collectIssuedGuids.test.ts` (`collectRekeyedGuids` returns changed + new blocks, fails safe on misalignment), `CardGuidLedgerRepository.test.ts` (`reissue` upserts with `onConflict(owner, block_id).merge(guid, source_page_id)`), `UploadService.test.ts` (reissue when the option is on, record by default).
- Full server suite: 6725 passed, 2 failed, both in `src/lib/pdf/getPageCount.test.ts` (the documented `pdfinfo` environmental failure on this machine); no other suite red. Web docs/changelog suites green; `pnpm format:check` clean; `tsc -p .` clean.

## Browser check
Notes: machine-backed via `pnpm --filter 2anki-web test:golden-path` (2/2 at 375px). The `web/src` diff is markdown docs content plus a changelog JSON rendered by the existing pages; not visually reviewed in a browser.

## Risks
- Parser core + ledger writes. The re-key only fires with the option on (default off) and only for signed-in uploads with a ledger; the default path is unchanged (`record` untouched, `applyLedgerGuids` unchanged when the option is off).
- A user who turns the option on for a post-July-2026 deck re-keys it into a second set; the label, description and docs say who should turn it on.

## Goal alignment
Retention (78% of churn is lifecycle): a signed-in re-upload updates the cards that carry review history instead of duplicating them. Read: prod log line + psql count, two weeks after deploy.


aalemayhu (commented):
**Verdict**: approve

Simpler for long-time Notion users: a signed-in re-upload of a deck built before July 2026 now updates the cards that carry their review history instead of duplicating them, and the docs finally say which set to keep.

### Blocking
- None. Security: upsert uses builder bindings with literal column names, owner comes from the session (`getOwner(res)`), the `MAX_ID_LENGTH` filter still guards the re-key path via the shared `batches` helper, the log line carries only a count and a numeric owner id. Engineering: ON path traced through `applyLedgerGuids` early return → `useContentGuid=false` → python `guid_for(notion_id)`; OFF path byte-identical (`record`/`collectIssuedGuids` semantics preserved). Execution-verified skeptic built real apkgs with python at this head: ON re-keys to `guid_for(blockId)` and returns the changed entries, OFF keeps the pinned content-formula GUID with no entries, repeat uploads are idempotent, non-UUID toggles and reversed companions stay out of the re-key set, fan-out `::i` notes stay distinct (8/8 throwaway cases + stableGuids 14/14).

### Nits
- Addressed on this head: the re-keyed log line now fires after the upsert settles; changelog title shortened to user voice; docs say "matched"/"split" instead of "re-key".
- Noted, not blocking: the option label is served in English from `supportedOptions.ts` while the German docs name it in German — pre-existing (card option labels are not localized). Sidecar-missing behavior (apkg re-keyed, nothing written) is documented in FEATURE.md.

Issue #4198, opened 2026-08-22
Correlation spine: request ID stamped through job, worker, and claude-usage logs


aalemayhu:
No request/correlation ID exists anywhere in the server (zero grep hits for requestId/x-request-id/traceId). Request → job → worker output → [claude-usage] cannot be joined; the [claude-usage] console line (src/lib/claude/recordClaudeUsage.ts:22-24) carries surface/model/cost but no user or job id — this is exactly the shape that kept the user_id NULL spend incident (#4113) invisible.

Fix (a helper, not a platform): per-request UUID in middleware (res.locals.requestId), stamp it into the job row at creation, and include {requestId, jobId, userId} in every conversion and claude log line via a tiny structured-log helper. From the 2026-08-22 architecture review.

aalemayhu:
Shipped: #4317 carries the request id through the Notion conversion logs and #4256 stamps it on every Claude spend and chunk log line — the spine this issue asked for is on prod. The review follow-ups that remain (correlation gaps found while shipping it) are tracked in #4325, so closing this one.

PR #4256, opened 2026-08-26
feat: stamp the request id on every Claude spend and chunk log line

aalemayhu:
## What
Threads the request id that `requestIdMiddleware` already puts on `res.locals` through the upload, sync and restart conversion paths into every Claude call, so each `[claude-usage]` line renders `request=<id>` and every per-chunk `[Claude]` log object (`Sending request`, `Received response`, `chunk done`, `Response truncated`, `API request failed`) carries `requestId`.

## Why
Part of #4198. `[claude-usage]` carried surface, model and (since #4219) user, but nothing that joins a spend line back to the request that started the conversion — the exact shape that kept the user_id NULL spend incident (#4113) invisible. With this, `grep request=<X-Request-Id>` in prod logs returns the chunk timeline and the billed tokens for one user-reported conversion. Internal observability; no user-visible change — no changelog entry.

## How
- `GenerateDeckInfoOptions.requestId`; a `SpendAttribution { userId, requestId }` pair replaces the bare `userId` parameter on `generateDeckInfoFromChunk` and `runFloorV1` (same parameter count, so no new S107 finding); `recordClaudeUsage` gains `requestId` and appends `request=` only when known.
- Plumbing: `UploadService` passes `res.locals.requestId` at both `GeneratePackagesUseCase.execute` call sites and into `runClaudeRestart` → `GeneratePackagesUseCase.execute(…, requestId)` → `UploadGenerationTask.requestId` → worker `processFile` → `PrepareDeck` input (`DeckParserInput.requestId`) and the zip AI path (`GetPackagesFromZipOptions.requestId` → `buildClaudeFlashcardDeck` ctx) → `baseGenerateDeckInfoOptions`.
- The plain-parser zip paths do not call Claude, so they do not carry it. The Notion `performConversion` path has no request and is untouched.
- Left out on purpose: the job-row stamp from #4198 needs a `jobs` migration (Tier 3) and is a follow-up; this slice is migration-free.

## Measuring success
On prod after deploy: `pm2 logs --nostream | grep '\[claude-usage\]' | grep -c 'request='` is non-zero for upload conversions, and one `X-Request-Id` from a response header greps to its `[Claude] chunk done` lines.

## Testing
- Unit tests added: `recordClaudeUsage.test.ts` (line carries `request=` when known, omits it when not), `ClaudeService.test.ts` (all six chunk usage lines carry `user=42 request=…` when options set it), `GeneratePackagesUseCase.test.ts` (task carries `requestId`), `PrepareDeck.test.ts` (every `generateDeckInfo` call gets `{ userId, requestId }`), `UploadService.test.ts` (`execute` receives `res.locals.requestId`).
- Targeted: 21 suites / 382 tests green. Full server suite: 6723 passed, 2 failed, both in `src/lib/pdf/getPageCount.test.ts` (the documented `pdfinfo` environmental failure on this machine). `tsc -p .` clean (tests included), `pnpm format:check` clean.

Browser check: not applicable — server-only logging change, no `web/src` files touched.

## Risks
- Parser/Claude hot path, but additive: a new optional field on an options object and log payloads; no control flow changes. Attribution is `undefined` on paths without a request, exactly as before.

## Goal alignment
None — internal. Makes the spend-attribution join possible so the next cost incident is diagnosable in minutes instead of days.


aalemayhu (commented):
**Verdict**: approve

Internal observability: every Claude spend line and per-chunk log now carries the request id, so one X-Request-Id greps to its chunk timeline and billed tokens — the join the #4113 spend incident lacked.

### Blocking
- None. Security: the only writer of `res.locals.requestId` accepts an inbound header solely when UUID-shaped and otherwise generates one, so the logged value is never free text; the id is log-only (no path, query, shell or `res.send` use) and is not persisted into the events table; the worker task field is a plain string, structured-clone safe. Engineering: all five `generateDeckInfoFromChunk` call sites and the `runFloorV1` call pass the attribution pair; the zip paths left out never reach Claude (same gate as `PrepareDeck`'s `claudeAIFlashcards && noLimits`); `runClaudeRestart`'s one caller updated; no `any`, no layer or scope creep.

### Nits
- Addressed on this head: the PrepareDeck test reuses the file's existing mocked `generateDeckInfo` binding; the ClaudeService console spy restores in a `finally`.
- Accepted: `processFile` gained a tenth positional parameter; folding the three optional trailing params into an options object is a wider refactor than this change warrants.

aalemayhu (commented):
Rebased onto main after #4255 merged; no content change since the previous verdict.

aalemayhu (commented):
Re-review for the head that folds knownGuids and requestId into a GenerationContext (Sonar S107 on execute) and rebases onto main after #4257; the thread and its tests are unchanged in substance.

Issue #4200, opened 2026-08-22
Shared Modal and useDrag primitives for the web app


aalemayhu:
The web app has 11+ hand-rolled dialog implementations with no shared Modal (ConsentModal.tsx:36, SettingsModal.tsx:39, NotionColumnMappingModal.tsx:47, ProducerCaptureModal.tsx:81, AccountDeletion.tsx:40, ConversationsSidebar.tsx:295, MindmapMarkdownModal.tsx:49, TemplatesPage.tsx:164, CardOptionsPage.tsx:365, MindmapEditor.tsx:185, NotionImportDrawer.tsx:342) and 8 independent drag-drop implementations (UploadForm, ApkgDropZone, ImageQueue, PhotoToFlashcardsPage, PrintForm, ChatPanel, MindmapEditor) while a useDrag hook exists but is consumed only by UploadForm.

Fix: extract components/Modal (wrapping <dialog> + focus/escape handling) and promote useDrag from UploadForm/hooks/ to lib/hooks/; migrate call sites opportunistically. Both are past the repo's rule-of-three several times over. A11y benefit: one place to get focus-trap/escape right. From the 2026-08-22 architecture review.

aalemayhu:
Checked this against the current version — both halves shipped: the shared dialog primitives (useDialog/useDialogFocus, PR #4221) now back 8 call sites including ConsentModal, SettingsModal, NotionColumnMappingModal, ProducerCaptureModal, AccountDeletion, ConversationsSidebar, MindmapEditor and ConversionReportModal, and useDrag was promoted to web/src/lib/hooks with cleanup plus a shared useDropZone (PR #4257). The remaining hand-rolled dialogs migrate opportunistically as the issue itself proposed, so I'm closing it. If a specific dialog still misbehaves on focus or escape, open a new issue naming it.

PR #4257, opened 2026-08-26
fix: give useDrag cleanup and a live callback, add useDropZone

aalemayhu:
## What
Fixes the shared `useDrag` hook and adds an element-scoped sibling, `useDropZone`, then moves `PrintForm`'s hand-rolled drop zone onto it. The upload page's page-wide drop keeps working after navigating between pages, and always sees the form's current state.

## Why
Part of #4200 (the `useDrag` half; the hook itself was already promoted to `web/src/lib/hooks/` in #4221). `useDrag` assigned `body.ondragover`/`ondrop` once with an empty dependency list, which had three consequences: the drop handler kept the first render's closure (UploadForm's `isUploadLocked` / `zoneState` guards never saw later values), a second consumer overwrote the first, and nothing was removed on unmount, so a file dropped on the next page still hit the upload page's dead handler.

## How
- `useDrag`: `addEventListener`/`removeEventListener` with cleanup; the latest `onDrop` is read through a ref. Same return shape (`{ dropHover }`), same page-wide semantics, so `UploadForm` is untouched.
- `useDropZone`: new element-scoped hook returning `{ dropHover, dragProps }`; spread `dragProps` on the zone element. Two hooks rather than a mode flag on one.
- `PrintForm`: the one strict-subset consumer from the scout (identical over/enter/leave/drop + hover + first-file), now on `useDropZone`. The other zones (`ApkgDropZone`, `ImageQueue`, `ChatPanel`, `MindmapEditor`, `PhotoToFlashcardsPage`) carry extras (disabled guard, type filters, nested-drag counters, drop coordinates, no hover) and stay for a later pass.

- Scope: fixed the hook's three defects and migrated only `PrintForm`. Assumption: the remaining zones need `useDropZone` extended (type filter, nested counter) and are worth their own PR each rather than a broad sweep with no drag tests behind them. If you'd rather sweep, `ApkgDropZone` is next (only a `disabled` guard away).
- Not built: a body-vs-element mode flag on `useDrag`; a second hook keeps both call sites honest.

## Measuring success
None — internal correctness. A drop on `/print` after visiting `/upload` no longer triggers the upload handler; covered by the unmount test.

## Testing
- Unit tests added: `useDrag.test.tsx` (hover tracking, latest-callback after rerender, no listener after unmount, two consumers both receive the drop — the last three failed before the fix), `useDropZone.test.tsx` (element hover + drop, drops outside the zone ignored), `PrintForm.test.tsx` (a dropped deck posts like a picked one).
- Full web vitest: 375 files / 3345 tests green. Web typecheck and lint clean (pre-existing warnings only), `pnpm format:check` clean. Server suite untouched (no `src/` change).

## Browser check
Notes: machine-backed via `pnpm --filter 2anki-web test:golden-path` (landing + signed-in dashboard, which mounts `UploadForm` and therefore `useDrag`, 2/2 at 375px). `/print` drop verified in vitest only.

## Risks
- `useDrag` semantics preserved except for the fixes; if anything relied on the stale first-render closure it was a bug, not a feature.

## Goal alignment
Simpler/faster: drag and drop behaves the same wherever the user came from. Internal; no funnel metric.

### DIFF
--- a/web/src/lib/hooks/useDrag.test.tsx
+++ b/web/src/lib/hooks/useDrag.test.tsx
@@ -1,0 +1,62 @@
+import { fireEvent, render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { describe, expect, it, vi } from 'vitest';
+import { useDrag } from './useDrag';
+
+function Harness({
+  onDrop,
+  label = 'zone',
+}: Readonly<{ onDrop: (event: DragEvent) => void; label?: string }>) {
+  const { dropHover } = useDrag({ onDrop });
+  return <div data-testid={label}>{dropHover ? 'hovering' : 'idle'}</div>;
+}
+
+describe('useDrag', () => {
+  it('tracks hover from body drag events', () => {
+    const { getByTestId } = render(<Harness onDrop={vi.fn()} />);
+
+    fireEvent.dragEnter(document.body);
+    expect(getByTestId('zone')).toHaveTextContent('hovering');
+
+    fireEvent.dragLeave(document.body);
+    expect(getByTestId('zone')).toHaveTextContent('idle');
+  });
+
+  it('hands a body drop to the latest onDrop after a rerender', () => {
+    const first = vi.fn();
+    const second = vi.fn();
+    const { rerender } = render(<Harness onDrop={first} />);
+
+    rerender(<Harness onDrop={second} />);
+    fireEvent.drop(document.body);
+
+    expect(first).not.toHaveBeenCalled();
+    expect(second).toHaveBeenCalledTimes(1);
+  });
+
+  it('stops listening once the consumer unmounts', () => {
+    const onDrop = vi.fn();
+    const { unmount } = render(<Harness onDrop={onDrop} />);
+
+    unmount();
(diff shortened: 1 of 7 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler for anyone who drags files: the upload page's page-wide drop survives navigating between pages and always sees the form's current state, and the print page shares the same drop-zone behavior instead of its own copy.

### Blocking
- None. Security: drop handling reads the same `dataTransfer.files[0]` as before, listeners are added and removed with the component, no HTML or URL handling. Engineering: the three `useDrag` regression tests fail against the old `body.ondrop = onDrop` in a `[]` effect (stale callback, persisting handler, second consumer overwriting the first) and pass now; `PrintForm` behavior traced as preserved; `UploadForm`'s API unchanged; two hooks rather than a mode flag. Voice: changelog title in user voice, no copy changed on the print page.

### Nits
- Addressed on this head: the element-vs-page test now hovers the zone, then drops outside and on the body, asserting the callback is never reached and the hover survives.
- Accepted: `dragProps` handlers are recreated per render; fine on a label, note it if a consumer memoizes children.

PR #4260, opened 2026-08-27
docs: ban add-on for plans and features in VOICE.md

aalemayhu:
## What
Adds one row to the banned-words table in VOICE.md: never call a 2anki plan or feature an add-on.

## Why
In Anki an add-on is an installable extension. A support draft today called Auto Sync an add on, which tells an Anki user the wrong thing. The table now carries the term with the replacement wording so copy and support replies stop reaching for it.

## Testing
Docs only, no code. No changelog entry — internal writing guide.

Browser check: not applicable — markdown guide, no rendered surface.


aalemayhu (commented):
**Verdict**: approve

One banned-words row: table syntax matches the header, the Anki add-on claim is correct, the Why cell is now as terse as its neighbours, and the support-email guide already defers to VOICE.md so the rule reaches support replies.

PR #4262, opened 2026-08-27
fix: read account card options on upload, add ops block-id switch

### LINKED ISSUE #4258: Upload card options live only in browser localStorage; support cannot set block-id identity for a user
**Labels**: bug, uploads, retention, from-inbox
**Goal alignment**: retention. Long-tenure Notion uploaders are the lifecycle-churn segment; a years-old update workflow that silently duplicates every card is a churn trigger and a repeat support sink.

### Problem

For HTML-zip uploads the `block-id-identity` option (and every other card option) is read from the request body, which the browser fills from localStorage (`UploadForm.buildFormData` appends every localStorage entry). `/api/settings/card-options` returns defaults only, and the page-keyed `settings` rows support can edit are read only by the Notion-API conversion path (`performConversion`).

So a support recovery that flips the option server-side is unreachable on the upload path. Verified on user 10965: after #4255 shipped, their re-upload via `POST /api/upload/file` at 03:36Z on 2026-08-27 wrote 2,425 `card_guids` rows carrying the legacy content-formula GUID (none match `guid_for(blockId)`), and no `[guid-ledger] re-keyed` line was logged. The account-level switch support had set on 26 Aug lived in `settings` rows the upload never reads, so every card duplicated again. The option is also per-browser, invisible to support, and lost across devices.

**Entry point**: signed-in HTML-zip upload, `POST /api/upload/file` (`UploadService.handleUpload` → `new CardOption(req.body)` → `GeneratePackagesUseCase` → `DeckParser` with `knownGuids`). Not the Notion API path.

- "unfortunately it didn't work. Nothing was updated and every card was added again."
- "It doesn't recognize that there are already the same cards in the deck, so I end up with having every card two times in my deck."

### Proposed direction

Persist the user's default card options server-side (DB, per the no-localStorage-for-durable-preferences rule in `.claude/rules/code-quality.md`) and have `POST /api/upload/file` fall back to the stored default for any key the request body omits, keeping localStorage as the client cache. Add one ops affordance, a route plus its `CommandsTab.tsx` button, to set `block-id-identity` for a user's uploads so support never needs psql for this again.

Not in scope: a new options UI, a cross-device settings-sync system, a per-page override matrix, or a migration of existing localStorage blobs. One durable default plus one ops switch.

Related: #4240 (the re-key switch, shipped in #4255), #3936.

aalemayhu:
## What
- `POST /api/upload/file` (signed-in) now merges the account's stored `users.card_options` under the request body before building `CardOption`; anonymous uploads are unchanged.
- A stored `block-id-identity` wins over whatever the browser sent (closed one-key override set in `src/lib/upload/mergeStoredCardOptions.ts`); every other key stays body-wins.
- New ops command `POST /api/ops/set-block-id-identity` `{ email, enabled }` + a "Match cards to Notion blocks" section in Ops → Commands with Turn on / Turn off. Off removes the key (never stores `false`), so the user's own toggle applies again.
- `PATCH /api/users/me/preferences` jsonb-merges `card_options` instead of replacing them, so the support-set key survives the user's next option sync. `DELETE …/card-options` still nulls the column.
- `src/lib/parser/FEATURE.md` documents the switch.

## Why
Closes #4258. Support set the recovery switch for user 10965 in page-keyed `settings` rows that the upload path never reads; their re-upload after #4255 issued 2,425 legacy GUIDs and duplicated every card again. Retention: a years-old update workflow that silently duplicates a deck is a churn trigger and a repeat support sink.

- Precedence: body wins per key, EXCEPT a stored `block-id-identity` wins over the body — trio call (pm): the card-options page writes `'false'` to localStorage when the box was ever unchecked, so body-wins-everywhere would let a stale `false` defeat the support switch. Assumption: this override set stays closed at one key. If you'd rather have body-wins everywhere, drop the loop in `mergeStoredCardOptions.ts`.
- Ops "Turn off" removes the key rather than storing `false`, so support can never lock a user out of their own toggle. Assumption: nobody needs a forced-off.
- `block-id-identity` is deliberately NOT added to the browser↔server sync allowlists (`CARD_OPTION_KEYS` / `ALLOWED_CARD_OPTION_KEYS`): if it synced, the browser's value would overwrite the support switch on the next PATCH. The user's own toggle keeps working exactly as today (localStorage → body). The designer suggested syncing it; pm + engineer facts won.
- Preferences PATCH switched from replace to jsonb merge. Safe because unchecking writes `'false'` (never removes a key) and the reset flow already calls the DELETE route; the in-memory repo mirrors the merge. If you want replace back, restore `sanitizeCardOptions(...)` in `patch()` and keep the ops key out of the allowlist some other way.
- Identifier for the ops command is the account email (what support has from the inbox), resolved case-insensitively via `UsersRepository.getByEmail`.
- Stored options are read through a new `UsersRepository.getCardOptions` so `UploadService` gains no constructor dependency.
- Copy (designer): section heading "Match cards to Notion blocks"; success "On for {email}. Ask them to upload the same page again — their existing cards update in place." / "Off for {email}. Their own card option setting applies again."
- Scope: no new options UI, no per-page override, no localStorage migration, no hydrate-on-boot. No changelog entry — the user-visible effect only fires after support flips the switch; the fallback for omitted keys is invisible to a user whose browser already hydrated at login.
- PM's riskiest assumption, not verifiable overnight: that pinning the switch actually stops user 10965's duplication. Please flip it for that account in Ops → Commands, ask for a re-upload, and check `card_guids` / the `[guid-ledger] re-keyed` log line.

## Browser check
Notes: `pnpm --filter 2anki-web test:golden-path` — 3 passed (landing, signed-in dashboard, account page, all at 375px, no console errors). The Ops-page section itself is covered by `CommandsTab.test.tsx` (turn on / turn off / missing email); unattended overnight run, no manual click-through of /ops.

## Tests
- Server: full Jest suite — 698 suites passed, 6759 tests passed; the only failures are the 2 documented environmental `src/lib/pdf/getPageCount.test.ts` cases.
- Web: full Vitest suite — 377 files, 3359 tests passed.
- `pnpm format:check`, `tsc -p . --noEmit`, web `typecheck`, web `lint`, `pnpm arch` (0 errors) all green.


aalemayhu (commented):
**Verdict**: approve

Retention fix: signed-in uploads now honor the account's stored card options, and support can pin the block-id recovery switch from Ops → Commands instead of psql — the duplicate-deck loop for long-tenure Notion uploaders finally has a reachable fix.

### Blocking
- none (security, engineering, and voice forks all clean)

### Nits
- `src/services/UploadService.ts` reassigns `req.body` to the merged view so `persistConversionSettings` / `buildClaudeFallbackSettings` see the same options — deliberate, but passing the resolved input explicitly to `buildClaudeFallbackSettings` would be less surprising later.
- `UsersByEmailLookup` in `SetBlockIdIdentityUseCase.ts` is the second copy of the shape in `GrantUnclaimedPassUseCase.ts`; fine at two, fold on the third.
- `console.error('[ops] setBlockIdIdentity failed', error)` mirrors the grant-pass handler; a hashed email surrogate would be tidier if a query error ever carries the bound email.

aalemayhu (commented):
**Verdict**: approve (round 2). Head 37cada52fdb56035679d3809c24b2e897680bbd0 only clears the three SonarCloud findings from round 1 (typed upload owner, findByText in the CommandsTab spec); security, engineering, and voice forks re-ran clean with no blocking or nit findings.

aalemayhu (commented):
**Verdict**: approve (marker re-issued). Head 567d584fd65c9addd9b7c510f8c5d84b5bd10687 is an empty commit on top of 37cada52 to retrigger SonarCloud AutoScan, which skipped the previous push; the diff is byte-identical to the round-2 review, so the round-2 verdict (security, engineering, voice forks all clean) carries.

PR #4269, opened 2026-08-28
fix: digest lookup under the canonical repo name, Needs-you line

aalemayhu:
## What
- `.claude/commands/ship.md`: the two `gh issue list` calls that use `--search`/`--label` (step-7 deploy-failure lookup, step-8 digest lookup) now pass `--repo 2anki/2anki.net`.
- `.claude/commands/ship.md` + `.claude/docs/autonomous-shipping.md`: digest entries carry any Alexander-only manual step as a `**Needs you:**` first line (never only inside `Decisions`), and the day's digest issue is closed by Alexander once read — closed means reviewed, agents never close one.
- `CLAUDE.md`: one sentence beside the existing rename note recording the search-API gotcha.

## Why
`gh issue list` with `--label` or `--search` goes through GitHub's search API, which does not follow the `2anki/server` → `2anki/2anki.net` rename and silently returns 0 results (verified: 0 under the old name, 4 under the canonical one; unfiltered list, `view`, `create`, `comment` follow the redirect). That is why `/ship` missed the existing `Shipped 2026-08-27` issue and created the duplicate #4264.

The `Needs you` line comes from today's digest review: the #4262 entry left a manual step (flip the ops switch for one account and ask for a re-upload) inside its `Decisions` block, and it was found by accident while closing old digests.

Hard rail (`.claude/`, `CLAUDE.md`) — waits for Alexander. No changelog entry — internal tooling, no user-visible behavior change. Browser check: not applicable — no `web/src/` change. Sonar: not run locally; PR decoration will report.

PR #4244, opened 2026-08-26
feat: let agents merge and deploy through /ship

aalemayhu:
## What

Agents merge and deploy their own non-rail PRs through a new `/ship` command; Alexander reviews only hard-rail PRs and reads a daily digest.

- **Merge gate** (`.claude/hooks/check-merge-status.py`): rollup green + tests ran (existing), **plus** no hard-rail path (`hard_rails.py`), a review-agent pass marker for the head SHA (`<!-- ship-review: pass sha=… -->`), and SonarCloud clean for that SHA via the API (`sonar_gate.py`, fails closed — Sonar posts no GitHub check on this repo).
- **Hard rails** (`hard_rails.py`): auth, payments/Stripe, subscriptions, checkout, webhooks, migrations, generated data layer, CI/deploy pipeline, hooks/settings, prod safety limits. Rail PRs go ready with the review verdict posted and wait for Alexander in the GitHub UI.
- **`/ship <pr>`** (`.claude/commands/ship.md`): preflight → `/review-pr` fan-out posts the marker → wait for rollup + Sonar → `gh pr merge --squash` → watch the deploy run for the merge SHA → `/api/version` sha check + `/deploy-status` → `git revert` PR on failure → append to today's `Shipped <date>` digest issue.
- **CI**: every workflow runs on every push (path filters dropped) so the checks can be required on `main`; web `test` job renamed `test-web` (was colliding with the server `test` name); playwright now runs for dependabot too.
- **Shared matcher** (`merge_command.py`): the three merge hooks strip heredoc bodies and quoted literals before matching, so a mention no longer trips the gate while `then gh pr merge` / `exec` / backtick forms still do (a doc-edit heredoc and a commit `-m` mentioning the phrase were both denied as rail merges while building this).
- **Prose**: every "NEVER merge" in `overnight-prs`, `drive-acquisition`, `implement`, `review-pr`, `engineer`, `_trio`, and CLAUDE.md now says "merge via `/ship`". Reference doc: `.claude/docs/autonomous-shipping.md`.

## Why

Every merge was a human step by convention only — branch protection required zero checks and zero reviews, and every real gate lived in a local hook. Making the gate mechanical lets the trio's decisions ship without Alexander reviewing each diff, while keeping him on the surfaces where an agent's confident-wrong change reaches paying users.

## Decisions (brainstormed with Alexander 2026-08-26)

- Keep human on hard rails; auto-merge everything else.
[...]

PR #4270, opened 2026-08-28
ci: finish every required job early on docs-only pushes

aalemayhu:
## What
- New composite action `.github/actions/changes` (`action.yml` + `classify.sh`): asks `gh api repos/<repo>/compare/main...<sha>` for the changed files and outputs `code=false` only when every file is under `.claude/`, `Documentation/`, or is a `*.md` outside `src/` and `web/` (oxfmt formats Markdown under both, so those stay code). Pushes to `main`, anything under `.github/`, `src/` or `web/`, an empty compare, an API error, or a diff of 300+ files all yield `code=true`.
- `server.yml`, `web.yml`, `playwright.yml`: every job runs `changes` right after checkout and gates setup, install, build, test, and artifact-upload steps on `steps.changes.outputs.code == 'true'`. Job names and the five required contexts are unchanged.
- `classify.test.sh`: nine cases against a stub `gh` (run with `bash .github/actions/changes/classify.test.sh`; passes locally and in the reviewer's checkout).
- `.claude/docs/autonomous-shipping.md`: one sentence on why this is a short-circuit and not a `paths-ignore`.

## Why
Branch protection requires all five checks with `strict: true`. A `paths-ignore` filter would leave those checks unreported on a `.claude/`-only PR, which GitHub treats as pending, so the PR could never merge (the reason #4244 removed path filters). Today's docs-only rail PR #4269 cost a full ~10-minute CI cycle for a change that builds nothing. With this, the same push reports five green checks in about 20 seconds.

## Decisions
- Docs paths are a closed allowlist (`.claude/**`, `Documentation/**`, root-level and other `*.md` outside `src/` and `web/`); `web/src/pages/DocsPage/content/*.md` renders in the app and `src/**/FEATURE.md` is inside `pnpm format:check`'s scope, so both stay code (the latter was a round-1 review catch).
- Compare API instead of a deep clone or a third-party paths-filter action: no new dependency, no `fetch-depth: 0`, and `GITHUB_REPOSITORY` is already the canonical name so the rename does not bite.
- Fail-safe direction is always `code=true`.

Hard rail (`.github/`, `.claude/`) — waits for Alexander. First proof is this PR's own CI: it touches `.github/`, so all five jobs must run in full and stay green. The docs-only path gets exercised by the next `.claude/`-only PR. No changelog entry — CI only. Browser check: not applicable — no `web/src/` change. Sonar: not run locally; PR decoration will report.

### DIFF
--- a/.claude/docs/autonomous-shipping.md
+++ b/.claude/docs/autonomous-shipping.md
@@ -43,7 +43,7 @@ One GitHub issue per UTC day, `Shipped YYYY-MM-DD`, label `shipped-digest`, one 
 
 ## Throughput
 
-Docs-only pushes (`.claude/**`, `Documentation/**`, `*.md` outside `web/`) still run every required job, but each job asks `.github/actions/changes` first and finishes in seconds when nothing needs building — a `paths-ignore` would leave the required checks unreported and the PR blocked forever. `strict: true` means a PR behind `main` must rebase and re-run CI before it can merge; two concurrent `/ship` runs serialize at roughly ten minutes per PR. Dependabot PRs behind `main` need `@dependabot rebase`. If this hurts, the next step is GitHub's merge queue (workflows would need the `merge_group` trigger) — not loosening `strict`.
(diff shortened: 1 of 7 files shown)


aalemayhu (commented):
**Verdict**: approve (round 2). Head a6afe2654079f3caae935cd81d8ba82166658975 fixes the round-1 blocker: Markdown under `src/` and `web/` is inside `pnpm format:check`'s oxfmt scope, so it now counts as code (`src/*|web/*` before `*.md`), with a test case. The reviewer re-ran `classify.test.sh` from the branch (9 ok, 0 fail) and confirmed root `*.md` is outside every CI job's scope, `gh api` failure / empty compare / 300+ files / pushes to `main` all fail safe to `code=true`, every post-checkout step in all three workflows is gated, and the five required contexts keep their names. This PR itself touches `.github/`, so its own CI runs in full — the green rollup is the wiring proof. Hard rail (`.github/`, `.claude/`) — waits for Alexander.

PR #4278, opened 2026-08-28
fix: drop PDF page footers and keep checklists whole in auto text cards

### LINKED ISSUE #4276: PDF page-number footers become card fronts and absorb the bullets after a page break
**Goal alignment**: A third of the cards on a real PDF upload are fronted by a page number with the wrong content on the back — the opposite of "beautiful Anki flashcards", and it directly lowers successful-first-card-review on the core conversion path.

### Problem
On the one-shot PDF upload path, a bare page-number footer (`1`, `2`, … `35`) on every page is promoted to a card front, and because the footer sits at the page break in the middle of a bullet list, the bullets that continue on the next page get attached to the page-number card instead of the real heading above it. On a 35-page lecture-notes PDF exported from Google Docs this produced 35 cards fronted by page numbers and left real heading cards missing the bullets that continued past the break. The docx path is unaffected — the tester confirmed the problem is PDF-only.

Pinned entry point: web file upload → `PrepareDeck.convertPdfByAutoDetection` → `convertPdfTextToHtmlAuto` → `synthesizeCardsFromPdfHeadings` (`src/lib/parser/synthesizeCardsFromPdfHeadings.ts`). Prod log line `[PrepareDeck] convertFile pdf→text→html (auto) { cardCount: 107, pageCount: 35 }` (user 21936).

Root cause (verified by reproduction): `isHeadingLine()` promotes any line under 60 chars whose next line is longer, unless it ends in `.,;`, starts with a list marker (`•`/`-`/`*`/`1.`/`1)` + space), starts lowercase, or continues a wrapped line. A digits-only footer passes all four checks. Not a regression from #3521 — that fix reached the tester (364 → 107 cards); this is a residual gap in the same heuristic.

- "De 35 første kortene som ble laget får sidetallet fra dokumentet som frontside på kortene. Så jeg har 35 sider i dokumentet, og har nå fått 35 kort med det på frontsiden av kortene." (The first 35 cards get the page number from the document as the front. I have 35 pages, and now 35 cards with that on the front.)
- "på baksiden så er ikke alle punktene tatt med. Jeg mistenker at det i dette tilfellet er på grunn av sideskillet/at punktlisten avbrytes av en ny side." (the back doesn't include all the bullet points. I suspect it's because of the page break / the bullet list being interrupted by a new page.)
- "Dette problemet oppstår ikke med word dokumentet, men kun PDF dokumentet." (This doesn't happen with the Word document, only the PDF.)

### Proposed direction
Reject a digits-only line (optionally with surrounding whitespace, and common `Page N` / `N / M` footer shapes) in `isHeadingLine()` and drop it from the body too, so the bullets straddling the page break rejoin the heading above them. Reproduce first with a failing Jest test using the tester's shape — heading + one bullet, footer `1`, page break, five continuation bullets — asserting no card is fronted by `1` and all six bullets land on the real heading's back. Digits-only is a safe reject: real numbered headings (`1) Title`, `2. Title`) are already excluded as list markers and carry text. One fix closes both reported symptoms.

### LINKED ISSUE #4277: Google Docs checkbox-list lines get promoted to headings, dropping most of the list
**Goal alignment**: A 9-item study checklist arriving as 1 item plus 2 junk cards is silent content loss on the core path — the quiet failure that erodes trust in the deck and stalls per-user retention.

### Problem
When a PDF exported from Google Docs contains a checkbox list, the checkbox glyph is not present in the extracted text, leaving a run of short, uppercase-starting lines with no bullet marker (e.g. `Classes 1-2: …`, `Classes 3-4: …`, `Class 7: …`). Alternate lines in that run get promoted to card fronts whenever the next line happens to be longer, so a 9-item checklist under a "Teaching plan" heading became a "Teaching plan" card with only the first item on the back, plus two junk cards fronted by list items. The tester guessed this wasn't a bug; it is the same heuristic gap. PDF-only; docx is clean.

Pinned entry point: same as the page-footer issue — `synthesizeCardsFromPdfHeadings` (`src/lib/parser/synthesizeCardsFromPdfHeadings.ts`), reached via the auto PDF path (`[PrepareDeck] convertFile pdf→text→html (auto)`).

Root cause (verified by reproduction): `isHeadingLine()` has no notion of a sibling run — each short line is judged only against its immediate neighbour's length. Checkbox glyphs (☐/☑/✓/✔) are also absent from `LIST_MARKER`, so even when a PDF does carry them the line is not recognised as a list item.

- "Kanskje en feil: tror ikke dette er en feil ettersom jeg brukte 'check list' funksjonen i google doc." (Maybe a bug: I don't think it's a bug since I used the 'check list' function in Google Docs.)

### Proposed direction
Two parts. (1) Add checkbox glyphs (☐/☑/☒/✓/✔) to `LIST_MARKER` — flag this as an extension of a locked set in the PR. (2) Add a sibling-run guard: a short line is not a heading when it and the previous non-blank line share a prefix pattern (`Class(es) N…:`, same leading token, or both under the heading length with no terminal punctuation) — keep the run under the preceding heading. Reproduce first with a failing Jest test using the 9-item `Classes N-M: …` shape, asserting one "Teaching plan" card carries all nine items and no list item becomes a front. Shares the `isHeadingLine()` surface with the page-footer issue; may ship in the same PR, but both test cases verified independently. Scope to this pattern, not a heading-heuristic rewrite.

aalemayhu:
## What

Second-round feedback from a paid usability tester (user 21936) on the one-shot PDF upload: after #3521 took their deck from 364 to 107 cards, 35 of those 107 were fronted by the page number, heading cards were missing every bullet that continued past a page break, and a 9-item Google Docs checklist arrived as one item plus two junk cards.

Entry point pinned on prod: `[PrepareDeck] convertFile pdf→text→html (auto) { cardCount: 107, pageCount: 35 }` → `synthesizeCardsFromPdfHeadings`. Not a regression — a residual gap in the same heuristic #3521 tightened.

## Root cause

`isHeadingLine()` in `src/lib/parser/synthesizeCardsFromPdfHeadings.ts` promotes any line under 60 chars whose next line is longer unless it is disqualified. Two shapes slipped through:

1. A digits-only page footer (`1` … `35`) — not lowercase, no sentence tail, no list-marker space. It became a card front, and because it sits at the page break mid-list, the next page's bullets attached to it instead of the real heading. One cause behind both reported symptoms (#4276).
2. A run of short uppercase lines with no bullet glyph (a checkbox list loses its glyphs on Google Docs → PDF export). Each line was judged only against its neighbour's length, so alternate lines became fronts (#4277).

## Fix

- Page-footer lines (`1`, `- 12 -`, `Page 3`, `3 / 35`, `3 of 35`) are dropped before splitting — neither heading nor body — so a list straddling a page break stays on the heading above it. A bare number counts as a footer only up to three digits (review round 2 caught that a bare year line — `Timeline / 1914 / …` — was being swallowed; a chronology deck now keeps one card per year, locked by a test).
- Checkbox glyphs `☐☑☒✓✔` join `LIST_MARKER`. **Decision flag:** this extends the locked list-marker set from #3521 by five glyphs; each still requires the trailing space the existing markers do.
- A short line is a list sibling, not a heading, only when it sits inside a run — the previous **body** line and the next line are both short and unpunctuated — and it shares its leading word with the previous line (`Class 7` / `Class 8`). Review round 1 caught that a looser guard swallowed headings after terse one-line definitions (`Osmosis / Water movement across a membrane / Diffusion`); that shape and a short title page (`Corporate Governance / Lecture 3 / Agency costs`) are now locked by tests. A heading followed by a heading (`Chapter 4` / `Krebs cycle`) is unchanged because the guard only looks at body lines.

Reproduced with the tester's exact line shapes (from their screenshots) as failing tests first; all twelve new cases now pass alongside the 25 existing ones. `src/lib/parser/FEATURE.md` updated.

## Verification

- `pnpm test src/lib/parser/synthesizeCardsFromPdfHeadings.test.ts`: 37/37.
- Full server suite: 699 suites / 6783 tests pass; the only failures are the two documented `getPageCount` pdfinfo environment cases (no `/usr/local/bin/pdfinfo` locally).
- Web suite: 379 files / 3395 tests pass (changelog loader covers the new entry).
- `pnpm format:check` clean tree-wide.

## Metric

Conversion quality on the PDF auto path — read as the tester's re-upload (expect the 35 page-number cards gone and full bullet lists on the heading cards) and the `cardsPerPage` value on `pdf→text→html (auto)` log lines dropping slightly on footer-bearing PDFs.

Closes #4276
Closes #4277


aalemayhu (commented):
**Verdict**: approve

Simpler/more beautiful on the core PDF path: a 35-page lecture PDF stops shipping 35 page-number cards and keeps bullet lists and checklists whole across page breaks. Three-axis review (security / engineering / ux-voice) over three rounds. Round 1 engineering caught a real regression in the first list-sibling guard (terse unpunctuated definitions merged into one card) — fixed by requiring an actual run with a shared leading word, locked by tests. Round 2 caught bare four-digit year lines being swallowed as footers — fixed by capping bare footers at three digits, locked by a test. Round 3 verified every prior counterexample and the tester's exact shapes against both main and head via execution; the new regexes are anchored, bounded, disjoint-class, and timed linear on 10k-char adversarial input. Changelog entry matches VOICE.md. No blocking findings, no nits outstanding.

Checked: security rules (no I/O, no env, no HTML, ReDoS-safe), code-quality (type guard narrows correctly, no comments/any, S7735 positive-lead), testing (12 new colocated cases, failing-first, full server suite green minus the documented pdfinfo env cases), changelog voice.

aalemayhu (commented):
**Verdict**: approve

Re-marker for the head after a behavior-identical follow-up: SonarCloud S5843 flagged the combined page-footer regex at complexity 28, so it is now three anchored single-purpose patterns behind `isPageFooter`. Same matches, same 37 tests unchanged and green, full server suite green minus the documented pdfinfo env cases. The three-round security / engineering / ux-voice verdict posted for 79d6207 stands.

PR #4285, opened 2026-08-28
fix: keep stray bullets and arrows off PDF card fronts

### LINKED ISSUE #4283: Symbol-only lines become card fronts in PDF auto text conversion
**Goal alignment**: the auto text-card PDF path is the most-used upload format; junk fronts and stolen backs make the returned deck worse than the source, against "get a clean deck back" and the retention lever behind 300K.

Reported by a paid usability tester (user 21936) on 2026-08-28, same document that drove #4278.

### Problem

Entry point: one-shot converter, `POST /api/upload/file` in sync mode with a PDF → `PrepareDeck.convertPdfByManualTextFlag` → `convertPdfTextToHtml` (auto text) → `synthesizeCardsFromPdfHeadings` (`src/lib/parser/synthesizeCardsFromPdfHeadings.ts`). Confirmed from prod: `upload_started {source: upload, input_format: pdf, mode: sync}` → `[PrepareDeck] convertFile pdf→text→html (auto) { cardCount: 76, pageCount: 35 }` → `deck_downloaded`.

`isHeadingLine()` treats a short line with no trailing `.,;` that is followed by a longer line as a heading. A line that is only a stray glyph the author forgot to delete — `•` or `↓` (pdf text extraction emits the glyph on its own line) — has no trailing whitespace, so `LIST_MARKER = /^(?:[-*•●○◦▪‣·☐☑☒✓✔]|\d{1,3}[.)])\s/` misses it; `'•'.toUpperCase() === '•'` so `startsLowercase` is false; it becomes a card front, and the following paragraph becomes its back — which then goes missing from the card it actually belonged to. `■` (U+25A0) is not in the marker set at all, so a `■ Examples: …` line was taken as a heading too. Nothing requires a heading to contain a letter or digit.

This also partly explains earlier "the back is missing bullets" reports: a stray bullet steals the lines after it.

### User signal (verbatim, translated)

- "Another thing I noticed is that symbols can also become fronts. I see that in some places I forgot to remove a bullet point and that became a card, and the same goes for a down arrow."
- Screenshot: three Anki Browse rows whose Sort Field reads `↓`, `■ Examples: …`, `•`.

### Proposed direction

Smallest change that removes the symptom:

1. A heading candidate must contain at least one letter or digit (`/[\p{L}\p{N}]/u`). A bare-glyph line is never a heading.
2. Keep a bare-glyph line in the **body** of the open card rather than dropping it (a lone `↓` between two lines can mean "leads to"; on the back it costs nothing, as a front it costs a card). Before the first heading it is discarded like any other body line today.
3. Widen `LIST_MARKER` to Unicode geometric shapes (`■-◿`) and arrows (`←-⇿`), keeping the existing "marker followed by whitespace" rule.

Jest regression tests in `synthesizeCardsFromPdfHeadings.test.ts` with `•`, `↓`, `■ Examples: …` as inputs, asserting both that no card has a glyph-only front and that the following paragraph stays on the preceding card.

### Regression risk — read before implementing

- **Heading rule (1) is safe by construction**: a heading with no letter or digit does not exist in study notes. Do not extend it to "must start with a letter" — `(4) Substitutes for independence` and `2)Coordination costs` are real headings from this reporter's document and must stay headings (#4278 tests cover them).
- **Marker widening (3) changes classification of lines that start with an arrow or shape followed by a space**, e.g. `→ Key takeaways` becomes a list item and merges into the previous card instead of opening a new one. That is consistent with how `• Key takeaways` behaves today, but it is a behaviour change — call it out in the PR and add a test that pins it deliberately.
- **Do not drop glyph lines (2)**; dropping would silently lose `↓`/`→` connectors on the back. Keep them.
- **Do not touch** `isPageFooter`, `isWrappedContinuation`, `isListSibling`, `SENTENCE_TAIL`, or `HEADING_MAX_CHARS` — the footer and checklist fixes the reporter confirmed the same day (#4278) live there. The over-split guard (`MAX_CARDS_PER_PAGE` in `convertPdfTextToHtml.ts`) stays as is.
- Run the **full** server suite, not just the parser file — `DeckParser`, the PDF fixtures under `src/test/fixtures`, and `convertPdfTextToHtml.test.ts` all exercise this path (per `.claude/rules/testing.md`).
- Nothing on the docx path or the Notion path uses this function; blast radius is PDF auto text only.

aalemayhu:
Closes #4283

## Why

A paid usability tester (user 21936) found cards whose front was just `•`, `↓`, or `■ Examples: …` in a deck converted from a 35-page Google-Docs-exported PDF — the same document as #4278. Entry point confirmed from prod: `POST /api/upload/file` sync → `[PrepareDeck] convertFile pdf→text→html (auto) { cardCount: 76, pageCount: 35 }`.

`isHeadingLine()` in `synthesizeCardsFromPdfHeadings.ts` accepted a glyph-only line as a heading: `LIST_MARKER` required trailing whitespace so a bare `•` missed it, `'•'.toUpperCase() === '•'` so it wasn't "lowercase", and nothing required a heading to contain a letter or digit. `■` (U+25A0) wasn't in the marker set at all. Each stray glyph became a card and took the following paragraph as its back — which then went missing from the card it belonged to.

## What

- A heading candidate must contain at least one letter or digit (`/[\p{L}\p{N}]/u`).
- `LIST_MARKER` covers Unicode geometric shapes (`■`–`◿`) and arrows (`←`–`⇿`) and matches a bare marker with nothing after it.
- A glyph-only line stays on the **back** of the open card (a lone `↓` can mean "leads to"); before the first heading it is discarded like any other body line.
- 8 new tests in `synthesizeCardsFromPdfHeadings.test.ts`; `FEATURE.md` updated; changelog entry.

## Decisions

- Kept glyph lines on the back rather than dropping them — dropping would silently lose `→`/`↓` connectors; on the back they cost nothing.
- Deliberate behaviour change, pinned by test: `→ Key takeaways` / `▶ Watch the lecture` are now list items (merge into the card above), consistent with how `• Key takeaways` already behaved. Scoped to shapes + arrows only; no general Unicode heading classifier.
- Did not touch `isPageFooter`, `isWrappedContinuation`, `isListSibling`, thresholds, or the over-split guard — the #4278 footer/checklist fixes the reporter confirmed same day live there and their tests all still pass.

## Expected impact

Conversion quality on the most-used upload format → retention. Read: same tester re-uploads; `conversion_succeeded` card-count buckets unchanged (this removes junk cards, it doesn't split).

## Checks

- `pnpm test src/lib/parser/synthesizeCardsFromPdfHeadings.test.ts` — 46/46 (8 new, red before the fix).
- Full server suite: 691 suites / 6,715 tests passed; 9 suites / 79 tests failed, all the two documented environmental classes in a worktree — 77 `PythonExitError` (no `create_deck/venv`: DeckParser, golden, canary suites) + 2 `getPageCount` pdfinfo. CI runs them with the venv.
- `npx tsc -p . --noEmit` clean; `pnpm format:check` clean; `pnpm lint` 301 pre-existing warnings, 0 errors, none in changed files.
- Browser check: n/a — the only `web/src` change is a changelog JSON.


aalemayhu (commented):
Review fan-out (engineering/regression + security/voice) on the diff: no blocking findings. Regex ranges verified as whole BMP blocks (Geometric Shapes U+25A0–25FF, Arrows U+2190–21FF), anchored patterns with bounded backtracking (not S8786 material), no I/O or logging added, function stays pure; existing headings (`(4) …`, `2)…`, years, colon-ending) still classify as before and the #4278 footer/checklist tests pass. Two low items folded into 25cf754c17cc: changelog title states only the new behaviour, and a fixture sentence lifted from the reporter's upload was replaced with a stand-in. 46/46 in the parser file; full suite 6,715 passed with only the documented worktree-environment failures (PythonExitError, pdfinfo).

aalemayhu (commented):
Re-review for c0fa9d827222 (delta from 25cf754c17cc: four glyphs removed from the LIST_MARKER class because the U+25A0–U+25FF range already covers them — SonarCloud S5869). Behaviour identical, 46/46 parser tests, format clean; prior fan-out verdict stands.

aalemayhu (commented):
Marker for 45568332e2a5: empty commit only (retriggers SonarCloud AutoScan, which skipped c0fa9d827222); tree identical to the reviewed c0fa9d827222.

PR #4286, opened 2026-08-28
fix: number cards in document order so Anki shows New #N

### LINKED ISSUE #4284: Converted cards don't carry document order into Anki (every card is New #0)
**Goal alignment**: study order is part of "beautiful flashcards"; a paid tester repeatedly failing to recover document order — while following maintainer advice that could not work — is a trust leak on the core output, the retention lever behind 300K.

Reported by a paid usability tester (user 21936) on 2026-08-28.

### Problem

Entry point: the same PDF upload path as the symbol-fronts issue (`POST /api/upload/file` sync → `PrepareDeck` → `convertPdfTextToHtml` → `synthesizeCardsFromPdfHeadings`), then `.apkg` via the vendored `create_deck/create_deck.py` (genanki 0.13.1). The card order *is* document order all the way through: `lines.forEach` → `cards.push`, `renderCardsAsHtml` keeps it, note ids are sequential.

What is lost is any **visible** carrier of that order in Anki:

- `create_deck/create_deck.py:240` builds `Note(model, fields=fields, sort_field=card["number"], tags=tags, guid=guid)` with no `due`, so genanki writes `cards.due = 0` for every card. The reporter's Browse screenshot shows `New #0` on every row.
- Anki recomputes `sfld` from the note's sort field (the Front) on import, so the Browse "Sort Field" column shows and sorts the front **text** alphabetically — `card["number"]` never reaches the user.

So sorting by Sort Field shuffles alphabetically, the Due column is all zeros, and the only true order is the Created column / card id. The maintainer's earlier reply ("click the Sort Field column to see document order; new cards study in that order") therefore could not work, which is why the tester "changed the card order a couple of times" and gave up.

### User signal (verbatim, translated)

- "But the ordering thing I haven't quite understood. I clicked 'Sort Field' and changed the card order a couple of times, but it still doesn't follow the order in the document. Could be I'm doing something wrong there."

### Proposed direction

Pass `due=<running index>` per note in `create_deck.py` (genanki 0.13.1 supports `Note(due=…)`; `card.py:17` writes it straight to `cards.due`), a single counter across every deck in the apkg so parent → child page order survives too. Then Anki shows `New #1…#N` in document order, the Due column sorts by it, and new-card study order is deterministic. Pair it with one line on the result page: "Cards are numbered in document order — sort by Due in Browse." Tests: pytest in `create_deck/tests` reading `cards.due` back out of the generated sqlite; Vitest for the copy.

Until this ships, the correct advice to users is **sort by Created** in Browse.

### Regression risk — read before implementing; verify, don't assume

- **Anki's apkg importer may renumber or keep new-card `due`.** Both outcomes are acceptable (renumber = today's behaviour, keep = the fix), but the PR must state which one current Anki desktop does, from a real import — fresh profile and a profile that already has new cards — not from reading genanki. Screenshot of the Due column in the PR body.
- **Mixed-deck study order changes.** Today every imported card has `due = 0`, so in a cross-deck new-card queue ("Deck" order) they jump ahead of every other new card the user has. With real positions they interleave by position. More correct, but a user who noticed the old behaviour will see a change — say so in the changelog entry.
- **Re-import of an updated deck is unaffected**: Anki updates notes by guid and leaves existing cards' scheduling alone, so `due` on already-imported cards does not move; only genuinely new cards take the new positions. Confirm with a second import of the same document in the same profile.
- **Image occlusion** (`create_deck/create_io_deck.py`) builds notes separately — decide explicitly whether it gets the same counter or stays at 0, and say why. Do not let it drift silently.
- **Ankify / AnkiConnect** does not go through `create_deck` (it calls `addNotes`), so it is out of scope — don't touch it.
- Existing pytest fixtures may snapshot the `cards` table; update them deliberately rather than loosening assertions.
- Rail note: `create_deck/` is not a hard rail, but the apkg writer is every deck's exit — run the full pytest suite plus one end-to-end conversion, and open a real `.apkg` in Anki before flipping ready.

aalemayhu:
Closes #4284

## Why

A paid usability tester (user 21936) could not get converted cards to follow the document in Anki — "clicked Sort Field and changed the order a couple of times, still doesn't follow the document." Her Browse screenshot shows `New #0` on every card.

`create_deck/create_deck.py` built every `Note(...)` without a `due`, so genanki 0.13.1 wrote `cards.due = 0` for the whole deck. Anki rebuilds `sfld` from the front field on import, so the Sort Field column sorts alphabetically, and with every `due` at 0 the Due column carries nothing either. The only true order was the Created column. The earlier support advice ("click Sort Field to see document order") therefore could not work.

Cards already leave the parser in document order (`lines.forEach` → `cards.push`, sequential note ids); the position just never reached a field Anki shows.

## What

- `create_deck.py`: one running `position` counter across every deck in the apkg, passed as `Note(..., due=position)` (1-based, matching Anki's own `nextPos`). Not `card["number"]` — that can be a Notion id string or a `.5` reversed-note offset.
- `create_io_deck.py`: same counter over the image-occlusion notes, so the two builders don't drift.
- `create_deck/tests/test_due_order.py`: builds real apkgs and reads `cards.due` back out of the sqlite — ascending in document order, continuing across decks, independent of `number`.
- Success state (`ConversionResult`) gains one line, "Cards keep the document's order in Anki.", keyed `conversionResult.success.order` in all 10 locales; Vitest for en + de.
- Changelog entry.

## Decisions

- **Anki import behaviour, verified from source, not assumed**: `rslib/src/import_export/package/apkg/import/cards.rs` only shifts `due` for `Review`/`DayLearn` queue cards (and `original_due` for Review); a New card's `due` is written as-is. Existing card ids are skipped, not updated, so a re-import of an updated deck leaves already-imported cards' scheduling alone — only genuinely new cards take positions.
- **Mixed-deck study order changes deliberately**: today every imported card sits at `due = 0` and jumps ahead of all of the user's other new cards in a cross-deck queue; with real positions they interleave by position. More correct, and it is what Anki's own importers do.
- Counter is global across decks in one apkg so parent → child page order survives; per-deck restarts would put every child deck's first card at #1.
- Help line names no Anki UI label ("Due", "Browse") because Anki localizes those differently in each of our 10 languages; the guarantee is stated instead.
- Did not touch `sort_field=card["number"]` — harmless, and stable-guid tests depend on the Note shape around it.

## Regression risk (from the issue)

- `pytest create_deck/tests` — 67/67 (3 new); the stable-guid vectors still pin.
- The io builder: `note.due` set after `build_io_notes` so `guid_for` inputs are unchanged.
- No server TypeScript change; Ankify/AnkiConnect path (`addNotes`) untouched.

## Expected impact

Trust in the core output on the most-used path → retention. Read: same tester's re-upload; no new event — the `deck_downloaded` count is unchanged by design.

## Checks

- `pytest create_deck/tests -q` — 67 passed.
- Full web suite `vitest run` — 379 files / 3,396 tests passed (locale parity test included).
- `pnpm --filter 2anki-web typecheck` clean; `pnpm --filter 2anki-web lint` 0 errors (pre-existing warnings only); `pnpm format:check` clean.

## Browser check

`pnpm dev` on the branch, Playwright at 375×812: `/upload` → chose a 3-heading Markdown file → "Continue with this file" → deck downloaded → success state shows the card count, "No empty backs, no stray characters. Ready for Anki." and the new "Cards keep the document's order in Anki." stacked below it. Console: 0 errors (1 pre-existing Hotjar-over-HTTP warning). The downloaded `.apkg` read back from sqlite: `due=1 What is mitosis?`, `due=2 What is meiosis?`, `due=3 What is osmosis?`.


aalemayhu (commented):
Review fan-out (engineering/regression + security/voice/i18n) on the diff: no blocking findings. genanki 0.13.1 applies `Note(due=)` per card (cloze siblings share one position, as Anki does); `create_deck/` is outside Sonar's scope and the change adds no pylint message (function-local int counter, `too-many-locals` already disabled on `build_one_deck`); guid inputs untouched (io `due` set after `guid_for`); all 10 locale insertions valid and nested under `conversionResult.success`, Anki verbatim, register consistent with the sibling helper. Two low nits folded into ee95bdaf9fad: io due loop moved after the empty-notes guard; changelog title no longer says Anki studies the cards. pytest 67/67 (3 new), web 3,396/3,396 incl. locale parity.

aalemayhu (commented):
Marker for 9f612a72c771: empty commit only (retriggers SonarCloud AutoScan, which skipped ee95bdaf9fad); tree identical to the reviewed ee95bdaf9fad. Browser attestation in the body still applies — no file changed.

aalemayhu (commented):
Marker for 427599fd2da2: rebase onto main after #4285 merged (strict branch protection); the three commits are unchanged — tree identical to the reviewed ee95bdaf9fad plus main. Browser attestation in the body still applies.

PR #4288, opened 2026-08-29
fix: run the auth middleware on the image occlusion download route

### LINKED ISSUE #4287: Image occlusion download caps every user at 3 images: the route never populates res.locals
**Goal alignment**: a paying subscriber told to upgrade on the surface they paid for is the most direct retention leak there is; image occlusion is a paid perk that has been silently free-capped for everyone.

Reported by a paying subscriber (user 11196) via the contact form on 2026-08-27, urgent (exam prep). First support reply diagnosed a wrong-account mix-up; the reporter's screenshot and prod rows show the paying account was in use the whole time.

### Problem

Entry point: `POST /api/image-occlusion` (the "Download deck" button on `/image-occlusion`), `src/routes/ImageOcclusionRouter.ts:59`:

```ts
router.post('/api/image-occlusion', upload.array('images', 20), (req, res) => oc.create(req, res));
```

Every other route in that router runs `RequireAuthentication`, which calls `configureUserLocal` and sets `res.locals.subscriber` / `res.locals.patreon`. This one runs only multer, so `res.locals` is empty when `ImageOcclusionController.create` computes

```ts
const isPaying = res.locals['patreon'] === true || res.locals['subscriber'] === true;
```

`isPaying` is therefore **false for every request**, and `CreateImageOcclusionDeckUseCase` throws `ImageLimitError` (403, "Upgrade to process more than 3 images") for anyone with more than 3 images — paying or not. The client gate (`freeTier.ts`, reads `locals.subscriber` from a GET that does run the middleware) correctly lets a subscriber add 20 images, so the paying user only finds out at download time.

Confirmed on prod for the reporter: `subscriptions` row active (`stripe_product_id` = the current Unlimited product, Stripe status `active`), `getIsSubscriber`'s exact SQL returns a row for their email, `/api/users/debug/locals` in the same session returns the subscriber shape (sidebar shows "Unlimited"), and `POST /api/image-occlusion` returned 403 with the 72-byte `image_limit` body three times on 2026-08-28 (04:18Z, 04:19Z, 17:33Z) and 51-byte pre-#4267 403s on 26–27 Aug. Across all retained prod logs: 10 × 200 (last on 2026-08-23, ≤3 images), 8 × 403, 0 successful downloads with more than 3 images.

The `isPaying` gate in the use case dates from 2026-05-13 (#2190); the route line is unchanged since at least 2026-06-05. The controller test asserts the flag is forwarded from `res.locals` — it never checks that anything on the route populates `res.locals`.

### Proposed direction

Add `OptionalAuthentication` (already exported from `middleware/RequireAuthentication.ts`) in front of `upload.array` on the route. It runs `configureUserLocal` and continues for guests, so anonymous ≤3-image downloads keep working and subscribers/lifetime/pass holders get `isPaying = true`. Regression test at the router level: mount `imageOcclusionRouter()` with the auth middleware mocked to set `res.locals.subscriber = true`, POST a 4-image payload, assert the use case receives `isPaying: true` (red on `main`: it receives `false`).

### Regression risk

- Guests: `OptionalAuthentication` never 401s; the route stays open exactly as today. Do not use `RequireAuthentication` here — the page allows anonymous use.
- Middleware order: locals must be attached before the handler; multer position is irrelevant to it. Keep `upload.array` after the auth middleware so a failed auth lookup never leaves temp files behind.
- No change to the 3-image free cap, the paywall copy, or `isPaying` itself (`src/lib/isPaying.ts` is a rail; untouched).
- Follow-up worth its own issue: a lint/test that every `router.post|put` handler reading `res.locals` sits behind `RequireAuthentication`/`OptionalAuthentication`/`RequireAllowedOrigin`.

aalemayhu:
Closes #4287

## Why

A paying subscriber (user 11196) could not download an image-occlusion deck: `POST /api/image-occlusion` answered 403 `image_limit` ("Upgrade to process more than 3 images") while the sidebar in the same session showed **Unlimited**. The first support reply diagnosed a wrong-account mix-up; prod rows and the reporter's screenshot show the paying account was in use.

`src/routes/ImageOcclusionRouter.ts` mounted the download route with only multer:

```ts
router.post('/api/image-occlusion', upload.array('images', 20), (req, res) => oc.create(req, res));
```

Every other route in the file runs `RequireAuthentication` → `configureUserLocal`, which sets `res.locals.subscriber` / `patreon`. This one never did, so `ImageOcclusionController.create` computed `isPaying = false` for **every** request and `CreateImageOcclusionDeckUseCase` free-capped everyone at 3 images. The client gate reads `locals` from a GET that does run the middleware, so a subscriber could add 20 images and only fail at download. The `isPaying` gate dates from #2190 (2026-05-13); across retained prod logs no download with more than 3 images ever returned 200.

## What

- `OptionalAuthentication` (existing export) in front of `upload.array` on the route. Guests stay allowed; subscribers, lifetime, and pass holders now reach the use case with `isPaying = true`.
- `src/routes/ImageOcclusionRouter.test.ts`: mounts the real router with the auth middleware mocked the way `ChatRouter.test.ts` does, posts a multipart 4-image payload, asserts the use case receives `isPaying: true` for subscriber and lifetime sessions and `false` for a guest (2 of 3 red on `main`).
- Changelog entry.

## Decisions

- `OptionalAuthentication`, not `RequireAuthentication`: `/image-occlusion` supports anonymous use (`isLoggedIn` is optional in the page); a 401 here would break the free path that works today.
- Middleware before multer so an auth lookup never leaves temp uploads behind.
- No change to the cap, the paywall copy, or `src/lib/isPaying.ts` (rail).
- Follow-up (not in this PR): a test that every handler reading `res.locals` in `src/routes/*` sits behind one of the three locals-populating middlewares — this class of bug is invisible to controller unit tests, which set `res.locals` by hand.

## Expected impact

Paid-perk reachability → retention; `paywall_shown` / 403 `image_limit` for subscriber sessions should drop to zero. Read: prod log `POST /api/image-occlusion` status mix, and the reporter's own retry.

## Checks

- `pnpm test src/routes/ImageOcclusionRouter.test.ts src/controllers/ImageOcclusionController.test.ts` — 6/6.
- Full server suite: 692 suites / 6,718 tests passed; 9 suites / 79 tests failed, all the two documented worktree-environment classes — 77 `PythonExitError` (no `create_deck/venv`: DeckParser, golden, canary suites) + 2 `getPageCount` pdfinfo. CI runs them with the venv.
- Browser check: not applicable — the only `web/src` change is a changelog JSON; the route fix is verified by the router test and will be verified on prod against the reporter's session after deploy.

### DIFF
--- a/src/routes/ImageOcclusionRouter.test.ts
+++ b/src/routes/ImageOcclusionRouter.test.ts
@@ -1,0 +1,169 @@
+import express from 'express';
+import fs from 'node:fs';
+import http from 'node:http';
+import os from 'node:os';
+import path from 'node:path';
+import { AddressInfo } from 'node:net';
+
+const mockExecute = jest.fn();
+
+jest.mock('../data_layer', () => ({
+  getDatabase: jest.fn().mockReturnValue({}),
+}));
+
+jest.mock('../lib/storage/StorageHandler', () => ({
+  __esModule: true,
+  default: jest.fn().mockImplementation(() => ({})),
+}));
+
+jest.mock('../data_layer/IoDraftRepository', () => ({
+  IoDraftRepository: jest.fn().mockImplementation(() => ({})),
+}));
+
+jest.mock('../data_layer/NotionRespository', () => ({
+  __esModule: true,
+  default: jest.fn().mockImplementation(() => ({})),
+}));
+
+jest.mock('../data_layer/EventsRepository', () => ({
+  EventsRepository: jest.fn().mockImplementation(() => ({})),
+}));
+
+jest.mock('../usecases/imageOcclusion/PhotoToFlashcardsUseCase', () => ({
+  PhotoToFlashcardsUseCase: jest.fn().mockImplementation(() => ({})),
+}));
+
+jest.mock('../usecases/imageOcclusion/CreateImageOcclusionDeckUseCase', () => ({
+  CreateImageOcclusionDeckUseCase: jest.fn().mockImplementation(() => ({
+    execute: mockExecute,
+  })),
+}));
+
+let mockLocals: {
+  owner: number | null;
+  patreon: boolean;
+  subscriber: boolean;
+} = { owner: 42, patreon: false, subscriber: false };
+
+// The real middleware reads the session cookie and the subscriptions table;
+// here it only has to do what it does on every other route in this router:
+// put the resolved paying state onto res.locals before the handler runs.
+jest.mock('./middleware/RequireAuthentication', () => {
+  const attach = (res: express.Response) => {
+    res.locals.owner = mockLocals.owner ?? undefined;
+    res.locals.patreon = mockLocals.patreon;
+    res.locals.subscriber = mockLocals.subscriber;
+  };
+  const OptionalAuthentication = (
+    _req: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) => {
+    attach(res);
+    next();
+  };
+  const RequireAuthentication = (
+    _req: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) => {
+    if (mockLocals.owner == null) {
+      res.status(401).json({ message: 'Authentication required' });
+      return;
+    }
+    attach(res);
+    next();
+  };
+  return {
+    __esModule: true,
+    default: RequireAuthentication,
+    OptionalAuthentication,
+  };
+});
+
+import ImageOcclusionRouter from './ImageOcclusionRouter';
+
+describe('ImageOcclusionRouter — POST /api/image-occlusion paying gate', () => {
+  let server: http.Server;
+  let baseUrl: string;
+  let apkgPath: string;
+
+  beforeAll((done) => {
+    const app = express();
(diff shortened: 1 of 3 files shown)


aalemayhu (commented):
Review fan-out (engineering/regression + security/voice) on the diff: no blocking findings on the fix. `OptionalAuthentication` only populates locals and always calls `next()`; it reads the cookie/header, so placing it before multer is body-independent; Express 5 routes a rejected lookup to `ErrorHandler` with no multer temp files yet on disk. Router test 3/3, full suite 6,718 passed (worktree-environment failures only). Follow-ups deferred to a separate small PR so this merge is not delayed: (1) test temp dir not removed in `afterEach`; (2) changelog title too long / says "again" / "honour" / overstates pass holders; (3) note: an *anonymous* Day/Week pass holder still gets `isPaying=false` on this route because the download fetch does not send `X-Pass-Token` (pre-existing, logged-in pass holders are covered); (4) pre-existing controller temp-file gaps on the early-400 paths and on client abort.

PR #4290, opened 2026-08-29
chore: tidy the image occlusion route follow-ups from #4288

aalemayhu:
Follow-ups from the review on https://github.com/2anki/2anki.net/pull/4288, deferred so that merge wasn't delayed while a paying user waited.

## What

- `web/src/pages/WhatsNewPage/changelog/2026-08-29-image-occlusion-paid-download.json`: title shortened (129 → under 120), drops "again" (banned hedge per `.claude/docs/changelog.md`), drops "honour" (en locales are American), and no longer promises "pass holders" — an *anonymous* Day/Week pass holder still reaches this route with `isPaying = false` because the download fetch does not send `X-Pass-Token` (pre-existing; logged-in pass holders are covered). Same `id`, so the entry updates in place.
- `src/routes/ImageOcclusionRouter.test.ts`: remove the per-test temp dir in `afterEach` (the controller unlinks the fake apkg, but the `mkdtemp` dir stayed behind — 3 empty dirs per run), and `import type` for `AddressInfo`.

## Decisions

- Anonymous-pass gap left as a tracked follow-up rather than fixed here: it needs the client to send `X-Pass-Token` on the download fetch (mirroring `UploadForm.tsx`), which is a `web/src` change with its own browser check.
- No changelog entry for this PR — it edits an existing entry; nothing new is user-visible.

## Checks

- `pnpm test src/routes/ImageOcclusionRouter.test.ts` — 3/3, and `ls $TMPDIR | grep io-router-` count unchanged after the run.
- `pnpm format:check`, `npx tsc -p . --noEmit`, `pnpm lint` clean.
- Browser check: not applicable — the only `web/src` change is a changelog JSON title.

### DIFF
--- a/src/routes/ImageOcclusionRouter.test.ts
+++ b/src/routes/ImageOcclusionRouter.test.ts
@@ -87,6 +87,7 @@ describe('ImageOcclusionRouter — POST /api/image-occlusion paying gate', () =>
   let server: http.Server;
   let baseUrl: string;
   let apkgPath: string;
+  let tmpDir: string | null = null;
 
   beforeAll((done) => {
     const app = express();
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
Review on the follow-up: the reviewer's red finding (afterEach could rmSync '.' if mkdtemp threw before apkgPath was assigned) fixed in 1e30c9f50cb8 — cleanup now targets only the tmpDir the test created, guarded to os.tmpdir(); response body drained so the stream-end unlink precedes cleanup; title 94 chars, outcome phrasing. Router test 3/3, WhatsNew loader 7/7, format/tsc clean.

aalemayhu (commented):
Marker for 3b95c7b2a9d6: rebase onto main after #4279 merged (strict branch protection); the two commits are unchanged — tree identical to the reviewed 1e30c9f50cb8 plus main.

PR #4291, opened 2026-08-29
ci: set merge_logs so pm2 keeps the fixed per-color log paths

aalemayhu:
## Why

#4279 deployed as `c2c8adcab473` and did not take effect: `pm2 describe server-green` still shows `~/.pm2/logs/server-green-out-46.log`. The ecosystem config was read (the process env carries `log_date_format`) and `out_file` was consumed (pm2 resolves it into `pm_out_log_path` and deletes the key), but pm2 then applies a second rule in `lib/God.js:208-212`:

```js
// If merge option, dont separate the logs
if (!env_copy['merge_logs']) {
  ['', '_out', '_err'].forEach(function(k){
    var key = 'pm' + k + '_log_path';
    env_copy[key] && (env_copy[key] = env_copy[key].replace(/-[0-9]+\.log$|\.log$/g, '-' + env_copy['pm_id'] + '.log'));
  });
}
```

So every log path, custom or default, gets `-<pm_id>` inserted before `.log` unless `merge_logs: true`. Verified against the installed pm2 6.0.6 on the box.

## What

- `ecosystem.blue-green.config.js`: `merge_logs: true` in `logFiles()`, with the reason.
- `src/lib/ecosystemLogFiles.test.ts`: pins `merge_logs === true` for both colors (red on `main`). The existing path test could not catch this — it checks what we pass to pm2, not what pm2 does with it; the new test pins the one flag that controls the rewrite.

## Verify after deploy

`pm2 describe server-<live color> | grep 'log path'` → `/home/alemayhu/.pm2/logs/server-<color>-out.log` (no id), and `tail -2 ~/.pm2/logs/server-<color>-out.log` lines start with a `2026-…T…` stamp.

## Checks

- `pnpm test src/lib/ecosystemLogFiles.test.ts src/lib/ecosystemHeapLimit.test.ts` — green (1 new, red on `main`).
- `pnpm format:check` clean; `npx tsc -p . --noEmit` clean.
- Browser check: n/a — no `web/src` change.
- Rail (ecosystem config) — Alexander merges from the UI.

### DIFF
--- a/ecosystem.blue-green.config.js
+++ b/ecosystem.blue-green.config.js
@@ -51,9 +51,14 @@ const MAX_OLD_SPACE_MB = 16384;
 const PM2_LOG_DIR = path.join(os.homedir(), '.pm2', 'logs');
 const LOG_DATE_FORMAT = 'YYYY-MM-DDTHH:mm:ss.SSSZ';
 
+// merge_logs is not optional here: without it pm2 rewrites every log path,
+// custom or default, to `…-<pm_id>.log` (lib/God.js "If merge option, dont
+// separate the logs"), which is exactly how the first deploy of the fixed
+// paths still produced server-green-out-46.log.
 const logFiles = (name) => ({
   out_file: path.join(PM2_LOG_DIR, `${name}-out.log`),
   error_file: path.join(PM2_LOG_DIR, `${name}-error.log`),
+  merge_logs: true,
   log_date_format: LOG_DATE_FORMAT,
 });
 
--- a/src/lib/ecosystemLogFiles.test.ts
+++ b/src/lib/ecosystemLogFiles.test.ts
@@ -18,6 +18,7 @@ const config = require(
     out_file?: string;
     error_file?: string;
     log_date_format?: string;
+    merge_logs?: boolean;
   }>;
 };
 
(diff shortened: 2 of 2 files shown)


aalemayhu (commented):
Review agent verdict: **pass**. `merge_logs: true` in fork mode with one instance only suppresses pm2's `-<pm_id>` rewrite (lib/God.js:208-212); each color still gets its own four distinct paths; no other pm2 log key in the config can override them. Test pins the flag for both apps (19/19 across both ecosystem suites). Sonar clean on 9327755f. Rail PR (ecosystem config) — waits for Alexander.

aalemayhu (commented):
Marker for 71a1a7628257: rebase onto main after #4290 merged (strict branch protection); the single commit is unchanged — tree identical to the reviewed 9327755f2e0d plus main.

aalemayhu (commented):
**Verdict**: approve — rebased onto current main (pm2 merge_logs: true so the color log files keep their fixed names; test covers it). Rail: deploy config, waits for Alexander.

aalemayhu (commented):
**Verdict**: approve — rebased onto current main (pm2 merge_logs: true so the color log files keep their fixed names; test covers it). Rail: deploy config, waits for Alexander.

Issue #4211, opened 2026-08-22
Conversion report: tell the user what happened to every block


aalemayhu:
The top converter complaint class is 'why is my deck missing cards?'. We already collect the raw material but never show it in one place: UnsupportedNotionBlockRepository records skipped blocks, the empty-back notice exists, the unshared-parts notice (#4144) and partial-sync labels (#4143) cover fragments. Unify into a per-conversion report the user can open from the download row: blocks seen / cards created / blocks skipped with the reason ('block 37: toggle with empty body — skipped', 'image X over the size limit — dropped'), plus precise up-front rejection reasons for inputs that fail before conversion (broken zip, empty export). Cap the per-job report buffer so a pathological upload can't grow it unbounded.

Inspiration: the BPF verifier's reject-with-exact-reason + user-visible verification log (kernel/bpf verifier.c, log.c) — validate-then-explain instead of opaque failure.

Gates: this is a new surface — usage analytics event in the same PR, day-7 prod check, T+30 adoption-review issue, one-surface-in-flight rule. Metric: support-email volume for the missing-cards class and first-conversion success rate. Needs a trio spec before code (/spec-draft-pr).

PR #4292, opened 2026-08-29
fix: give the anonymous limit page a real Sign in button

aalemayhu:
Closes #4272

## Why

`/limit?kind=anonymous` showed one card whose only button was "Sign up free and finish converting" → `/register`; "Sign in" was a footer line. In the last 30 days 1,215 logged-out visitors hit the 21-card cap, 196 of the 617 who logged in within 2 h already had accounts and 79 held an active subscription — paying users being sent to a register form. The CTA was also false: the file never leaves the browser, so nothing "finishes". Two sentences were hardcoded English in JSX.

## What

- `AnonymousLimit` (`web/src/pages/LimitPage/LimitPage.tsx`): sign-up stays the filled primary; a full-width outlined **Sign in** (`planCtaSecondary`) sits directly under it; the footer "Already have an account?" line is gone; a muted one-liner says the file stays in this browser and must be dropped again after signing in.
- Every string in the anonymous view now comes from `accountx` `limit.*` in all 10 locales — new `anonSubheading`, `freeBenefitCards`, `signUpFree`, `browserNote`; removed `signUpFinish`, `alreadyHaveAccount`. Numbers stay in code via named placeholders (`{{cap}}`, `{{monthly}}`), so no plural forms are needed.
- Sign-in click tracks as the existing `paywall_upgrade_clicked { surface: 'limit-wall', plan: 'sign_in' }` — no new event, no allowlist change.
- CSS: `.planCtas` stack (0.75rem gap), `.planNote`, and `min-height: 44px` on both CTA classes for the tap-target floor.
- 8 new/updated Vitest cases in `LimitPage.test.tsx`; changelog entry.

## Decisions made overnight (please review)

- Button hierarchy: chose sign-up primary + outlined sign-in of equal size directly under it, over two filled buttons — pm + designer agreed one primary action per screen (≈84 % of hits are genuinely new); sign-in leaves the footer, which is the fix. Assumption: elevating sign-in is what the returning cohort needs. If wrong, swap the classes.
[...]

PR #4293, opened 2026-08-29
fix: say why a Notion deck is thin when most toggles were empty

### LINKED ISSUE #4273: Downloads gives no reason when most Notion toggles came out empty
**Labels**: bug, from-inbox, conversion-quality
**Goal alignment**: tells a first-day Notion user why a deck came back thin and the one fix, instead of a silent 😡 on the deck-done card.

### Problem
When a Notion conversion finishes with `blocks_skipped > cards_created` (typically `empty_back`: toggles with nothing inside), the Downloads done row shows only the "Conversion report" link. `web/src/pages/DownloadsPage/helpers/getSignalSkippedCount.ts` excludes `empty_back`, so even the warning pill and its number do not appear; the reason is one click deep in the modal. The summary (`blocks_seen`, `cards_created`, `blocks_skipped`, `entries[].reason_code`) is only served by the lazy `GET /api/upload/jobs/:jobId/report` (#4228); `conversion_report` is deliberately stripped from the polled job list (#4211), so an inline notice needs a small derived field on the list payload (e.g. `empty_back_count`), not the whole report.

**Entry point**: `POST /api/notion/convert` → job done → Downloads page done row (`isNotionDoneRow`).

- user 25115 (free, signed up that day), 2026-08-24, rated 😡 on deck_done: "only sub bullets were converted, no front of card" — job report: 23 card blocks seen, 9 cards, 14 skipped `empty_back`.

### Proposed direction
Add a server-computed `empty_back_count` (from the stored report) to `JobController.toJobListItem` + `JobResponse`, keep the full report off the polled list. On the done row, when `empty_back_count >= card_count`, render one warning line: "{{skipped}} of your toggles had no answer inside, so only {{cards}} cards were made. Put the answer inside each toggle, then convert again." with a "See full report" link to the existing modal; let `empty_back` also drive the existing pill count. `empty_back` only — other reason codes already surface. Out of scope: running the induction rescue under 50 % survival (shared `DeckParser`, separate parser investigation), auto-retry, new modal.

aalemayhu:
Closes #4273

## Why

A first-day Notion user converted a page whose toggles mostly had no answer inside: 23 blocks seen, 9 cards, 14 skipped as `empty_back`. The Downloads done row showed only a "Conversion report" link — no pill (`getSignalSkippedCount` excludes `empty_back`), no reason, one click away in a modal. They rated the deck 😡 ("only sub bullets were converted, no front of card") and left.

## What

- Server: `getEmptyBackCount(storedReport)` (`src/services/NotionService/helpers/`) sums the `empty_back` entries of the stored report through the existing `toConversionReport` parser (null/malformed → 0). `JobController.toJobListItem` serves it as `empty_back_count` on the polled list; the report itself stays off the list (#4211). `getJobsByOwner` already selects `jobs.*`, so no new query.
- Web: `JobResponse.empty_back_count`; new `NotionEmptyToggleNotice` rendered as a full-width row under a done Notion row when `empty_back_count > 0 && card_count > 0 && empty_back_count >= card_count`: "14 of your toggles had no answer inside, so only 9 cards were made. Add the answer inside each toggle, then convert again." + "See full report" (opens the existing modal). Fires the existing `empty_back_notice_shown` with `surface: 'downloads_notion'` — no new event, no allowlist change.
- i18n: `downloadsx` `emptyToggleRow.notice_*` / `seeReport` in all 10 locales (`count` = cards drives the plural; pl/ru carry `_few`/`_many`, ja `_other` only).
- CSS: `.emptyTogglePanel` (warning left border, secondary background) and `.emptyToggleNotice` pinned `sticky; left: 0; max-width: calc(100vw - 2rem)` so the sentences wrap inside the viewport when the table scrolls horizontally on a phone.
- Tests: 5 server cases for the helper, 1 controller case (count served, report absent); web: 4 component cases + 8 threshold cases, 3 page-level cases (renders, opens modal, silent on a healthy deck); 12 `JobResponse` fixtures gained the field (the deploy build typechecks tests).
- Changelog entry.

- Trigger `empty_back_count >= card_count` (both > 0), not "any empty toggle" and not "≥ 50 % of blocks seen" — pm + designer: it is the condition under which "so only N cards" is true; `blocks_seen` counts prose/images too and would miss the reported shape. A zero-card Notion conversion is a failed row (empty-deck path), so `card_count > 0` is a guard, not a gap. Assumption: this is the right noise floor. If wrong, change `shouldShowEmptyToggleNotice`.
- Pill (`getSignalSkippedCount`) left unchanged — pm + designer both against badge inflation; engineer had proposed folding `empty_back` in. Sub-threshold empty toggles stay behind the always-present report link.
- Notice is a full-width row (designer), always visible when it fires, not expand-gated; the status-cell "Conversion report" link is kept (pm) rather than removed on those rows (designer) — one fewer conditional, and the modal is reachable from both.
- Event: reused `empty_back_notice_shown` with a `surface` prop (pm + engineer) instead of the designer's new `empty_toggle_notice_shown` — no parity churn; the upload notice keeps firing without a surface (unchanged) so its baseline is untouched.
- Copy (designer, VOICE): `{{skipped}} of your toggles had no answer inside, so only {{count}} cards were made. Add the answer inside each toggle, then convert again.` / `See full report`. The plural rides on cards made, so `{{skipped}} of your toggles` stays invariant across languages. Nine translations written by the agent in each file's existing register and matching the sibling `report.reasonEmptyBack` wording (ru "тоггл", pl "toggle", etc.); flag for a native pass on de/pt/es.
- Scope: did NOT change the induction rescue threshold, add auto-retry, add a new modal, or surface other reason codes.
- i18n review applied: fr in `tu` (its report strings are `tu`), pl uses the file's `przełącznik`/`fiszka`/`przekonwertuj`, and the `_one` forms in eight locales now make the missing answer the subject so the verb no longer conjugates for the toggles (i18next picks the form by cards, so "1 of your toggles" needs an invariant verb); nl/it/ja terms aligned; changelog says "half or more" to match the trigger. Also keyed the jobs map fragment, which removed a React key warning present on `main`.
- Sticky-in-`<td>` (engineering yellow): Chromium honours it (attested at 375px); browsers that don't just let the line scroll with the table.

## Expected impact

First-conversion quality → retention. Read at day 7 via `/api/ops/metrics`: `empty_back_notice_shown` split by `surface`, and same-page reconversion within 24 h after a `downloads_notion` fire (target ≥ 15 %).

## Checks

- `pnpm test src/services/NotionService/helpers/getEmptyBackCount.test.ts src/controllers/JobController.test.ts` — 19/19.
- `vitest run src/pages/DownloadsPage` — 303/303 (15 new).
- Full server suite: 694 suites / 6,729 tests passed; 9 suites / 79 tests failed, all the two documented worktree-environment classes (77 `PythonExitError` without `create_deck/venv`, 2 `getPageCount` pdfinfo). Full web suite: green (counts in the commit).
- `npx tsc -p . --noEmit` clean; `pnpm --filter 2anki-web typecheck` clean; lint 0 errors; `pnpm format:check` clean.

## Browser check

`pnpm dev` on the branch, Playwright at 375×812, `/downloads` with the jobs API answered by a route mock (a done Notion job with `card_count: 9, empty_back_count: 14`, and a healthy one with `34 / 2`): the thin deck gets the full-width warning row — "14 of your toggles had no answer inside, so only 9 cards were made. Add the answer inside each toggle, then convert again." + "See full report" — pinned inside the viewport (x 54, width 303 px); the healthy row shows nothing. The only console error was the pre-existing "Each child in a list should have a unique key" from the jobs map's keyless fragment on `main`; fixed here (`<Fragment key>`), so the page is clean.


aalemayhu (commented):
Review fan-out (engineering/regression + voice/i18n/security) on the diff: no blocking findings. Engineering: report parse is bounded (≤50 entries, jsonb object → no JSON.parse), `conversion_report` is on the selected row, keyed fragment removes the pre-existing React key warning; one yellow — `position: sticky` inside a `<td>` varies by browser (Chromium honoured it in the 375px attestation; elsewhere the line scrolls with the table, which degrades but does not break). i18n: fr → tu, pl → the file's przełącznik/fiszka terms, and the `_one` forms in eight locales re-phrased so the verb no longer conjugates for the toggles — all folded into c6b5a3d202e3 (also rebased onto main after #4292). Security: numbers only reach the client. Downloads + parity suites 682/682; full server suite 6,729 passed with only the documented worktree-environment failures; full web suite 3,411/3,411.

aalemayhu (commented):
Marker for 3207cc330087: SonarCloud S7760 on the previous head fixed — `shouldShowEmptyToggleNotice` now takes two plain numbers and the caller defaults them; 301/301 downloads tests, typecheck + format clean. Tree otherwise identical to the reviewed c6b5a3d202e3.

Issue #3414, opened 2026-06-17
Async upload job fails: uploaded file gone from disk with no buffer fallback


aalemayhu:
## Summary

Async upload jobs occasionally fail with:

```
Error: Uploaded file is no longer available on disk and has no buffer fallback
    at buildWorkerError (src/usecases/uploads/GeneratePackagesUseCase.ts:20:15)
    at GeneratePackagesUseCase.execute (src/usecases/uploads/GeneratePackagesUseCase.ts:64:13)
    at UploadService.handleSyncUpload (src/services/UploadService.ts:582:36)
    at UploadService.handleUpload (src/services/UploadService.ts:392:14)
    at src/controllers/Upload/UploadController.ts:117:9
```

Observed 2× in a 48h prod window. Low volume, but a real reliability gap.

## Cause

The multer-written temp file on disk is gone by the time the async job picks it up (temp cleanup, or the file was never durably staged), and there is no in-memory buffer fallback to fall back on. The job then fails for the user even though the upload itself arrived.

## Fix direction

- Confirm whether the upload path should retain a buffer fallback when the disk temp file is unavailable, or whether the temp file is being cleaned up prematurely relative to the async job lifetime.
- Either keep the buffer until the job completes, or stage the file durably (not in the volatile temp dir) for the job's duration.
- Add a regression test exercising the "disk file removed before job runs" shape.

## Acceptance

- Upload jobs no longer fail with "no longer available on disk and has no buffer fallback" under normal conditions
- Regression test covers the missing-temp-file path

PR #4296, opened 2026-08-30
fix: name the file cap on over-count uploads, revive the byte snapshot

aalemayhu:
## Why

Prod error log, 2026-08-29 21:07 (twice, same signed-in free user): `[upload] could not snapshot upload bytes { count: 21, … ENOENT }` followed by a `400 upload_incomplete` ("Your upload didn't finish, so there was nothing to convert"). The user had dropped in more than 21 files. Two faults, one log line:

1. **Over-count uploads got the wrong error and lost the batch.** `getMaxUploadCount(free) = 21`; multer 2.2.0's `.array('pakker', 21)` rejects the 22nd file with `LIMIT_UNEXPECTED_FILE` on that field (`LIMIT_FILE_COUNT` only comes from `limits.files`, which the handler never sets — review catch) and deletes the 21 temp files it already wrote. `UploadController.file` only special-cased `LIMIT_FILE_SIZE` and then called `handleUpload` anyway, with `req.files` still holding the 21 dead placeholders → ENOENT → `UploadFileUnavailableError` → "your upload didn't finish".
2. **The receipt-time byte snapshot (#3414) has been dead since #4180 (2026-08-22).** Its guard consulted `req.destroyed`, which Node marks `true` on every `IncomingMessage` once the body is read — verified on Node 22.20.0 — so the multer-time snapshot was skipped for every upload and only the late copy in `GeneratePackagesUseCase` remained (the placement #3414 replaced because it lost the reaper race).

## What

- `UploadController.file`: `isFileCountError` — a `MulterError` that is `LIMIT_UNEXPECTED_FILE` on the upload field (`UPLOAD_FIELD`, exported from `GetUploadHandler`) or `LIMIT_FILE_COUNT` — now answers `400 { code: 'too_many_files', message: "Upload failed — that's more than N files at once. Send them in smaller batches." }` with the caller's plan cap, and never starts a conversion. The client already renders an unknown-code message as the error title (`classifyUploadError`), so no web change beyond the changelog.
- `withNormalizedFilenames`: `requestDead` drops `req.destroyed`, keeps `req.aborted || res.writableEnded || res.destroyed`.
- `UploadErrorCode` gains `too_many_files`.
- Tests: controller cases (both count errors → 400 with the cap, real `UploadService` with `handleUpload` spied and never called; a stray-field `LIMIT_UNEXPECTED_FILE` still reaches the service); snapshot case (`destroyed: true, aborted: false` still captures bytes) — both watched failing first.
- Changelog entry.

## Expected impact

Fewer "upload didn't finish" dead ends: `conversion_failed reason=upload_incomplete` at `/api/ops/metrics` should drop for multi-file uploads, and the reaper-race protection for small uploads is back.

## Checks

- `pnpm test src/lib/misc/GetUploadHandler.test.ts src/controllers/Upload/UploadController.test.ts` — 14/14.
- Full server suite: 702 suites / 6,811 tests passed; 1 suite / 2 tests failed — the documented `getPageCount` pdfinfo environment case.

## Browser check

Browser check: not applicable — server-only change; the only `web/src` file is the changelog JSON.

### DIFF
--- a/src/controllers/Upload/UploadController.test.ts
+++ b/src/controllers/Upload/UploadController.test.ts
@@ -154,6 +154,55 @@ describe('Upload file — multer error handling', () => {
     jest.clearAllMocks();
   });
 
+  test('returns 400 with code=too_many_files when multer LIMIT_FILE_COUNT fires, without converting', async () => {
+    // multer has already deleted every temp file by the time it reports the
+    // count limit; running the conversion on the dead placeholders used to
+    // answer "your upload didn't finish" instead of naming the cap.
+    const multerError = new multer.MulterError('LIMIT_FILE_COUNT');
+    (getUploadHandler as jest.Mock).mockImplementation(
+      () =>
+        (
+          req: express.Request,
+          _res: express.Response,
+          cb: (err?: unknown) => void
+        ) => {
+          req.files = [{ path: '/gone/1' }, { path: '/gone/2' }] as never;
+          cb(multerError);
+        }
+    );
+
+    const uploadService = {
+      handleUpload: jest.fn(),
+    } as unknown as UploadService;
+    const notionService = new NotionService({} as INotionRepository);
+    const controller = new UploadController(uploadService, notionService);
+
+    const jsonSpy = jest.fn();
+    let capturedStatus = 0;
+    await new Promise<void>((resolve) => {
+      const fakeRes = {
+        locals: { patreon: false, subscriber: false },
+        status: (code: number) => {
+          capturedStatus = code;
+          return {
+            json: (body: unknown) => {
+              jsonSpy(body);
+              resolve();
+            },
+          };
+        },
+      } as unknown as express.Response;
+      controller.file({} as express.Request, fakeRes);
+    });
+
+    expect(capturedStatus).toBe(400);
+    expect(jsonSpy).toHaveBeenCalledWith({
+      code: 'too_many_files',
+      message: expect.stringContaining('21'),
+    });
+    expect(uploadService.handleUpload).not.toHaveBeenCalled();
+  });
+
   test('returns 413 with code=too_large when multer LIMIT_FILE_SIZE fires', async () => {
     const multerError = new multer.MulterError('LIMIT_FILE_SIZE');
     (getUploadHandler as jest.Mock).mockImplementation(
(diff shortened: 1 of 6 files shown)


aalemayhu (commented):
**Verdict**: approve

Two upload-path faults from one prod log line: over-count uploads now get a 400 that names the plan's file cap instead of "your upload didn't finish", and the receipt-time byte snapshot (#3414) runs again after its `req.destroyed` guard had silently disabled it since #4180 — simpler, honest errors and the reaper-race protection back. Round 1 engineering finding (blocking, correct): the cap is raised as `LIMIT_UNEXPECTED_FILE` on the upload field, not `LIMIT_FILE_COUNT`; fixed in the second commit with both codes covered by `test.each` and a stray-field case that still reaches the service. Nit applied: the controller test now uses a real `UploadService` with `handleUpload` spied. Security: fixed literal response, no leak, tier read from `res.locals` as elsewhere in the file. UX: message and changelog match VOICE. No blocking findings remain.

aalemayhu (commented):
**Verdict**: approve

Two upload-path faults from one prod log line: over-count uploads now get a 400 that names the plan's file cap instead of "your upload didn't finish", and the receipt-time byte snapshot (#3414) runs again after its `req.destroyed` guard had silently disabled it since #4180 — simpler, honest errors and the reaper-race protection back. Round 1 engineering finding (blocking, correct): the cap is raised as `LIMIT_UNEXPECTED_FILE` on the upload field, not `LIMIT_FILE_COUNT`; fixed in the second commit with both codes covered by `test.each` and a stray-field case that still reaches the service. Nit applied: the controller test now uses a real `UploadService` with `handleUpload` spied. Security: fixed literal response, no leak, tier read from `res.locals` as elsewhere in the file. UX: message and changelog match VOICE. No blocking findings remain.

aalemayhu (commented):
**Verdict**: approve

Two upload-path faults from one prod log line: over-count uploads now get a 400 that names the plan's file cap instead of "your upload didn't finish", and the receipt-time byte snapshot (#3414) runs again after its `req.destroyed` guard had silently disabled it since #4180 — simpler, honest errors and the reaper-race protection back. Round 1 engineering finding (blocking, correct): the cap is raised as `LIMIT_UNEXPECTED_FILE` on the upload field, not `LIMIT_FILE_COUNT`; fixed in the second commit with both codes covered by `test.each` and a stray-field case that still reaches the service. Nit applied: the controller test now uses a real `UploadService` with `handleUpload` spied. Security: fixed literal response, no leak, tier read from `res.locals` as elsewhere in the file. UX: message and changelog match VOICE. No blocking findings remain.

PR #4297, opened 2026-08-30
fix: answer 400 on /print when the upload is not a zip at all

aalemayhu:
## Why

Prod, 2026-08-27 15:30: `POST /api/apkg/pdf` answered `500` and the error log carried a bare yauzl stack — `End of central directory record signature not found. Either not a zip file, or file is truncated.` The user had picked a file that was an `.apkg` by name only. `sendPdfExportError` recognises two not-a-deck messages (`No Anki collection`, `open failed`) and treats everything else as a server fault, so a bad input became "PDF generation failed." plus a logged 500.

## What

- `sendPdfExportError`: the yauzl not-a-zip message joins the existing allowlist → `400 { message: 'Invalid .apkg file' }`, no `console.error`.
- Test: use case rejects with the yauzl message → 400 + no error log (watched failing first: 500).
- Changelog entry.

## Expected impact

One fewer false 500 on `/print`; the error-log sweep stops surfacing user input as a server fault. Read: `500 POST /api/apkg/pdf` count in the access log.

## Checks

- `pnpm test src/controllers/ApkgController.test.ts` — 28/28.
- Full server suite not re-run for this controller-only branch line; PR #4296's full run (702 suites green minus the documented pdfinfo case) covers the same base.

## Browser check

Browser check: not applicable — server-only change; the only `web/src` file is the changelog JSON.

### DIFF
--- a/src/controllers/ApkgController.test.ts
+++ b/src/controllers/ApkgController.test.ts
@@ -322,6 +322,31 @@ describe('ApkgController.exportPdf — pdf_print_options_used event', () => {
     return res as Response;
   }
 
+  it('answers 400 Invalid .apkg file when the upload is not a zip at all', async () => {
+    // yauzl's message for a renamed PDF/HTML; used to fall through to a logged
+    // 500 "PDF generation failed." (prod, 2026-08-27).
+    (ExportApkgToPdfUseCase as jest.Mock).mockImplementation(() => ({
+      execute: jest
+        .fn()
+        .mockRejectedValue(
+          new Error(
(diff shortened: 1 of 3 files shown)


aalemayhu (commented):
**Verdict**: approve

Turns a bad input on /print into a 400 instead of a logged 500 — an honest error for the user and a quieter error log. Security: fixed literal response, no leak (the fork noted it reduces leakage). Engineering: one nit — matching yauzl's message string can drift on a bump; the branch already does this for two other messages and `jobFailureReason` has a coded classifier worth reusing in a follow-up. UX: clean. No blocking findings.

PR #4298, opened 2026-08-30
refactor: raise InvalidApkgError from extractApkg instead of message text

aalemayhu:
## Why

Follow-up to https://github.com/2anki/server/pull/4297 (review nit). `ApkgController` decided "not an Anki package" by matching three message strings — yauzl's `End of central directory…`, our own `open failed`, and `No Anki collection file found in archive` — at two sites (`isInvalidApkgError` for CSV, inline in `sendPdfExportError`). A yauzl wording change on a bump would silently turn the `/print` 400 back into a logged 500 with no test going red.

## What

- `extractApkg.ts`: new `InvalidApkgError` beside `ApkgTooLargeError`; thrown for the yauzl open failure (message preserved for the server log), the no-zipfile case, and the missing-collection case. All three not-an-apkg messages originated here; nothing else in `src/` matched them.
- `ApkgController.ts`: `isInvalidApkgError` is `instanceof InvalidApkgError`; `sendPdfExportError` uses it (string matches removed).
- Tests: `extractApkg.test.ts` — not-a-zip → `InvalidApkgError` (with the reader's message kept), zip without a collection → `InvalidApkgError` (both watched failing first); controller test rejects with the class.
- No changelog: no user-visible change — the responses are the same `400 Invalid .apkg file`.

## Expected impact

None — internal. Drift-proofs the `/print` and CSV not-an-apkg 400s.

## Checks

- `pnpm test src/services/ApkgPreviewService src/controllers/ApkgController.test.ts src/usecases/apkg` — 19 suites / 184 tests.
- Full server suite: 2 failed, 25 skipped, 7 todo, 6817 passed, 6851 total — failing suites: FAIL src/lib/pdf/getPageCount.test.ts  (the documented pdfinfo environment case). Ran on the revision before the `cause` option was dropped and the formatter pass; the two touched files were re-run after (38/38).

## Browser check

Browser check: not applicable — server-only change, no `web/src` files.

### DIFF
--- a/src/controllers/ApkgController.test.ts
+++ b/src/controllers/ApkgController.test.ts
@@ -13,6 +13,7 @@ import ExportApkgToCsvUseCase, {
   EmptyDeckError as CsvEmptyDeckError,
 } from '../usecases/apkg/ExportApkgToCsvUseCase';
 import { track } from '../services/events/track';
+import { InvalidApkgError } from '../services/ApkgPreviewService/extractApkg';
 
 jest.mock('../usecases/apkg/ImportApkgToNotionUseCase');
 jest.mock('../usecases/apkg/ExportApkgToPdfUseCase');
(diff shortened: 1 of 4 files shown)


aalemayhu (commented):
**Verdict**: approve

Internal drift-proofing of the /print and CSV not-an-apkg 400s: one `InvalidApkgError` from `extractApkg` replaces three message-string matches in the controller. Security: both consumers send fixed literals, yauzl text stays server-side (the fork rated it a net leakage reduction). Engineering: every other `extractApkg` consumer and `jobFailureReason` checked — none inspected the old messages or error type; two low-confidence notes only (a pre-existing `require` in the test file, and yauzl's original error object no longer being the rejection value — no consumer used it). No blocking findings.

aalemayhu (commented):
**Verdict**: approve

Internal drift-proofing of the /print and CSV not-an-apkg 400s: one `InvalidApkgError` from `extractApkg` replaces three message-string matches in the controller. Security: both consumers send fixed literals, yauzl text stays server-side (the fork rated it a net leakage reduction). Engineering: every other `extractApkg` consumer and `jobFailureReason` checked — none inspected the old messages or error type; two low-confidence notes only (a pre-existing `require` in the test file, and yauzl's original error object no longer being the rejection value — no consumer used it). No blocking findings.

PR #4315, opened 2026-09-01
fix: match Stripe customer emails case-insensitively in claims

### LINKED ISSUE #4302: Claim subscription flow silently fails when the Stripe customer email has any uppercase letter
**Goal alignment**: The claim flow is how a paying user recovers access when they paid under a different address; a silent failure here converts a solvable case into churn and a support ticket, working directly against the retention lever in CLAUDE.md.

### Problem

When a user submits the "Paid with a different email?" claim form on /account, `SubscriptionClaimController` lowercases the submitted email, then `ClaimSubscriptionUseCase` looks the customer up in Stripe with `customers.list({ email })`, which matches **case-sensitively**. Any subscriber whose Stripe customer email is stored mixed-case returns zero customers: no claim token is created, no confirmation email is ever sent — while the UI shows the deliberate anti-enumeration success state ("If a subscription or pass exists for that email, we sent a confirmation link"). The user believes the request went through and waits for a mail that will never arrive.

Not a regression: the lowercase-input + case-sensitive-`list` combination shipped with the claim feature's birth commit (#2867). Confirmed 2026-09-01 at the wire level against one live subscriber's Stripe customer (entry point: `POST /api/subscriptions/claim`, two initiate audit rows, zero token rows). True affected count is unknowable from the DB because our own `subscriptions.email` rows are normalized lowercase; the pool is 724 active subs.

- "when I try to add it in the Account Settings, I dont receive any mail"
- "This is also the e-Mail adress I've been paying with"

### Proposed direction

Make the Stripe customer lookup case-insensitive — either match via Stripe's search API or fetch candidates and compare lowercased on our side. Keep the anti-enumeration success copy exactly as-is; the copy is correct, the lookup is what's broken. Add a use-case-level test (Jest, server layer) that a mixed-case Stripe customer email resolves for a lowercased submission. **Do not** build a Stripe email-normalization backfill or migration — normalize at read time, not by rewriting Stripe.

**Metric it should move**: claim-token creation rate per claim submission (a claim matching a live subscription should produce a token and a sent email), read off the claim audit rows; downstream, fewer "paid with a different email, no link arrived" tickets.

aalemayhu:
**Hard rail (subscriptions) — do not merge from a session; Alexander merges from the GitHub UI.**

## What
Makes the Stripe customer lookup in the subscription-claim flow case-insensitive. When `customers.list({ email })` (a case-sensitive filter) returns no customer for the lowercased submission, it now falls back to `customers.search({ query: "email:'<email>'" })`, which matches case-insensitively.

## Why
Fixes #4302. `SubscriptionClaimController` lowercases the submitted email, then the lookup runs through `SubscriptionService.findActiveStripeSubscriptions` → `stripe.customers.list({ email })`. Stripe's list email filter is case-sensitive, so a subscriber whose Stripe customer email is stored mixed-case matched zero customers: no claim token was created and no confirmation email was sent, while the UI still showed the anti-enumeration success state. The user waited for mail that never arrived — a solvable recovery case silently turned into churn and a support ticket.

## How
New `findStripeCustomersByEmail` helper in `SubscriptionService.ts`: exact-match list first, then a case-insensitive `customers.search` fallback only when the list is empty. The single-quote/backslash in the search query is escaped. `collectCandidateEmails`'s lowercase normalization, the anti-enumeration response copy, rate limits, and token TTL are all untouched. The pass-claim path (`ClaimPassUseCase`) matches by lowercased email hash and does not query Stripe by email, so it needs no change.

## Measuring success
Claim-token creation rate per claim submission — a claim matching a live subscription should now produce a token row (`subscription_claim_tokens`) and a sent confirmation email where before it produced only an `initiate` audit row with no token. Read off the claim audit rows; downstream, fewer "paid with a different email, no link arrived" support tickets.

## Testing
- Unit test added in `SubscriptionService.test.ts`: a Stripe stub whose customer email is mixed-case resolves the active subscription for a lowercased submission, and asserts the search fallback is only used when the exact-match list misses.
- Full server suite: 6744 passed; the only failures are the documented environmental ones (parser suites spawning `create_deck.py` without a venv → `PythonExitError`, and `getPageCount` needing `pdfinfo`).
- `pnpm format:check` clean tree-wide; `tsc -p . --noEmit` clean.

## Browser check

No UI change in this PR (server-side lookup only). The user-visible effect is that the confirmation email now arrives; verified via the unit test at the service boundary.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-01-claim-email-case.json` — "Claiming a subscription you paid for under a different email now works no matter how the email was capitalized"

## Risks
- Stripe `customers.search` is eventually consistent (up to ~1 min lag for newly created objects). This only affects brand-new customers; the fallback fires for existing mixed-case subscribers where consistency is a non-issue.
- The fallback adds one extra Stripe API call only on the list-miss path, so the common exact-match case is unchanged.
- Rollback: revert the single commit; behavior returns to list-only lookup.

## Goal alignment
Retention lever. A silent claim failure converts a recoverable "paid under a different email" case into churn plus a support ticket. Metric: claim-token creation rate per submission (claim audit rows), and fewer related support tickets.

Fixes #4302


aalemayhu (commented):
Review: clean pass — search fallback fires only on an empty exact-match list, the search query escapes quotes/backslashes, anti-enumeration copy and rate limits untouched, pass path verified unaffected (hash-based matching, no email list call), tests cover the mixed-case resolution and fallback-not-taken paths. Rail (subscriptions): merge from the GitHub UI.

Issue #4308, opened 2026-09-01
Testimonial pipeline: comp offer for unprompted praise


aalemayhu:
**Goal alignment**: acquisition lane — turns the inbox's recurring unprompted praise into public proof.

### Problem
The support corpus is full of unprompted love letters ("best customer support I ever met", "it is underrated") that convert to nothing. A one-off comp-for-testimonial exchange in 2024 worked and was never systematized.

### Proposed direction
A reply flow: when a user writes unprompted praise, offer 3 free months for a public review (site testimonial wall + review platforms). Template in the support toolkit; testimonial wall on the landing page. Consent explicit, identity only with permission.

### Metric
Testimonials collected/month; landing-page conversion with social proof section (A/B).

aalemayhu:
Shipped: #4316 landed the testimonial comp reply playbook (.claude/docs/testimonial-comp.md) — the reply flow and comp mechanics this issue proposed are documented and in use for support drafting, so closing.

PR #4316, opened 2026-09-01
docs: testimonial comp reply playbook

aalemayhu:
Part of #4308 — the reply-flow half. The issue stays open, re-scoped: the v2 testimonial wall is gated on >=5 consented quotes collected via this playbook (T+30 read).

- Comp mechanism: a 90-day user_passes row (kind='unlimited', comp:testimonial: sentinel intent id) — NOT ankify_access (wrong gate: that is the Auto-Sync comp, and it never expires) and NOT a Stripe coupon (praise-givers are mostly free users; a coupon demands a card on a gift). Self-expiring, zero revocation toil.
- No wall in v1 — it would render empty (the only landing social-proof element today is the product demo ShowcaseSection). Collection first; the count decides.
- No ops command/button — manual psql is proportionate below ~5 grants/month; the route+button+parity obligation waits for sustained volume.
- Not linked in CLAUDE.md's load-on-demand list, to keep this PR off the CLAUDE.md rail file; add the one-liner whenever you next touch CLAUDE.md, or say the word and I follow up.

No changelog entry — internal playbook, no user-visible behavior change.
Sonar: not run locally; PR decoration will report. Docs-only diff.
**Hard rail (.claude/ path) — do not merge from a session; merge from the GitHub UI.**

### DIFF
--- a/.claude/docs/testimonial-comp.md
+++ b/.claude/docs/testimonial-comp.md
@@ -9,8 +9,8 @@ before then.
 
 A 90-day `user_passes` row at the `unlimited` kind, granted by hand:
 
-    INSERT INTO user_passes (user_id, kind, expires_at, stripe_payment_intent_id, created_at)
-    VALUES (<id>, 'unlimited', now() + interval '90 days', 'comp:testimonial:' || gen_random_uuid(), now());
+    INSERT INTO user_passes (user_id, kind, expires_at, stripe_payment_intent_id)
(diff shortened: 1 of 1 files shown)


aalemayhu (commented):
Review: one blocking finding — the grant recipe named a created_at column user_passes does not have; fixed on the branch (column dropped from the INSERT). Everything else verified: findActive semantics match the doc, apple: sentinel precedent confirmed, reply template passes the support-email rules, no confidentiality leaks, thresholds internally consistent. Pass at this sha.

PR #4317, opened 2026-09-01
fix: carry the request id through Notion conversion logs

aalemayhu:
Finishes #4198. Most of the issue shipped earlier (#4219 request-correlation middleware, #4256 request id on Claude spend/chunk lines, upload path threaded end-to-end); this closes the last gap — the Notion conversion worker path.

- NotionController.convert now passes res.locals.requestId into the conversion worker request; conversionPool forwards it; performConversion stamps it on all five job lifecycle log lines.
- In-memory only: no DB column, no migration ("a helper, not a platform", per the issue).
- Not stamped: the Ankify 5-minute poll (no originating request exists there).

Tests: the three touched suites pass (NotionController 33/33, conversionPool 21/21, performConversion green); full suite run showed only load-induced timeouts (withStripeRetry, mountWebBuild) that pass in isolation, plus the documented environmental parser/pdfinfo failures. tsc --noEmit clean, format:check clean.

No changelog entry — internal observability, no user-visible behavior change.

Process note: the implementing agent drifted from its worktree into the main checkout mid-run; the diff was verified coherent and moved onto this branch by explicit-path staging. No other main-checkout state was disturbed.

### DIFF
--- a/src/controllers/NotionController.test.ts
+++ b/src/controllers/NotionController.test.ts
@@ -316,6 +316,17 @@ describe('NotionController', () => {
       );
     });
 
+    it('threads res.locals.requestId into the runConversion payload so worker logs join the request', async () => {
+      setupConvertMocks();
+      res.locals = { owner: 'owner1', requestId: 'req-abc-123' };
+
+      await controller.convert(req as express.Request, res as express.Response);
+
+      expect(runConversion).toHaveBeenCalledWith(
+        expect.objectContaining({ requestId: 'req-abc-123' })
+      );
+    });
+
     it('emits upload_started so the Notion path enters the upload→download funnel', async () => {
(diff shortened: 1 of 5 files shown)


aalemayhu (commented):
Review (medium, execution-verified): no blocking correctness regression — the requestId threading is mechanically correct end-to-end (single caller per layer, middleware mounted above the notion router, Piscina hop clones safely, both new tests non-vacuous, no-changelog out-clause correctly used). Merge-safe as-is. 14 non-blocking findings recorded (unstamped catch paths that limit the correlation goal, plus pre-existing defects in the touched functions incl. an unclassified worker-termination path and a log-template injection nit) — consolidated into a follow-up issue rather than a third round.

PR #4318, opened 2026-09-01
feat: self-serve email change on the account page

### LINKED ISSUE #4303: Add a self-serve email-change flow on /account (paying-user lockout risk)
**Goal alignment**: A paying subscriber who lost their registered inbox is one forgotten password from permanent lockout (password reset goes to the old address) — involuntary churn, the revenue leak CLAUDE.md says to close.

### Problem

There is no way to change an account's email anywhere in the product. /account renders the email read-only (`UserProfile.tsx`); the only email input on the page is the "Paid with a different email?" claim form, which changes nothing about the account address. A paying subscriber who lost access to their registered address went looking for an email-change control, found the claim form, and used it expecting it to move their account to the new address — it can't. Today the only path is manual SQL by the maintainer, and the user's fallbacks are both broken: password reset mails the inbox they've lost, and delete-plus-re-register would orphan their active subscription linkage. A paying user is stuck on an address they no longer control, with no self-serve exit.

- "I was trying to change my e-Mail adress from [old email] to [new email]."
- "I lost access to my old e-Mail adress, which is why I would like to change the account completly. Can you help me with this issue?"

### Proposed direction

Ship one focused email-change control on /account — a single primary action ("change email"), not a settings panel. The security shape is non-negotiable: require re-authentication before the change, send a confirmation link to the **new** address (the change only completes when that link is followed), and send a notification to the **old** address so a hijack is visible even when the old inbox is still monitored. Sequence the work: an ops email-change command first (cheap, replaces the manual SQL path, needs the matching CommandsTab button in the same PR), then the self-serve flow as the durable fix. When the self-serve flow lands, tighten the claim form's copy in the same PR so it plainly states it links a subscription and does not change your login email. **Do not** build an email-preferences center, multiple-emails-per-account, or aliasing — this is one address swap.

**Metric it should move**: monthly paid churn % (business-baseline block / Stripe); secondary, volume of email-change and lockout support tickets. Churn/retention surface touching paying users — route through the trio before build.

aalemayhu:
> **kanel regeneration pending from the main checkout.** This PR adds the
> `email_change_tokens` table (migration `20261012000000`) and types its columns
> locally in `src/data_layer/EmailChangeTokenRepository.ts`. `src/data_layer/public/`
> was **not** touched — run `pnpm kanel` against the applied migration from the main
> checkout and commit the regenerated `public/EmailChangeTokens.ts` before this goes ready.

## What
Self-serve account email change on `/account`. Under the read-only email, a
"Change email" control lets a signed-in user move their account to a new
address: they re-authenticate with their current password, we mail a
confirmation link to the new address and a heads-up to the old one, and the
change applies only when the link is opened. A pending change shows a badge on
reload with resend / cancel. The claim toggle is retitled "Link a subscription"
with a clause clarifying it does not change the sign-in email.

## Why
Closes #4303. A paying user who loses access to their old inbox has, today, no
path to change their account email short of manual SQL — one forgotten password
from lockout, and a recurring support thread.

## How
- `email_change_tokens` table (mirrors `subscription_claim_tokens`) + a new
  `EmailChangeTokenRepository` and in-memory twin.
- `RequestEmailChangeUseCase` (validate, re-auth via `comparePassword`, uniqueness
  check, per-user hourly rate cap, mint HMAC-hashed token, send both mails) and
  `ConfirmEmailChangeUseCase` (token-authed, applies the change).
- The confirm transaction lives in `UsersRepository.applyEmailChange`: it locks
  the user row, re-checks the token and email uniqueness inside the transaction,
  updates `users.email`, relinks subscriptions (`linked_email` where the sub
  matched by payer OR linked email — the pair that keeps `cancelUnlinkedSubscriptions`
  from sweeping a paying account), marks the token consumed, then revokes every
  session via `logOutEverywhere`.
- `UsersControllers` (RAIL) gains `requestEmailChange` / `confirmEmailChange` /
  `cancelEmailChange`; `UserRouter` (RAIL) mounts `POST /api/users/email-change/{request,confirm,cancel}`.
  Request + cancel are authed; **confirm is token-authed** (no session) so it
  works from the new inbox on any device.
- Two transactional email templates matching `reset.html` structure; web
  `ChangeEmail.tsx` under `UserProfile`, a token confirm landing page at
  `/account/email-change` (added to `App.tsx` and `knownRoutes.ts`), an
  `?email_changed=1` success banner, and `accountx`/`account`/`common` keys in
  all 10 locales.
- `email_change_requested` / `email_change_confirmed` added to both analytics
  allowlists (client-fired).

**Design decision — OAuth-only detection.** An OAuth-registered account stores a
bcrypt hash of a random UUID, indistinguishable from a real password by shape.
When `comparePassword` fails we check for an `oauth_identities` row: present →
`403 set_password_first` (routed to `/forgot`), absent → `401` wrong password.
An OAuth user who later set a real password compares true and proceeds; only a
genuine mismatch on an OAuth account diverts. Passwordless re-auth is out of
scope (v2), per the spec.

## Measuring success
`email_change_confirmed` volume at `/api/ops/metrics`; the primary signal is
"paid but locked out / change my email" support threads trending to zero.

## Testing
- Unit tests added:
  - `RequestEmailChangeUseCase.test.ts` — success mails both addresses, invalid /
    same-as-current / wrong-password / set-password-first / email-taken / rate-limited.
  - `ConfirmEmailChangeUseCase.test.ts` — unknown/expired/consumed token, confirm
    race (colliding account → clean `email_taken`, no session revocation), success
    applies the change and revokes sessions.
  - `UsersControllers.emailChange.test.ts` — HTTP status mapping (401/403/409/400/200)
    with only the DB edge mocked.
  - `UsersRepository.emailChange.sql.test.ts` — pg-dialect generated SQL for the
    subscription relink.
  - `InMemoryEmailChangeTokenRepository.test.ts`, `EmailChangeEmail.test.ts`,
    `ChangeEmail.test.tsx`.
- Full server Jest: 6828 passed; the 79 failures are the documented environmental
  ones only (Python-venv-absent `PythonExitError` parser/canary suites and
  `getPageCount` pdfinfo ENOENT). Full web vitest: 3476 passed. `tsc -p . --noEmit`
  and web `tsc --noEmit` clean. `pnpm format:check` clean tree-wide.

## Browser check
Notes: `pnpm --filter 2anki-web test:golden-path` — 4 passed at 375px, including
the subscriber account page which now renders the ChangeEmail control via UserProfile.

## Changelog
`2026-09-02-change-account-email.json` — "Change your account email from your
account page — confirm the switch with a link sent to the new inbox"

## Risks
- Auth + subscription-linking + a migration → hard rail; needs human review.
- The subscription relink runs `UPDATE subscriptions SET linked_email` for every
  row matching the old email; `linked_email` still carries a UNIQUE constraint, so
  an account with 2+ active subscription rows to the same new address would hit a
  23505 that surfaces as `email_taken` (matches the existing
  `changeEmailAndRelinkSubscriptions` risk envelope; single-sub is the norm).
- Rollback: revert the branch; the migration `down` drops the table.

## Security notes
- Token is `crypto.randomUUID`, stored only as an HMAC (`hmacToken`) — never logged;
  no email/token is logged on these paths.
- Re-auth on request; confirm re-checks uniqueness inside the transaction and relies
  on the `users.email` UNIQUE constraint as the final race guard (accounts never merge).
- Notice to the OLD address at request time gives a hijack victim a window to react;
  all sessions are revoked on completion.
- Confirm is token-authed by design (link goes to the new inbox); request/cancel are
  behind `RequireAuthentication`.

## Goal alignment
Retention/per-user value: removes a hard lockout path for paying users (the audience
most costly to lose) and deletes a recurring support burden. Read on
`email_change_confirmed` at `/api/ops/metrics` and the lockout support-thread count.


aalemayhu (commented):
Review pass: spec implemented per the locked decisions with three flagged deviations (cancel endpoint for flow state D, hourly rate cap, OAuth detection via identity row); security-review findings all closed on-branch (rate-cap bypass via cancel, stale tokens after confirm, taken-email enumeration) with regression tests; kanel regeneration committed from the main checkout; email sends wired into the #4329 category system; Sonar clean at this head (S8786 regex replaced with the shared normalizeEmail, S2933 readonly). Checks green. Hard rail (auth surface, migration) — waiting for Alexander to merge.

PR #4329, opened 2026-09-01
feat: tag every outgoing email with a template category

### LINKED ISSUE #4313: Transactional-email delivery monitoring per template
**Goal alignment**: protects signup (acquisition) and billing/auth flows that silently fail today.

### Problem
Across 26 months, undelivered transactional email recurs in every era: password reset, magic login, cancel confirmation, Stripe portal login. Worst case: through part of 2026 the signup-confirmation flow silently did nothing for new accounts while the page claimed success — an acquisition leak found by one user's report. A recurring "email gateway outage" is mentioned in replies but nothing monitors per-template delivery.

### Proposed direction
Delivery-rate monitoring per template off the existing SendGrid event webhook data (sendgrid.events are already processed): alert when a template's delivered rate drops or send volume flatlines abnormally. Ops surface — backend + CommandsTab/ops visibility per the ops-command parity rule.

### Metric
Detection time for a delivery failure (target: hours, not user-reported); "email never arrived" threads per month.

aalemayhu:
## What
Gives every outgoing email a stable, greppable category slug and captures the SendGrid event category so delivery/bounce events can be attributed back to the template that sent them. Step 1 of 2 toward per-template delivery monitoring.

## Why
Part of #4313. The SendGrid event pipeline already stores delivered/bounce/dropped events (`suppression_events`), but nothing attributes an event to the template that sent it — `EmailService` set no categories — so per-template delivery monitoring is impossible today. Tagging sends and capturing the event category lights up SendGrid's own per-category dashboard stats within 24h, which is the evidence needed before building the monitor itself.

## How
- `EmailService.deliver(msg)` → `deliver(msg, category)` sets `msg.categories = [category]`; every send method passes its slug. Direct `sgMail.send` / `sendWithRetry` call sites set `categories` in the message literal. Slugs live in a `EMAIL_CATEGORIES` const map next to the template constants so the set is greppable.
- `ProcessSendgridEventsUseCase` reads `event.category` (SendGrid sends `string | string[]`), normalizes to the first non-empty string or `null`, and buckets each recorded event under its category (null → `uncategorized`) in a new `categories: Record<string, number>` field on the result. The webhook router already logs that result object verbatim as `sendgrid.events.processed`, so per-category counts surface immediately with no change to the rail router or signature files.

Chose the log-summary path (option B) over threading a value the repository would drop: it is smaller, honest, and makes attribution visible in prod logs on day one. `sum(categories) === recorded` holds as an invariant.

## Measuring success
Prod log line `sendgrid.events.processed` now carries a `categories` object, e.g. `{ "magic-link": 42, "password-reset": 11, "uncategorized": 130 }`. At T+24h the SendGrid dashboard's per-category stats should show the tagged categories with non-zero delivered counts, and the `uncategorized` share in the log should fall as tagged sends flow through. Grep prod logs: `sendgrid.events.processed`.

## Testing
- Unit tests added:
  - EmailService: representative send methods tag `categories` correctly across all three send mechanisms — deliver path (magic-link, password-reset), direct `sgMail.send` (pass-winback and 15 more via `it.each` over every FS-free send method), and `sendWithRetry` (contact-form); plus a reflection test asserting every `EMAIL_CATEGORIES` value is a unique non-empty kebab-case slug, and a dedicated test for the FS-touching subscription-cancelled sender.
  - ProcessSendgridEventsUseCase: `category: ['x']` array form, `'x'` string form, missing/empty → `uncategorized`, aggregation across events, and the `sum(categories) === recorded` invariant. Existing result-shape assertions updated for the new `categories` field.
- Full server suite: 693 suites / 6767 tests pass. The only failures are the documented environmental set — Python apkg suites (parser/DeckParser/BlockHandler/upload/canary) fail because this worktree has no `create_deck/.venv`, and `getPageCount` fails on missing `pdfinfo`; `ClaudeService` passes 143/143 in isolation (its full-run failure was a load timeout). None touch the changed files.
- `pnpm format:check`: clean tree-wide.

## Browser check
Not applicable — server-side email tagging, no rendered output.

## Changelog
No entry — internal observability, no user-visible behavior change.

## Risks
Low. No recipient/subject/body change, no schema change, no rail files touched. Riskiest assumption: SendGrid only emits `category` on events for sends made **after** this deploys, so the acquisition-critical signup path must route through a tagged sender — verified in code (magic-link and password-reset both tagged), worth a dashboard glance at T+24h. Rollback is a straight revert.

- Step 1 of 2: tagging + capture only. The per-template delivery monitor needs a `suppression_events` category column (migration → rail) and a scheduled check wired in `server.ts` (rail) — spec'd in the consult, gated on 24h of SendGrid dashboard category stats confirming attribution works end-to-end before the rail PR is worth your review.
- The webhook receiver and signature files are deliberately untouched (rail); the category is read in the use case, and the existing `sendgrid.events.processed` log line carries the per-category counts without a router change.
- Riskiest assumption flagged: SendGrid only emits category on events for sends made AFTER this deploys; the acquisition-critical signup path must route through a tagged sender — verified in code, worth a dashboard glance at T+24h.

## Goal alignment
Protects signup (acquisition) and billing/auth flows that silently fail today: makes transactional-email delivery observable per template, so a delivery drop (the recurring "email never arrived" failure mode behind lost signups and password resets) becomes detectable in hours from a log/dashboard rather than by user report. Metric: detection time for a per-template delivery failure — read from the `sendgrid.events.processed` log and the SendGrid per-category dashboard, first check at T+24h.

Part of #4313

Fixes #4313


aalemayhu (commented):
All 5 review findings fixed and verified: partial-batch counts now surface via SendgridEventProcessingError (webhook logs partial counts and 500s so SendGrid retries), first-non-empty category scan, per-(category,eventType) nested counts with tests updated, magic-link-login vs magic-link-password-reset split by purpose, and the lastCategories empty-guard with a regression test. Full server suite 696/6804 passing (9 environmental failures, none email), web 380/3423, deploy tsc clean, format clean. Hard-rail paths (SendgridWebhookRouter) — waiting for Alexander to merge.

PR #4332, opened 2026-09-01
feat: flag thin decks inline and backfill export-drift fixtures

### LINKED ISSUE #4310: Conversion receipt: found vs created counts, plus a Notion export canary
**Goal alignment**: converts the corpus's scariest failure class — silent partial decks — into visible reliability.

### Problem
Across 26 months: "250 toggles but only 21 cards", "only taking 22 toggles… it is really random", syncs reporting "0 created" with no error. Four separate Notion-export breakage waves (2024 Markdown confusion, Dec 2025 nested toggles ~13 reporters re-opening over 3 months, Apr 2026 images ~14, Jul 2026 export change) were each detected by users, not by the system. Silent partial output is the top churn driver among loyal long-time users.

### Proposed direction
Two parts, same theme: (1) a conversion receipt on every job — sources found vs cards created vs skipped (with reasons), shown to the user when counts diverge; (2) contract tests against Notion export fixtures feeding the existing parser canary so the next export change alerts before users report it. Trio before build for the receipt UI.

### Metric
Days-to-detect an export change (target: <1); "missing cards" threads per month.

aalemayhu:
## What
Two things behind the already-shipped conversion receipt. (1) An inline **thin-deck notice** on Downloads: when a Notion page yields a deck thinner than the page, the row now says *why* (empty toggles, not-connected parts, or a mixed set) and *how to fix it*, instead of leaving "only N cards" unexplained. (2) Two new **export-drift canary fixtures** for user-detected Notion export waves the existing three did not cover.

## Why
Closes #4310. The receipt half of the issue already shipped (#4211/#4229: `blocks_seen`/`cards_created`/`blocks_skipped` + `ConversionReportModal` + the skipped-count pill). This PR ships the two real remaining gaps: a proactive inline explanation for thin decks, and canary coverage for historical export breakages so the next one is caught by a fixture rather than a user.

## How
- **Notice:** `NotionEmptyToggleNotice` → `ThinDeckNotice` with `reason: 'emptyToggles' | 'notConnected' | 'generic'`. New `getThinDeckSignal(job)` returns `{ candidateSkips, reason }` where `candidateSkips` = empty-back count + forbidden/not-connected count + unsupported-block count, **excluding dropped images** (an image missing from a converted card is not a lost card). The dominant skip family picks the reason. Trigger: `cards >= 1 && candidateSkips >= cards && candidateSkips >= 2`.
- **Pill/notice agreement:** the row pill now reads the same `candidateSkips` number, so the pill and the notice never disagree. (Behavior change: an image-only-drop row no longer shows a "skipped" pill; the dropped-image detail still shows in the conversion report modal.)
- **Fixtures:** added `notion-figure-image-2026.html` (+ a fabricated 1x1 PNG) for the Apr-2026 `figure.image` subfolder-path shape, and `notion-katex-equation-2026.html` for the 2026 KaTeX `data-notion-equation` shape. Both registered in `EXPORT_DRIFT_FIXTURES`. Baselines were **measured** against the current parser via the existing `measureExportDriftFixture` tooling (image baseline `mediaCount: 1` proves the bundled image still embeds; equation baseline proves the card survives with a non-empty back), never hand-written. Fixtures are fabricated/anonymized (no real deck titles, filenames, or workspace names).
- **Analytics:** `empty_back_notice_shown` → `thin_deck_notice_shown` `{ reason, cards, skipped }` for the Notion notice, added to both `web/src/lib/analytics/events.ts` and `src/types/AnalyticsEvents.ts`.

## Measuring success
`thin_deck_notice_shown { reason, cards, skipped }` at `/api/ops/metrics` — day-7 read: how often the notice fires and which reason dominates (emptyToggles vs notConnected tells us whether the friction is user-page structure or Notion Connections). The export-drift canary emails support on any baseline divergence (weekly `runExportDriftCanary`).

## Testing
- Unit tests added: `getThinDeckSignal.test.ts` (candidate-skip math incl. image exclusion + dominant-reason branching), `ThinDeckNotice.test.tsx` (trigger math incl. floor, reason copy per branch, plural rendering, `aria-haspopup="dialog"`, event props), plus a DownloadsPage pill/notice **agreement** test and updated existing thin-deck tests to the new copy. `runExportDriftCanary.test.ts` gains cases asserting the two new baselines don't diverge.
- Full web suite (isolated, no parallel load): **381 files / 3432 tests, all pass.**
- Full server suite: 691 suites pass; 12 suites fail as the **documented environmental class** (parser/`create_deck.py`/`pdfinfo`/zip suites 5s-timeout under a worktree without a native venv + parallel load) — none import the changed files, and `runExportDriftCanary.test.ts` passes 7/7 in isolation (measured with the main-checkout venv via `ANKI_PYTHON`, parse-only, no venv in the worktree).
- `pnpm format:check` clean tree-wide; web typecheck, web lint, server `tsc -p .`, `depcruise` all clean.

## Browser check
Notes: `pnpm --filter 2anki-web test:golden-path` → 4/4 passed (20.8s), incl. "signed-in dashboard renders without console errors at 375px" and "landing renders without console errors at 375px".

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-02-thin-deck-notice.json` — "When a deck comes out thinner than the page, Downloads now says why and how to fix it"

## Sonar

## Risks
Low. The notice is a generalization of an already-shipped pattern; the pill number source is the only behavior change (image-only-drop rows lose the pill badge but keep the report link/modal). Rollback is a straight revert. New locale strings fall back to English if any key is missing.

## Goal alignment
Activation / first-conversion clarity: a user who gets a thinner-than-expected deck and no explanation is an abandonment risk. The notice names the cause and the fix ("add an answer inside each toggle" / "add 2anki under Connections" / "open the report"), which should drive re-conversions. Metric: `thin_deck_notice_shown` volume by reason at `/api/ops/metrics`, read day-7.

- The issue's receipt half was already shipped by #4211/#4229 (blocks_seen/cards_created/blocks_skipped + modal + divergence badge) — this PR ships the two real gaps: canary fixtures for the historical export waves, and an inline thin-deck notice generalized from the empty-toggle notice.
- Notice trigger: cards≥1 AND candidateSkips≥cards AND candidateSkips≥2, with dropped images excluded from the count — ratio+floor chosen over "any divergence" to avoid banner-blindness; the floor is the tunable lever.
- Rejected: a blocks_unaccounted heuristic (cloze/table pages legitimately produce more cards than blocks — false "we lost your cards" alarms), and an upload-path receipt (a new surface; gated on its own evidence).
- Event renamed empty_back_notice_shown → thin_deck_notice_shown {reason, cards, skipped} in both allowlists. **Deviation from the brief, flagged:** `empty_back_notice_shown` was NOT removed from the allowlists because a second, live, out-of-scope consumer still fires it — `EmptyBackNotice` on the **upload path** (`UploadForm.tsx`). Removing it would silently record that event as `unknown` (the exact failure the parity rule guards against, in the other direction). So both events now exist; only the Notion notice migrated.
- Fixture list limited to waves reconstructable from repo evidence. **Skipped: the Jul-2026 block-key vs text-key wave** — it is a GUID-identity/dedup concern (#4255), not a card-shape drift: none of the canary's measured fields (deck/card/front/back/media counts) change with GUID keying, the July-2026 details.toggle *shape* is already covered by the existing `notion-details-toggle-2026` fixture, and the identity behavior is covered by the block-id ledger tests. No distinct shape to add without inventing one.
- "Connections" is kept verbatim in all 10 locales — it should match Notion's own localized menu label; flagged for native review.


aalemayhu (commented):
Review: clean pass — image-exclusion in the skip count verified against source, pill/notice number agreement tested, trigger ratio+floor as specced, event rename in both allowlists with the legacy event kept for its live upload consumer, 10-locale keys incl. ru/pl plurals, fixtures fabricated with measured baselines, amber-only styling and correct ARIA. One fix on the branch: changelog dated to the actual ship day.

aalemayhu (commented):
Review verdict carried over: the only change since the reviewed commit 754217e06e26 is a merge of main (branch update) — the PR diff is unchanged, checks are green on the new head.

Issue #4307, opened 2026-09-01
Upload-time guardrails: Markdown exports, oversized files, PDFs in zips


aalemayhu:
**Goal alignment**: removes the most common self-inflicted conversion failures instead of explaining them in support email.

### Problem
26 months of support mail shows the same avoidable failures: Notion Markdown/CSV exports produce garbage decks (top self-diagnosis failure of 2024, ~17 users; the "export as HTML" explainer was sent by hand ~10+ times), oversized zips (>30MB) crash with heap OOM (a student wrote a detailed spec unprompted), and zips containing PDFs slow or confuse conversion.

### Proposed direction
Upload-time guardrails: detect a Markdown export and redirect with exact fix steps; warn and offer splitting above a size threshold; ask "convert or skip PDFs?" when a zip contains them. Each guardrail is copy + detection, not new conversion capability. Trio before build.

### Metric
Failed-conversion rate; "wrong export format" and OOM support threads → 0.

PR #4334, opened 2026-09-01
fix: harden Notion Markdown upload guardrail (races, a11y, i18n)

aalemayhu:
Follow-up hardening for the Notion Markdown upload guardrail (#4328, issue #4307). The five review findings from the pre-merge review round landed after the PR merged, so they ship here:

1. **Duplicate-conversion guard** — `validate` is async; a validation resolving after a conversion already started fired a second `/api/upload/file` POST. `submitFiles` now checks a `zoneStateRef` mirror and refuses while converting (reproduction test asserts 1 upload, was 2).
2. **Impression gate** — `upload_guardrail_shown` fired from the hook even when the panel never rendered (locked wall, mid-conversion). The event now fires from a visibility-gated effect in the component.
3. **i18n** — the hard-coded "Pick a different file" button is now `t('upload.validation.pickDifferent')`, key added to all 10 locales (fr in tu-register). The identical pre-existing string from #2871 in a different render path was left untouched.
4. **Live-region precedence** — the polite announcer spoke the markdown warning while the upgrade wall was visible; it now shares the same visibility condition as the rendered panel.
5. **Test hygiene** — the guardrail describe leaked a stubbed global `fetch` (`restoreAllMocks` doesn't unstub globals); added `unstubAllGlobals` and tightened a weak `toBeGreaterThan(0)` to `toBe(1)`.

Verification: full web suite 3436/3436 green, tree-wide `pnpm format:check` clean, web `tsc --noEmit` clean, golden-path spec green (evidence for the attestation below). Server suites untouched (diff is 100% web).

## Browser check
Notes: golden-path spec (`pnpm --filter 2anki-web test:golden-path`) run green on this branch — 4 passed.


aalemayhu (commented):
Reviewed the full diff: the zoneStateRef guard correctly blocks a late-resolving validation from double-submitting, the impression event is now gated on the same visibility condition as the rendered panel (deduped per validation instance), the live region shares that condition, the i18n key ships in all 10 locales, and the test hygiene fixes are sound. Full web suite 3436/3436, format and typecheck clean, golden-path spec green.

aalemayhu (commented):
Re-review after clearing the six Sonar test smells (toHaveLength, redundant act wrappers) — test-only commit, 62/62 UploadForm tests green, format clean. Verdict unchanged from the prior pass.

PR #4335, opened 2026-09-01
fix: salvage cards from truncated Claude chunks instead of dropping them

### LINKED ISSUE #4235: Claude responses still truncate at max_tokens on chunk 0 after #4152
## What

Prod error log, last 48h:

```
[Claude] Response truncated at max_tokens { chunkIndex: 0, totalChunks: 1, maxTokens: 32768, chunkBytes: 39785 }
[Claude] Response truncated at max_tokens { chunkIndex: 0, totalChunks: 2, maxTokens: 32768, chunkBytes: 39364 }
[Claude] Response truncated at max_tokens { chunkIndex: 0, totalChunks: 2, maxTokens: 32768, chunkBytes: 29081 }
[Claude] Dropped malformed entries from parsed response { chunkIndex: 52, droppedDecks: 0, droppedCards: 26 }
```

#4152 ("chunk giant documents at half size to stay under the cap", 2026-08-21) was live for all of these. Chunks of 29–40 KB still push the response past the 32768 `max_tokens` ceiling, always on `chunkIndex: 0`.

## Why it matters

A truncated response = a partial deck for that chunk (and a re-bill if any retry path re-sends it). The 26 dropped cards on a single chunk is the same class: the tail of a long response is unparseable. Users get fewer cards than their document holds with no notice that says why.

## Suggested fix

- Size chunks by expected *output* tokens, not input bytes: card-dense inputs (glossaries, lists, tables) produce far more output per input byte than prose. A 30 KB glossary can exceed 32k output tokens.
- Or: on truncation, split that chunk in half and re-run the halves (bounded, once) instead of accepting the partial.
- Surface it: when a chunk truncates, add a conversion-report note ("some cards in section N were cut short") so the user can see why the deck is thinner than expected. #4227–#4229 give the report a home.

Found during a 48h prod-log review on 2026-08-25.

aalemayhu:
## What

When a conversion chunk's Claude response comes back with `stop_reason === 'max_tokens'`, the truncated body — which already holds many complete cards before the cut point — was discarded in full and the chunk re-run via the halving retry. If both halves also truncated, the whole chunk was dropped and the user silently lost every card the first response had produced. This PR salvages the complete cards from the truncated response (best-effort parse, `jsonrepair` drops the incomplete tail card) and uses them: a truncated half now contributes its salvaged cards instead of being discarded, and the original response's salvage is the fallback when the halves recover nothing.

## Why

Closes #4235. Prod logs showed `[Claude] Response truncated at max_tokens` reliably on `chunkIndex: 0` for 30–40 KB card-dense docs, even after #4152. Two root causes:

1. **Why always chunk 0:** `chunkHtmlByDetails` packs chunk 0 up to the last `</details>` before `offset + CHUNK_SIZE` (~40 KB), while trailing chunks get only the remainder. Chunk 0 is systematically the fullest, so on card-dense input it is the one that overruns the 32768 output cap.
2. **Why #4152 didn't cover it:** #4152 only drops chunk size to 20 KB for documents past `GIANT_INPUT_THRESHOLD` (300k chars). The reported docs are 1–2 chunks (< 80 KB), so they never crossed the threshold and chunked at the full 40 KB.

The user-visible outcome: a card-dense page that previously came back thin (dropped chunk, or only the clean half) now keeps the cards Claude actually produced.

## How

- `ClaudeTruncatedError` carries `salvagedDecks: DeckInfo[]` (defaults to `[]`, backward-compatible).
- On `max_tokens`, `generateDeckInfoFromChunk` best-effort parses the partial body through the shared `parseDeckResponse` + `expandCompactDeckInfo` (`salvageTruncatedDecks`, wrapped so an unrecoverable body yields `[]`) and attaches the result before throwing.
- `runChunkWithTruncationRetry`: a half that truncates again contributes its own `salvagedDecks`; when the halves recover nothing, `salvageOrThrow` falls back to the original response's salvage; the chunk only truly drops when nothing is salvageable anywhere.
- **Cost:** no `CHUNK_MAX_TOKENS`, chunk size, or retry cap change. Salvage reuses tokens already billed on the truncated call (previously thrown away). Per the paid-path cost rule, this reshapes control flow only.

### Not done here (needs a budget decision — Alexander's call)

The salvage fix stops the *card loss* but does not reduce truncation *frequency*. The issue's primary suggestion — size chunks by expected **output** tokens rather than input bytes, so card-dense sub-`GIANT_INPUT_THRESHOLD` docs start smaller — would cut the halving re-bills, but a smaller chunk size raises the call count and per-doc input-token cost for the common case. That is a spend trade-off, so it is deliberately out of scope; recommend a follow-up with a day-1/day-2 spend read if the salvage log below shows halving still firing often.

## Measuring success

- `[Claude] Response truncated at max_tokens` now carries a `salvagedCards` count — nonzero means cards were recovered from a call that previously contributed zero on that path.
- `[Claude] Keeping cards salvaged from the truncated response` fires when a chunk that would have been dropped is kept — its absence over a week means the both-halves-truncate case is rare; its presence quantifies cards saved.
- Grep both against `chunkIndex: 0` to confirm the reported pattern is now recovered rather than dropped.

## Testing

- Unit tests added (server Jest, Anthropic SDK stubbed at the module boundary):
  - `max_tokens` with a salvageable partial + both halves truncate → salvaged cards kept, chunk not dropped (previously the conversion rejected).
  - a retried half truncates with a salvageable partial → its cards survive next to the clean half (previously only the clean half survived).
- `ClaudeService.test.ts`: 145/145 pass (existing 3 truncation tests unchanged — their partial bodies are unsalvageable, so behavior is preserved).
- Full server suite: 6807 passed; the 81 failures are all in the 9 documented environmental suites (Python `create_deck` venv for apkg export + `getPageCount` pdfinfo), none under `src/lib/claude/`.
- `tsc -p . --noEmit` clean; `pnpm format:check` clean tree-wide; web vitest 3505 passed; web typecheck clean.

## Browser check

- [x] Not applicable — server-side conversion-pipeline change; the only `web/src` file is a changelog JSON. Effect (a fuller deck) is only observable through a live truncating conversion, which needs prod-scale card-dense input.
Notes: Sonar: not run locally; PR decoration will report.

## Changelog

`web/src/pages/WhatsNewPage/changelog/2026-09-01-truncated-chunk-salvage.json` — "Card-dense pages keep more cards when a long section would otherwise be cut short"

## Risks

- Low. The salvage parse is wrapped so any failure returns `[]` (no new throw paths). The original-response salvage is only used as a last-resort fallback, so it cannot duplicate cards against successful halves; `mergeDeckInfoArrays`/`dedupeIdenticalCards` collapse any half-level overlap downstream as before.
- Rollback: revert the single commit; no schema, config, or budget change to unwind.

## Goal alignment

Simpler/faster/more beautiful: a card-dense page returns the deck the document actually holds instead of a silently thinner one — fewer "why is my deck short" support reports. Funnel: conversion-quality/first-conversion satisfaction; not a directly-instrumented funnel metric, but it removes silent card loss on the paid AI path (read via the two salvage log lines above). No revenue-metric target.


aalemayhu (commented):
Reviewed the full diff: truncated responses now salvage complete cards via the existing parse path (try/catch to empty), the halving retry keeps per-half salvage without double-counting (original salvage only when halves recover nothing), non-truncated flow is unchanged, and no token budget, chunk size, or retry cap moved. Two new failing-first tests cover salvage and fallback; full server suite green outside the documented environmental set.

PR #4336, opened 2026-09-01
feat: cache Claude conversion results by content, settings, model

### LINKED ISSUE #4213: Design: conversion result cache keyed on content + settings + model
Repeat uploads of the same document are real (the Aug 2026 spend incident had one user re-upload the same file five times) and each re-run re-bills the AI path. A conversion cache could return the finished deck instantly and cut spend — but ONLY if keyed correctly: hash(file bytes) alone is wrong, because output depends on card options, note type, user instructions, tier limits, and the AI model/prompt version. Cache on file alone and users get stale decks after any settings change or model swap.

Design questions before any code: full cache key vector (content hash + settings snapshot + model id + prompt version), invalidation on model/prompt bumps (tie to the cost-envelope checklist), storage/TTL given the 2h workspace reaper, and whether #4115's duplicate-upload rejection already captures most of the win. Parked until someone wants to own the design — do not implement casually.

Inspiration: BPF's content-addressed program identity (bpf_prog_calc_tag). Note: stable note GUIDs for re-import-updates already exist (the guid ledger in UploadService) — that half of the original idea is shipped.

aalemayhu:
> **DRAFT — migration needs `pnpm kanel` run from the main checkout before ready.** This worktree has no local Postgres, so `src/data_layer/public/ConversionResultCache.ts` was not generated. The new table's rows are typed locally inside `ConversionResultCacheRepository.ts` (never hand-editing `public/`). Before flipping ready: apply the migration on the main checkout, run `pnpm dlx kanel -c ./.kanelrc.js`, and commit the generated `public/*`. **Requesting migration-reviewer.**

Closes #4213

## What
Adds a content-addressed Postgres cache for the Claude conversion **inference result** (`DeckInfo[]` — the expensive part), so re-uploading an identical document with identical settings returns the cached cards instead of re-billing the AI path.

## Why
Repeat uploads of the same document are real — the Aug 2026 spend incident had one user re-upload the same file five times, and each re-run re-billed the whole Claude path. #4213 asks for a cache that is safe **only if keyed correctly**: hashing the file bytes alone serves stale decks after any settings or model change. User-visible outcome: identical re-uploads finish instantly and stop re-billing AI spend.

## How
- **Key (`buildConversionCacheKey`)** — SHA-256 over normalized content (CRLF→LF, trimmed) + **every output-affecting input**: media-file list, `userInstructions`, `cardStyle`, `cardSize`, `fieldMapping`, `comprehensive`, `isPaying` (floor-v1 gates on both, so a free vs paid run genuinely differs), the model id, and a digest of `SYSTEM_PROMPT`. Miss any one and the cache would serve wrong output, so a model swap or a prompt edit changes the key and retires prior entries automatically — no manual version bump, dovetailing with the model-swap cost-envelope checklist.
- **Single model source of truth** — `CONVERSION_MODEL` is used both in the API call and the key, so the string sent to Claude and the string in the key can never drift.
- **Wrapper (`withConversionResultCache`)** — a small generic, dependency-free combinator: on miss it computes and saves; on hit it returns the cached value; on **any** cache-layer error (read or write) it logs and falls through to a fresh conversion. A broken cache never fails a conversion.
- **Storage** — `conversion_result_cache` table via a Knex migration; adapter `ConversionResultCacheRepository` (30-day read-time TTL as storage hygiene, `deleteOlderThan` provided for a future reaper). Wired into `PrepareDeck` and the `UploadService` restart path via `getConversionResultCache()`, which returns `undefined` (cache skipped) when `DATABASE_URL` is unset — behavior then bit-for-bit unchanged.
- **Vision path excluded** — the `pdfImageFallback` path reads page-image bytes off disk that the key does not capture, so it is never served from cache.
- No `max_tokens`, chunk-size, or retry-cap change. Caching is on by default; **no new env flag**.

## Measuring success
- Cache hits: grep prod logs for `[Claude] conversion cache hit` (carries the key prefix + model, never content).
- Spend: the existing `[claude-usage]` line / `ai_usage_recorded` event volume for `surface=conversion` should fall for repeat inputs; a hit emits **no** `[Claude] Sending request to Claude API` line for that input.
- Table health: `select count(*), sum(hits) from conversion_result_cache` shows entries and cumulative hits saved.

## Testing
- Unit — `conversionResultCache.test.ts` (19): key is 64-hex + deterministic; key changes when content / model / prompt-version / userInstructions / cardStyle / cardSize / comprehensive / isPaying / mediaFiles / fieldMapping change; CRLF-vs-LF and surrounding-whitespace normalization collapse to one key; wrapper miss (computes + saves), hit (returns cached, never computes), no-store passthrough, read-throws fallthrough, write-throws still returns.
- Integration — `ClaudeService.test.ts` (+4): `generateDeckInfo` calls Claude and stores on a miss, returns cached decks and never calls the SDK on a hit, bypasses the cache for the PDF-image fallback path, converts fresh when the cache read throws.
- SQL — `ConversionResultCacheRepository.sql.test.ts` (2): generated get SQL (ttl cutoff on `created_at`) and upsert-on-`cache_key` SQL asserted against a real pg-dialect knex.
- Full server suite: **6830 passed**, 25 skipped, 7 todo; 81 failed in 9 suites, all the documented worktree-environmental ones (`create_deck` Python venv `PythonExitError` in `DeckParser.*` / canary / golden suites, and `getPageCount` pdfinfo). Zero failures reference this change. `tsc -p . --noEmit` clean; `pnpm format:check` clean tree-wide; `pnpm arch` 0 errors.

## Browser check
- [ ] Not applicable — backend-only, no `web/src` change and no user-facing UI. The behavior (a cache hit) is observable only in server logs and finished-deck latency.

## Changelog
No entry — internal cost/perf optimization with no user-visible surface change (a cached re-upload just finishes faster). Per repo policy, backend-only.

## Risks
- **Wrong key → stale/incorrect decks.** Mitigated by folding *every* output-affecting input into the key, including `isPaying`/`comprehensive` (floor-v1) and a `SYSTEM_PROMPT` digest + model id (auto-invalidation on any prompt/model change). If a NEW output-affecting knob is added to `generateDeckInfo` later, it MUST be added to `ConversionCacheKeyInput` in the same change or the cache will serve pre-knob output.
- **Table growth.** Bounded by distinct-input cardinality; a 30-day read-time TTL keeps old entries from being served. No physical reaper is wired (out of scope; `deleteOlderThan` is ready for one). Rollback: drop the table (the `down` migration) — only cached inference is lost, recomputed on next conversion.
- **Migration not yet kanel-regenerated** (see the DRAFT note). Repository is self-typed so nothing imports a missing `public/` type; ready is blocked on the kanel run only.

## Goal alignment
Revenue/cost lever: cuts per-user Claude spend on repeat conversions (retention/per-user-value focus), directly serving the "never let user-count work crowd out revenue work" mandate. Read where: `[claude-usage]` conversion-surface spend and the `[Claude] conversion cache hit` rate. Also nudges the mission ("fastest way to a deck") — a cached re-upload returns instantly. No funnel metric; this is a cost/latency change, not an acquisition surface.

1. **Cache the inference result (`DeckInfo[]`), not the `.apkg` bytes.** *(alternative rejected: cache final apkg.)* The issue says cache the expensive Claude step; the apkg is cheap to re-export and its media bytes come from the requester's own workspace. Assumption: re-export/media-bundling cost is negligible vs the Claude call. If wrong (export becomes the bottleneck), add a second apkg-bytes cache layer keyed on the same key + media digest.

2. **User-agnostic, content-addressed key (no `owner` column, no FK).** *(alternative rejected: per-user cache.)* Two users uploading the byte-identical file with identical settings share one entry — maximizes hit rate. Assumption/privacy check: no leak, because the second uploader supplied that same content and receives cards derived only from it (Claude never sees media bytes, only the filename list which is in the key; the requester's own media bytes are bundled at export). If this is ever deemed too aggressive, add `owner` to the key — it only lowers the hit rate, never correctness.

3. **`isPaying` and `comprehensive` are in the key.** *(alternative rejected: content+settings only.)* Floor-v1 (`comprehensive && isPaying`) produces materially more cards, so a free and a paid user with the same content get different output. Missing these would be the exact "serve wrong output" trap the issue warns about.

4. **Prompt version = a live digest of `SYSTEM_PROMPT`, not a hand-bumped constant.** *(alternative rejected: manual `PROMPT_VERSION = 2`.)* Any edit to the system prompt changes the digest and retires prior entries with zero ceremony — ties correctness to the actual prompt instead of a number someone forgets to bump. Model id is a separate explicit constant so a model swap invalidates the same way (per the cost-envelope checklist). Assumption: per-call prompt fragments (card-style/size/field-mapping) are deterministic functions of settings already in the key, so only the global `SYSTEM_PROMPT` needs its own digest.

5. **Vision/`pdfImageFallback` path is never cached.** *(alternative rejected: cache it too.)* Its Claude input is page-image bytes read from disk, not captured by the key — caching would risk cross-serving on same-text/different-image inputs. The text path, whose entire Claude input is the HTML + the keyed settings, is the only safe surface. If vision caching is wanted later, add a per-page image-bytes digest to the key.

6. **Content normalization is minimal (CRLF→LF + trim), not whitespace-collapsing.** *(alternative rejected: collapse all whitespace.)* Aggressive normalization risks colliding semantically-distinct inputs (whitespace matters inside `<pre>`); minimal normalization still catches re-exports that differ only in line endings/padding. If re-exports commonly differ in ways minimal normalization misses (lowering the hit rate), widen it carefully.

7. **30-day read-time TTL, no wired reaper.** *(alternative rejected: a startup reaper job.)* A hit is always correct (the key encodes model+prompt+content+settings), so TTL is storage hygiene only. Wiring a reaper touches `server.ts` (a hard-rail path) and is scope creep; `deleteOlderThan` is ready for a future job. If table growth becomes a concern before that, run `deleteOlderThan` via an ops command.

8. **Wrapper made generic + dependency-free** to keep `conversionResultCache.ts` from importing `ClaudeService` types (which would add a circular-dependency warn mirroring the existing `generateDeckInfoFromPdfImages` one). `card_count` is computed in the repository (which owns the `DeckInfo` shape) rather than passed through the generic wrapper.


aalemayhu (commented):
Review pass: 21 new cache tests + 4 end-to-end green, full server suite green outside the documented environmental set, arch clean, migration-reviewer verdict SHIP with both minors applied (atomic hit counter, created_at index). Hard rail (migration) — waiting for Alexander to merge.

PR #4337, opened 2026-09-01
fix: stamp Notion conversion logs and harden worker failures

### LINKED ISSUE #4325: Correlation gaps and latent defects around the Notion conversion path (review follow-ups from #4317)
Follow-ups from the execution-verified review of https://github.com/2anki/server/pull/4317 (merged; verdict was merge-safe — none of these blocked). Grouped by class, each traced to file:line in the review comment on the PR.

### Correlation gaps (the PR's goal, incomplete on the paths that most need it)
- `NotionController.ts:343` — the fire-and-forget `.catch` on `runConversion` and the outer enqueue-failed catch log without requestId/jobId/owner while `res.locals.requestId` is in scope (~4-line fix).
- `conversionPool.ts:221` — the notion-token-expired early return emits no stamped line; the most common worker-side failure grep-joins to nothing.
- `performConversion.ts:462` — the bare `console.error(error)` stack dump is unstamped; interleaved workers make stacks unattributable.
- `performConversion.ts:224` — start line lacks the `[conversion]` prefix and no success/terminal line carries requestId (duration/completion underivable); failure-line objects render requestId on its own physical line (hoist into the message string like recordClaudeUsage's `request=`).
- `persistJobFailureWithRetry.ts:50` + the three telemetry-helper error lines (`performConversion.ts:84/98/159`) — unstamped despite sharing the `[conversion]` prefix.
- Log key `jobId` binds the Notion page UUID, but users see the DB row id from the 202 — join fails on the one id a user can report.

### Pre-existing defects in the touched functions
- `NotionController.ts:343` — Notion path doesn't classify `isWorkerTerminationError` the way the upload path does (UploadService.ts:562/1058): a killed worker strands the job `started` forever instead of `interrupted`.
- `conversionPool.ts:229` — token-expired exit never fires `track('conversion_failed')`: funnel under-counts notion_token_expired to zero.
- `performConversion.ts:437` — `MarkNotionTokenInvalidUseCase.execute` awaited unguarded inside the catch; a DB blip escapes past `persistJobFailureWithRetry`, stranding the job non-terminal.
- `performConversion.ts:224` — raw `req.body.id` interpolated into the log template (CWE-117 log injection; controller checks only `!id`).
- `performConversion.ts:244` — workspace dir orphaned when settings/rules load rejects before `workspaceLocation` is assigned (rmSync gated on it).

### Hardening
- `conversionPool.ts:233` — `ConversionWorkerRequest` hand-transcribed field-for-field into `ConversionRequest` with no structural link; a dropped field compiles clean and arrives undefined (the exact bug class #4198 fixed).
- Tests assert plumbing, not log output — a refactor dropping requestId from the log objects stays green.
- Webhook routers mount above `requestIdMiddleware` (server.ts:163-165 vs 171); no client surface retains the echoed X-Request-Id; inbound UUID-shaped X-Request-Id honored verbatim.

aalemayhu:
## What
Closes the correlation gaps and latent defects that the execution-verified review of #4317 recorded on the Notion conversion path. Every conversion lifecycle log line now carries `[conversion] job=<dbRowId> request=<id>`, and five pre-existing defects in the touched functions are fixed.

## Why
The #4198 request-id threading stopped short on exactly the catch paths that a failing conversion hits, so a user-reported job could not be joined to server logs. `Closes #4325`.

### Correlation gaps (all fixed)
- **NotionController fire-and-forget `.catch` + enqueue-failed catch** — both now log `requestId`, `jobId` (the DB row id from the 202), and `owner`. The worker `.catch` also classifies `isWorkerTerminationError` (see defect below).
- **conversionPool token-expired early return** — now emits a stamped `[conversion] job=… request=… notion token expired` line before failing the job.
- **performConversion bare `console.error(error)`** — removed. It was an unstamped duplicate of the stamped `[conversion] … failed` line, which already logs the error object (with stack). Nothing is lost; the unattributable dump is gone.
- **Start line + success/terminal line + hoisted requestId** — start line is now `[conversion] job=… request=… started`; a new `… completed cards=N duration_ms=M` line makes completion/duration derivable; every failure/state line hoists the ids into the message string (like `recordClaudeUsage`'s `request=`) instead of rendering `requestId` on its own object key.
- **persistJobFailureWithRetry warn + the three telemetry-helper error lines (`recordUnsupportedBlocks`/`recordConversionOutputStats`/`recordDeckScore`)** — all now receive and stamp the same `[conversion] job=… request=…` prefix.
- **`jobId` bound the Notion page UUID; users see the DB row id** — log key is now `job=<jobDbId>` (the id returned in the 202, i.e. the one a user can report); the raw Notion page id moves to a structured `pageId` field.

### Pre-existing defects in the touched functions (all fixed)
- **Killed Notion worker stranded the job `started`** — the controller `.catch` now marks it `interrupted` with the existing `WORKER_INTERRUPTED_REASON`, mirroring the upload path (UploadService.ts:562/1058).
- **Token-expired exit never fired `conversion_failed`** — now tracks it with `reason: notion_token_expired`, so the funnel stops under-counting to zero.
- **`MarkNotionTokenInvalidUseCase.execute` awaited unguarded inside the catch** — now wrapped so a DB blip can no longer escape past `persistJobFailureWithRetry` and strand the job non-terminal.
- **Raw `req.body.id` interpolated into the start log template (CWE-117)** — the page id is out of the format string; it goes to a structured field where `util.inspect` escapes control characters.
- **Workspace dir orphaned when settings/rules load rejects before `workspaceLocation` was assigned** — `CreateJobWorkSpaceUseCase` now cleans up the workspace it created if the load rejects, then rethrows.

### Hardening
- **Hand-transcribed `ConversionWorkerRequest` → `ConversionRequest`** — the request is now spread structurally (`{ ...request, api }`) and `ConversionRequest extends ConversionWorkerRequest`, so a dropped field can no longer compile clean and arrive `undefined` (the #4198 bug class). The shared shape moved to `src/lib/conversionRequestTypes.ts` so the two modules link through a type instead of a hand copy, with no runtime import cycle.
- **Tests asserted plumbing, not log output** — added assertions on the actual log lines (start prefix + no page-id injection, completion line, requestId hoisted into the failure message, stamped retry warning, stamped token-expired line) so a refactor that drops the ids from the log objects now goes red.

## Not actionable (observations, no change warranted)
- **Webhook routers mount above `requestIdMiddleware` (server.ts:163-165 vs 171).** Moving them is a middleware-order change on a hard-rail path (`src/server.ts`), for marginal benefit — webhook traffic is server-to-server, there is no user to report an echoed id — and it risks the raw-body ordering that HMAC verification depends on. It would also flip this otherwise-safe PR onto the manual hard-rail track. Deliberately out of scope; a separate PR if we ever want webhook correlation.
- **No client surface retains the echoed `X-Request-Id`.** Surfacing the id in the error UI so a user can quote it is a user-facing UX change that needs trio review; out of scope for an internal-correlation follow-up.
- **Inbound UUID-shaped `X-Request-Id` honored verbatim.** This is deliberate (a proxy or the native app threads its own id) and the value is UUID-validated, so it cannot carry a log-injection payload. Not a defect.

## Measuring success
- Grep prod logs for `[conversion] job=<id>` and join by `request=<uuid>` across controller, worker, and failure lines — previously the failure paths dropped out of the join. The new `… completed … duration_ms=` line makes per-job duration derivable.
- `/api/ops/metrics`: `conversion_failed` with `reason: notion_token_expired` should now be non-zero (it was structurally zero before).

## Testing
- Unit tests added/updated: NotionController (worker-failure log fields, interrupted-on-drain, enqueue-failed log fields), conversionPool (stamped token-expired line + `conversion_failed` track, structural spread), performConversion (start-line prefix + no page-id injection, completion line, hoisted-requestId failure line, mark-token-invalid guarded), persistJobFailureWithRetry (stamped retry warning), CreateJobWorkSpaceUseCase (orphan cleanup).
- Full server Jest suite: 697 suites / 6815 tests pass; the 9 failing suites are the documented environmental exclusions only — 8 parser/canary suites that spawn Python (no `create_deck/.venv` in the worktree) and `getPageCount` (pdfinfo). None of the five touched suites are among them.
- Full web Vitest suite: 384 files / 3505 tests pass (the changelog loader validated the new entry). `tsc -p . --noEmit` clean; web `tsc --noEmit` clean; `pnpm format:check` clean; `pnpm arch` 0 errors (no new circular warning).

## Browser check
Browser check: not applicable — the only `web/src` change is a changelog JSON data file; no component or rendered logic changed.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-01-notion-conversion-interrupted-retry.json` — "A Notion conversion caught by a server restart now tells you to convert again right away". Reuses the already-shipped `WORKER_INTERRUPTED_REASON` string, so no new i18n string. The rest of the PR is internal observability/hardening with no user-visible surface.

## Risks
Low. The changes are logging-format and error-handling within existing catch paths; the one behavioral change (marking a drained Notion worker `interrupted`) mirrors the upload path and guards against overwriting a `done` job. Rollback is a straight revert — no schema, no migration, no new env flag.

## Goal alignment
None on the acquisition funnel — internal reliability/observability. It does move the retention/quality lever: stuck-forever Notion jobs now resolve to a clear retry, the token-expired funnel stops mis-reading zero at `/api/ops/metrics`, and every failed conversion becomes joinable to a user-reported job id, which shortens support turnaround on the paid Notion path.

### DIFF
--- a/src/controllers/NotionController.test.ts
+++ b/src/controllers/NotionController.test.ts
@@ -10,6 +10,7 @@ import { INotionRepository } from '../data_layer/NotionRespository';
 import { IEmailService } from '../services/EmailService/EmailService';
 import UsersRepository from '../data_layer/UsersRepository';
 import { buildNativeOAuthState } from '../services/NotionService/nativeOAuthState';
+import { WORKER_INTERRUPTED_REASON } from '../lib/workerTermination';
 
 jest.mock('../lib/conversionPool', () => ({
   __esModule: true,
@@ -392,8 +393,9 @@ describe('NotionController', () => {
       });
     });
 
-    it('fires runConversion without awaiting and catches worker rejections', async () => {
+    it('fires runConversion without awaiting and logs worker rejections with the request id, job id, and owner', async () => {
       setupConvertMocks();
+      res.locals = { owner: 'owner1', requestId: 'req-abc-123' };
       const consoleErrorSpy = jest
         .spyOn(console, 'error')
         .mockImplementation(() => undefined);
@@ -406,8 +408,63 @@ describe('NotionController', () => {
 
       await new Promise((resolve) => setTimeout(resolve, 10));
       expect(consoleErrorSpy).toHaveBeenCalledWith(
-        'notion convert worker:',
-        workerError
+        '[notion/convert] worker failed',
+        expect.objectContaining({
+          requestId: 'req-abc-123',
+          jobId: 77,
+          owner: 'owner1',
+          error: workerError,
+        })
+      );
+      consoleErrorSpy.mockRestore();
+    });
+
+    it('marks the job interrupted when the worker is killed by a pool drain', async () => {
+      setupConvertMocks();
+      const updateJobStatus = jest.fn().mockResolvedValue(undefined);
+      (JobRepository as unknown as jest.Mock).mockImplementation(() => ({
+        updateJobStatus,
+      }));
+      const consoleInfoSpy = jest
+        .spyOn(console, 'info')
+        .mockImplementation(() => undefined);
+      (runConversion as jest.Mock).mockRejectedValue(
+        new Error('Terminating worker thread')
+      );
+
+      await controller.convert(req as express.Request, res as express.Response);
+
+      expect(res.status).toHaveBeenCalledWith(202);
+
+      await new Promise((resolve) => setTimeout(resolve, 10));
+      expect(updateJobStatus).toHaveBeenCalledWith(
+        'page-abc',
+        'owner1',
+        'interrupted',
+        WORKER_INTERRUPTED_REASON
+      );
+      consoleInfoSpy.mockRestore();
+    });
+
+    it('logs the request id and owner when the enqueue itself fails', async () => {
+      setupConvertMocks();
+      res.locals = { owner: 'owner1', requestId: 'req-enqueue-9' };
+      (FindOrCreateJobUseCase as jest.Mock).mockImplementation(() => ({
+        execute: jest.fn().mockRejectedValue(new Error('db down')),
+      }));
+      const consoleErrorSpy = jest
+        .spyOn(console, 'error')
+        .mockImplementation(() => undefined);
+
+      await controller.convert(req as express.Request, res as express.Response);
+
+      expect(res.status).toHaveBeenCalledWith(500);
+      expect(consoleErrorSpy).toHaveBeenCalledWith(
+        '[notion/convert] enqueue failed',
+        expect.objectContaining({
+          requestId: 'req-enqueue-9',
+          owner: 'owner1',
+        })
       );
       consoleErrorSpy.mockRestore();
     });
--- a/src/controllers/NotionController.ts
+++ b/src/controllers/NotionController.ts
@@ -2,6 +2,10 @@ import { Request, Response } from 'express';
 
 import { runConversion } from '../lib/conversionPool';
 import {
+  isWorkerTerminationError,
+  WORKER_INTERRUPTED_REASON,
+} from '../lib/workerTermination';
+import {
   InProgressJobError,
   JobLimitError,
 } from '../lib/storage/jobs/helpers/errors';
@@ -292,6 +296,9 @@ class NotionController {
       },
     });
 
+    const safeString = (v: unknown): string | undefined =>
+      typeof v === 'string' && v.length > 0 ? v : undefined;
+
     try {
       if (!paying) {
         const usersRepository = new UsersRepository(database);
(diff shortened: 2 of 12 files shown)


aalemayhu (commented):
Reviewed clean: requestId threading consistent across the conversion path, no token or email values in any new log line (state descriptions only — spot-checked the full diff), the interrupted state mirrors the upload path, the token-invalid guard cannot swallow the job-failed write, and the shared request type is type-only with no runtime cycle. Full server suite 6815 green outside the documented environmental set; tests assert log output.

PR #4322, opened 2026-09-01
feat: Semester Pass and term-start winback

aalemayhu:
## What
Adds a **Semester Pass** — a one-time `120d` (~4 month) pass that mirrors the existing Day/Week pass rails end to end: a `POST /api/checkout/pass/120d` checkout route off `PASS_120D_PRICE_ID`, the Stripe webhook grant, the anonymous claim flow, the #4327 checkout-email auto-link, the account access banner, and a third pricing card localized in all 10 locales. The pass-winback audience now also includes lapsed Semester Pass buyers, so the existing term-start winback ops command reaches them with no new machinery.

Closes #4305.

## Why
79% of churn is lifecycle ("exams are over"), not price. Day and Week passes cover nothing longer than a week, so a returning student who wants to pay for a term either resubscribes (past the legacy-rate one-way door) or doesn't come back. A Semester Pass matches billing to the exam calendar and turns lifecycle churn into scheduled, one-time revenue.

## How
- **New pass kind `120d`** added to the single source of truth: `PASS_DURATION_MS` (120 days), `AnonymousPassKind`/`isAnonymousPassKind`, `PassKind`, and `passKindLabel` → "Semester Pass".
- **Every hardcoded `24h`/`7d` pass-kind gate extended together** through the shared `isAnonymousPassKind` predicate — the spec called out three (webhook grant branch, `isAnonymousPassKind`, `ValidateAnonymousPassUseCase`); implementation found **two more that would silently take money and grant nothing** and fixed them in the same PR (see Deviations): `GrantUnclaimedPassUseCase` (the #4327 claim/auto-link path) and `mapEntitlement`. `PassUnlockMonitorService`'s safety set gained `120d` too so a failed grant is still monitored.
- **Checkout:** `CheckoutRouter` mirrors the `7d` route (503 when `PASS_120D_PRICE_ID` unset); `config.ts` WARN_VARS + `src/env.example` document the new var.
- **Web:** third `PricingCard` in `PassCards` (grid reflows via `auto-fit`, so `LimitPage`'s existing 2-card layout is untouched — the Semester card renders only where `onSemesterPass` is passed), `PricingPage` handler/state, `Backend.startPassCheckout`, `AccessBanner` label, `payment.links` price, `pricingFaq`, and `pricing` + `pricingtable` i18n keys across all 10 locales (Stripe plan names + price tokens kept verbatim per i18n rules).
[...]

PR #4340, opened 2026-09-02
feat: dedup AI cards against the user's existing decks

### LINKED ISSUE #4243: Dedup AI cards against the user's existing decks, not only the same upload
## Problem

#3971 (closed #3917) suppresses duplicate cards across files **inside one upload**. Nothing looks at decks the user already converted. Users who upload a chapter today and its lecture transcript next week still get the same fact twice.

## Signal

Two independent heavy AI-conversion users now:

- user 18996, the original #3917 request (July 2026).
- A second paying med-school user (email thread 2026-08-26) who, on hearing about the same-upload dedup, immediately asked whether everything under My decks is now connected and what deleting a deck would do. That is the mental model they expected; the answer today is no.

Filing so the second signal is not lost. Not scheduled; a third request or a retention read on multi-upload users should promote it.

## Shape

Candidate: at conversion time, load the normalized fronts of the user's last N completed AI jobs (the `collectExistingFronts` / `buildTopUpInstruction` path #3971 exported) and thread them into the prompt plus the deterministic backstop. Open questions: N and recency window, cost of the extra prompt tokens on the path that is ~88% of Anthropic spend, whether users want an opt-out per upload.

## Not in scope

Deduping already-exported decks retroactively, or touching Anki-side notes.

## Metric

Per-user card count per overlapping-source upload, and 30-day retention of users with 2+ AI conversions on overlapping sources. Read via `/api/ops/metrics` and prod psql.

aalemayhu:
## What
Extends the existing same-upload dedup so an AI conversion also suppresses cards the user already has in their **other** decks. Same-upload dedup (#3917/#3971) only ever saw one upload; a user who converted a chapter today and its lecture transcript next week still got the same fact twice. This seeds the existing cross-file backstop with the user's prior card fingerprints so a byte-identical (normalized) card already in another of their decks is dropped from the new deck.

Closes #4243.

## Why
Two independent heavy AI users asked for this (issue #4243). The mental model paying med/law users expect is "everything under My decks is connected." Retention lever: fewer redundant cards per overlapping-source upload for exactly the multi-upload power users who churn on lifecycle, not price.

## How
- **Data source (the honest boundary).** The server retains **no card-level content history** for AI uploads: `jobs` holds aggregate `conversion_report` only, `uploads` holds file metadata, and `card_guids` is keyed by Notion `blockId` (never populated for AI-upload cards). So a real implementation needs a place to persist per-user card identities. New `ai_card_fingerprints` table: `(owner → users.id ON DELETE CASCADE, fingerprint char(64), created_at)`, unique `(owner, fingerprint)`, index `(owner, created_at)`.
- **Reuses the existing mechanism, does not invent a parallel one.** `CrossFileDedupState` gains a `historicalKeys` seed and a `crossDeckSuppressed` counter. `absorbFileIntoCrossFileDedup` now keys on `cardFingerprint(card)` = `sha256(cardDedupeKey(card))` (same normalized NUL-separated front+back, just hashed so the persisted store holds a fixed-width surrogate, never study content). Seeding history into `seenKeys` makes the same deterministic backstop suppress cross-deck duplicates.
- **Load/record mirror the guid ledger exactly** (no DB in the worker): `UploadService.loadExistingCardFingerprints` seeds on the main thread (threaded to the worker as `existingCardFingerprints`), `collectNewFingerprints(state)` (produced minus seed) is recorded via `recordCardFingerprints` after the run. Wired on both the async and sync upload paths.
- **Gate** `shouldDedupeAcrossDecks` (signed-in + paying + `claudeAIFlashcards` + ≥1 AI content file), OR'd with the existing `shouldDedupeAcrossFiles`. Anonymous / non-paying / non-AI uploads are bit-for-bit unchanged.

## Cost envelope (paid inference path)
- **Latency impact:** none material — comparison is in-process hashing + one indexed `SELECT ... LIMIT 5000` before the run and one chunked `INSERT ... ON CONFLICT DO NOTHING` after.
- **Token / cost delta:** **zero.** History is seeded as hashes only and never reaches the prompt (`state.fronts` still accumulates within-upload fronts only). No prompt-hint tokens added, no extra Claude calls. Cards suppressed by the backstop are removed *after* generation, so this holds spend flat and can only reduce downstream re-conversion.
- **Prompt caching:** unaffected — the Claude call inputs are untouched.
- **Conversion result cache (#4336):** untouched. Dedup runs strictly after `withConversionResultCache` returns `DeckInfo[]`; the cache stays user-agnostic and its key semantics are unchanged. Two users converting the same file share one cache entry but each dedupes only against their own history (per-owner isolation).

1. **Backstop only, no prompt hint.** Decision: suppress cross-deck duplicates with the deterministic post-generation backstop; do NOT thread historical fronts into the prompt. Rejected alternative: the issue's "thread prior fronts into the prompt" candidate — it adds input tokens on the path that is ~88% of Anthropic spend and does not reliably reduce output. Assumption: the constraint "reduce or hold spend, never add calls" outranks a marginal model-side hint. If wrong: we could add a bounded, sampled front hint later, but only behind a spend read.
2. **Store a sha256 hash, not the card text.** Decision: persist `sha256(cardDedupeKey)`. Rejected: storing normalized fronts/backs (needed only for the prompt-hint path we rejected). Assumption: equality is all the backstop needs, and not retaining study content server-side is the safer privacy/retention posture. If wrong (you want the prompt hint): this table would need to store text or a companion column.
3. **Recency window = 5000 most-recent fingerprints, no time cutoff.** Decision: seed the newest 5000 per owner. Rejected: a 30-day window — study material is long-lived, and a time cutoff silently reintroduces duplicates for the exact returning user this serves. Assumption: 5000 bounds query + worker-payload cost with the `(owner, created_at)` index; unique `(owner, fingerprint)` bounds table growth to distinct-cards-ever. If wrong: tune `AI_DEDUP_HISTORY_LIMIT`.
4. **Always-on, no per-upload opt-out.** Decision: no toggle. Rejected: a per-upload opt-out (issue's open question). Assumption: the backstop is loss-free *by construction* — it drops only a card whose normalized front AND back already exist in another of the user's decks; the user still has that exact card. **Surprise risk to weigh:** a user could want the same fact in two decks as separate study contexts; today it's dropped from the new deck silently. I did not add a user-facing notice (would need 10-locale i18n + a web surface); the `cross_deck_duplicates_suppressed` metric is the read to decide whether a notice/opt-out is worth building. If the day-7 read shows high suppression volume, add a notice.
5. **Scope = file-upload AI path only.** Decision: cross-deck dedup covers `doGenerationWork` / `UploadService` (both async and sync upload paths). Rejected/deferred: Notion-page AI conversions (different orchestration, no worker accumulator). This matches the issue's scenario (uploading a chapter, then a transcript). If wrong: a follow-up can seed the Notion path from the same table.

## Measuring success
- `ai_conversion_completed` now carries `{ cross_deck_duplicates_suppressed, historical_cards_considered }` (server-only event, no new `KNOWN_EVENTS` entry — free-form props on the existing event). Day-7 prod read: sum `cross_deck_duplicates_suppressed` over multi-upload users via `/api/ops/metrics` + prod psql; the issue's metric (per-user card count per overlapping-source upload) is the retention read.
- Log/DB confirmation: `select count(*), count(distinct owner) from ai_card_fingerprints;` shows the ledger filling; the event props show suppression firing.

## Testing
- Unit tests added: `cardFingerprint` (stable sha256, collision only on front+back), cross-deck seed suppression + counter split (cross-file vs cross-deck), `collectNewFingerprints` excludes the seed, history never reaches `state.fronts`, `AiCardFingerprintRepository` generated SQL (PG `on conflict do nothing`) + behavior (dedup/empty), `shouldDedupeAcrossDecks` gate, worker seeds/records/returns fingerprints and emits the cross-deck props, single-file signed-in path now threads a state.
- Full server Jest: **6969 passed, 4 failed, 25 skipped, 7 todo (7005 total)**. The only failures are the documented environmental `getPageCount` pdfinfo ENOENT suite (binary not installed locally); all feature-adjacent suites (ClaudeService, worker, PrepareDeck, UploadService, GeneratePackagesUseCase, AiCardFingerprintRepository) green.
- `tsc -p . --noEmit`: clean. `pnpm format:check` (tree-wide): clean. Web changelog vitest: 4 passed.
- Kanel: `ai_card_fingerprints` migration applied to the local DB and `src/data_layer/public/AiCardFingerprints.ts` regenerated by kanel in the same PR (not hand-written); unrelated regen churn discarded.

## Browser check
Notes: No web runtime change — the only web file is a changelog JSON (validated by the loadChangelog vitest). Server-only behavior change.

## Changelog
"AI decks no longer repeat a card you already have in another one of your decks" (`web/src/pages/WhatsNewPage/changelog/2026-09-02-cross-deck-ai-dedup.json`).

## Risks
- Behavioral surprise (decision 4): a card intentionally wanted in two decks is dropped from the newer one. Mitigation: loss-free by construction (card still exists elsewhere) + the metric to decide on a notice. Rollback: revert the PR; the migration `down` drops the table and dedup falls back to same-upload behavior — no card is lost, only the suppression signal.
- Hard-rail PR (adds a migration) — not self-merging; awaiting review.

## Goal alignment
Retention lever for multi-upload paying power users (the med/law lifetime/Auto-Sync cohort), the segment the issue flags. Metric to move: per-user card count per overlapping-source upload + 30-day retention of users with 2+ AI conversions on overlapping sources, read via `/api/ops/metrics` + prod psql. Simpler/faster/more beautiful: fewer redundant cards to wade through in a new deck.


aalemayhu (commented):
Review pass: migration-reviewer verdict was BLOCK on one item — the duplicate CREATE TABLE against the migration that leaked to main via #4322 — fixed by dropping the file; the code and committed kanel types now target the live table. Everything else passed: correct indexes for both query paths, CASCADE-safe against inactive-user deletion, idempotent chunked inserts, cache-error fallthrough so a fingerprint failure never fails a conversion, zero added Claude spend. Full server suite green outside the documented environmental set. Hard rail (data_layer/public) — waiting for Alexander to merge.

PR #4341, opened 2026-09-02
docs: remove the design-partner council SOP

aalemayhu:
Alexander decided against running the council as a standing program (the inbox already delivers the same power-user signal unprompted — the roster's top candidate has 88 unprompted messages). Removing the SOP rather than keeping unowned process debt; git history keeps it recoverable if a pre-GA surface ever wants a private beta list. The local invite prep file is deleted too. Docs-only: no deploy, no changelog.


aalemayhu (commented):
Docs-only removal, verified: the deleted SOP was the only council reference outside changelog history, no code or hook reads it, and the Downloads prep file is gone. Direct instruction from Alexander.

PR #4357, opened 2026-09-07
feat: per-user AI spend alert email and daily runaway breaker

aalemayhu:
## Summary

Trio-reviewed (pm + designer + engineer) follow-up to the AI-profitability review: per-user AI spend was invisible in ops and unbounded per account. Two invisible guardrails, no user-facing limits:

- **Ops visibility + email alert ($25/30d):** `/api/ops/ai-usage` now returns a `by_user` breakdown (top 20 spenders) rendered as a "Spend by user" table in the ops System tab, with an "over $25" badge. When any user crosses $25 of AI spend in a trailing 30 days, an alert email goes to the support CC address, deduped to one per user per 7 days via an `ai_spend_alert_sent` event.
- **Runaway breaker ($50/24h):** `guardAiSpend(userId)` runs before every metered Claude call on the three surfaces carrying ~98% of spend (conversion chunks, PDF page vision, chat). A user crossing $50 in a trailing 24h gets `AiSpendCapError` (429) and an immediate ops email (deduped 24h via `ai_spend_cap_tripped`). The threshold sits ~8× above the heaviest human month on record — this protects against a billing retry loop or a scripted client, not heavy legitimate use.

The guard **fails open** on any read/notify error so a metrics or SendGrid outage can never block a paying conversion, and skips anonymous callers (bounded by the paid gates). Both thresholds are unconditional code registered in `src/lib/limits.ts` — no env flags.

## Expected metric impact

None on funnel/MRR — margin protection. Guardrail metric: AI spend as % of net revenue (11% at review time), watched via the ops AI usage tab; the per-user table is the leading indicator.

## Testing

- Full server suite: 720 suites / 7053 tests passed (pre-existing environmental `getPageCount` pdfinfo failure only).
- Full web suite: 387 files / 3540 tests passed.
- New: SQL-shape tests for the three new repository builders, 8 unit tests for `guardAiSpend` (cap trip, alert, dedup, fail-open), per-user table + badge render test.
- `tsc -p . --noEmit` clean (all `IEmailService` test mocks updated with the new method — the deploy build typechecks tests).

No changelog entry — internal ops/observability guardrail, no user-visible behavior change (the breaker threshold is unreachable by legitimate use).

Sonar: not run locally; PR decoration will report.

## Browser check
- [x] Golden path on localhost:3000
- [x] No console errors at 375px
[...]

PR #4360, opened 2026-09-07
feat: per-template email delivery rates in ops

aalemayhu:
## Summary

Transactional-email delivery watch — the one still-open item from the support-corpus synthesis top four (P4: password reset, magic login, cancel confirmation reported undelivered in every era; one signup flow claimed success while silently sending nothing).

- `ProcessSendgridEventsUseCase` now tracks an `email_delivery_event` (template category + event type) into the existing `events` table for every recorded SendGrid event — reusing the category tagging shipped in #4329. No migration.
- New `GET /api/ops/email-delivery` aggregates per-category delivered/bounce/dropped/blocked/deferred/spamreport/unsubscribe counts plus a failure rate (hard failures ÷ attempts) over a 7d–90d window.
- New "Email delivery" section in the ops System tab: per-template table, red badge when a template hits ≥10% failure with ≥10 attempts.

Data accrues from this deploy forward — the table is empty until SendGrid events arrive with categories.

## Expected metric impact

None on funnel/MRR directly — observability. Guardrail metric: delivered-rate per template; the leading indicator for the "email never arrived" support-thread class.

## Testing

- New: SQL-shape test for the aggregation builder, service pivot/failure-rate tests, use-case test asserting the tracked event, web section render + badge + error-banner tests.
- Full server suite + full web suite green (counts in comments below if needed); `tsc -p . --noEmit` clean; `pnpm format:check` clean.

No changelog entry — internal ops/observability, no user-visible behavior change.

## Browser check
Notes: walked the upload page at 375px in the browser (page render, file select, guardrail dialog — zero console errors from this diff; the single error is the pre-existing logged-out 401 preferences probe, present on main). The Playwright MCP disconnected mid-flow, so the conversion itself was completed against the same running dev server via the API golden path: POST /api/upload/file returned 200 with a valid 69,850-byte .apkg, byte-identical to the full-browser run performed on this tree an hour earlier for #4357. This diff's only UI change is the auth-gated ops System tab, covered by the new vitest render/badge/error tests.

### DIFF
--- a/src/controllers/OpsController.ts
+++ b/src/controllers/OpsController.ts
@@ -26,6 +26,7 @@ import { GetCustomerSignalsUseCase } from '../usecases/ops/GetCustomerSignalsUse
 import { GetPassUnlockMonitorUseCase } from '../usecases/ops/GetPassUnlockMonitorUseCase';
 import { GetPaidValueMonitorUseCase } from '../usecases/ops/GetPaidValueMonitorUseCase';
 import { GetAiUsageMetricsUseCase } from '../usecases/ops/GetAiUsageMetricsUseCase';
+import { GetEmailDeliveryMetricsUseCase } from '../usecases/ops/GetEmailDeliveryMetricsUseCase';
 import { SetChatAttachmentsLifecycleUseCase } from '../usecases/ops/SetChatAttachmentsLifecycleUseCase';
 import { GrantUnclaimedPassUseCase } from '../usecases/passes/GrantUnclaimedPassUseCase';
 import { SetBlockIdIdentityUseCase } from '../usecases/ops/SetBlockIdIdentityUseCase';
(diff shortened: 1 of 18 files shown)


aalemayhu (commented):
Ship review (three-fork fan-out: security / engineering / ux-voice), verdict on this head: **pass**. Security: clean — tracked props are category+type only (no PII, track() strips matching keys anyway), aggregation builder fully parameterized with a generated-SQL test, route behind RequireOpsAccess, cardinality bounded by the HMAC-verified webhook + our own EMAIL_CATEGORIES constants, no error leakage. Engineering: clean — duplicate sg_event_ids verified to skip tracking (now pinned by a test), deferred excluded from both numerator and denominator so retries can't skew rates, events-table growth ~doubles an already-accepted write (buffered sink), all 28 positional controller args walked and position-correct. UX: clean — register matches sibling ops sections, empty-state copy made timeless, nine-column table wrapped in tableScroll, Unsubscribe column added so fetched counts render, a11y wiring mirrors siblings. Known residue: 'blocked' column will likely read 0 (SendGrid reports blocks as bounce events — inherited from the existing TRACKED_EVENT_TYPES, not this PR).

PR #4363, opened 2026-09-08
fix: fire account_created for every signup path, not just password

### LINKED ISSUE #4362: Ops upload-funnel Signup tile undercounts real signups by ~84%
Found during a read-only funnel audit (2026-09-07 overnight run).

The ops Upload funnel tab Signup tile keys off the `account_created` event, which recorded 58 signups in the last week — against 353 actual `users.created_at` rows and 309 `signup_completed` events in the same window. Any rate computed off it (e.g. `download_to_signup_rate_pct`) is unreliable.

Relevant files:
- `src/services/ops/UploadFunnelService.ts` (keys off `account_created`, not `signup_completed`)
- `web/src/pages/OpsPage/UploadFunnelTab.tsx`

Fix direction: key the tile off `signup_completed` (or `users.created_at`), or find and fix why `account_created` fires for only ~16% of signups. Decide which event is canonical and align the funnel service on it.

Context: the same audit found the biggest funnel leak is Downloaded → pricing-page (72% of successful downloaders never see the pricing page) — that insight is trustworthy because it was computed from cohort traces on `users`/`events` directly, not from this tile.

aalemayhu:
## Summary

Fixes https://github.com/2anki/server/issues/4362 — the ops signup funnel undercounted real signups by ~84% because `account_created` only fired on the password registration path. Google, Microsoft, Apple (web + native share `#upsertAppleUser`), Notion OAuth, and magic-link signups never emitted it (58 tracked vs 353 actual users last week).

- The event now fires inside `UsersService.register` — the single creation choke point — so no current or future signup path can forget it.
- A `RegisterTelemetry` argument threads `method` (`password | google | microsoft | apple | notion_oauth | magic_link`), the anon-id cookie, and the first-touch referrer into the event; the controller's duplicate emission is removed.
- `UploadFunnelService` needs no change — its `account_created` key now sees every signup, and the event gains a `method` prop for free.

Closes #4362

## Expected metric impact

Ops upload-funnel Signup tile becomes accurate from this deploy (expect ~6x jump in the tile — that's the undercount closing, not a growth spike). `download_to_signup_rate_pct` becomes trustworthy. New: signup method split available via the `method` prop.

## Testing

- Full server suite: 724 suites / 7074 tests passed (documented `getPageCount` pdfinfo env failure only).
- `UsersService.register` tests extended: telemetry-threaded event shape + method-unknown fallback.
- All five controller register paths now assert their `method` value reaches `register`.
- `tsc -p . --noEmit` clean; `pnpm format:check` clean.

No changelog entry — internal analytics correctness, no user-visible behavior change.

Browser check: not applicable — server-only diff, no web/src changes.

### DIFF
--- a/src/controllers/UsersControllers.test.ts
+++ b/src/controllers/UsersControllers.test.ts
@@ -643,7 +643,8 @@ describe('UsersController.requestMagicLink', () => {
       'al@example.com',
       'login',
       null,
-      undefined
+      undefined,
+      expect.objectContaining({ anonymousId: null })
     );
   });
 
@@ -662,7 +663,8 @@ describe('UsersController.requestMagicLink', () => {
       'al@example.com',
       'login',
       null,
-      '/upload'
+      '/upload',
+      expect.objectContaining({ anonymousId: null })
     );
   });
 
@@ -685,7 +687,8 @@ describe('UsersController.requestMagicLink', () => {
       'al@example.com',
       'login',
       null,
-      undefined
+      undefined,
+      expect.objectContaining({ anonymousId: null })
     );
   });
 
--- a/src/controllers/UsersControllers.test.ts
+++ b/src/controllers/UsersControllers.test.ts
@@ -171,11 +171,16 @@ describe('UsersController.register', () => {
 
     await controller.register(req, res, next);
 
-    expect(trackMock).toHaveBeenCalledWith('account_created', {
-      userId: 1,
-      anonymousId: 'anon-abc-123',
-      props: { signup_origin: '/notion-to-anki' },
-    });
+    expect(register).toHaveBeenCalledWith(
+      '',
+      'hashed',
+      'jane.doe@example.com',
+      '/notion-to-anki',
+      expect.objectContaining({
+        method: 'password',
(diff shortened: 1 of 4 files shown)


aalemayhu (commented):
Ship review (security + engineering forks), verdict on this head: **pass**. Security: clean — event props are three sanitized keys (allowlist-regexed origin, host-only referrer, closed method union), telemetry is server-constructed at every site, no req.body flow, EventsSink is sync-buffered so tracking can never fail a login, and the anonymousId linkage is byte-identical to the removed controller emission. Engineering: clean — all six creation sites thread telemetry, the event fires only after a successful insert inside the transaction, exactly one account_created emitter remains, all downstream consumers key on name/user_id and are unaffected, and buildSignupProps has zero remaining references. Both reviewers' minors are fixed on this head: the anon-id cookie now threads through the four OAuth paths and the magic-link path (previously dropped), and a service test pins a non-null anonymousId through to the event.

aalemayhu (commented):
Re-posting the ship-review pass marker for the current head — the prior marker carried a mistyped SHA (same review verdict: security clean, engineering clean, both minors fixed on this head).

PR #4367, opened 2026-09-08
fix: map dominant Q/A tables to one card per row

### LINKED ISSUE #4366: DOCX Question/Answer table: bullets in the answer cell fan out into separate cards
From emoji feedback id 107 (rating 1, downloads/deck_done, 2026-09-08): "Bullet points in the right-handed column were each turned into a separate card instead of sticking to the Question column and Answer column."

A user this week converted a .docx whose content is a two-column Question/Answer table with bullet points inside the answer cells. Every bullet became its own card; the table structure was ignored. They retried the same conversion six times in under an hour before leaving the 1-star rating — runtime entry point confirmed from prod logs as the sync upload path (`POST /api/upload/file`, mimetype docx) feeding `PrepareDeck` → `DeckParser`.

Root cause: the correct behavior already exists — the `table` induction rule in `src/lib/parser/induction/induceCardsFromDom.ts:166` maps each top-level table row to one card (first cell = front, remaining cells = back, cell contents kept intact). But induction runs only from the zero-cards rescue (`DeckParser.induceCardsFromStructure`, `src/lib/parser/DeckParser.ts:658`). A Q/A-table document never reaches it, because the bullet/list flattening path happily produces N wrong cards first.

Fix direction (needs care, this is parser core): when the parsed DOM's dominant structure is a two-column table (e.g. most of the document's text lives inside one `<table>` with ≥2 columns and ≥2 rows), prefer the table row→card mapping over per-bullet fan-out — either as a pre-pass in `handleHTML` (like the page-list/paragraph/lines detections) or by letting a high-scoring table induction outrank a low-quality flattened parse. The locked-set rule applies: extending when induction may run is a decision, not a silent change; parser core warrants the heavier review lane.

Repro shape for the regression test: a docx-derived HTML doc whose body is one `<table>` with header row Question|Answer and 2+ data rows where the answer cell contains a `<ul>` of 3 bullets — expect one card per row with all bullets on the back, not one card per bullet.

First-time-fix checklist applies (.claude/docs/first-time-fix.md): entry point pinned (sync upload, docx), reporter's data shape captured above; fix must be verified against that shape, and any reply to the reporter says "shipped a change; your next conversion should show X" — not "fixed" — until confirmed on their path.

aalemayhu:
Closes #4366

## Root cause

Two mechanisms behind "bullets in the answer column each became a separate card":

1. **Web docx/html uploads (parser bug, this PR's main fix).** A body that is one two-column Q/A table produces zero cards from the normal parse, so every `<li>` fans out into a front-only note. The `columns` rescue then produces the correct one-card-per-row deck — but with fewer than 3 data rows it is rejected by `FLOOR_MIN_CARDS = 3` and the fan-out ships. Fixture-verified red/green: the 2-row case failed exactly this way; the 4-row case already passed via rescue. Prod corpus: 191 docx + 23 html fan-out conversions in the last 30 days (median back length 0).
2. **The reporter's actual path (MCP).** `create_deck` receives pre-split `{front, back}` cards — the calling model chose one card per bullet before 2anki was involved. The `cards` schema description now instructs the caller to preserve the source's card boundaries (a table row = one card, bullets together on the back). Tool descriptions are the only server-side lever on that path.

## Changes

- `hasDominantTable(dom)` (induction module): exactly one non-nested table, ≥80% of body plain text, ≥2 data rows of **exactly two cells** (header rows excluded; any wider data row disqualifies — a timetable/data matrix is not a Q/A pairing and stays on the old pipeline).
- `DeckParser.induceDominantTableCards`: runs first in the zero-cards path, routes straight to the existing `columns` rule, exempt from `FLOOR_MIN_CARDS` **only** — every other floor check (over-split, duplicate fronts, blank backs) still applies via `clearsQualityFloorExceptCount`. Predicate miss or floor fail → identical fall-through.
- MCP `create_deck` `cards` description: card-boundary guidance.
- Changelog entry.

## Decisions

- Reused the existing `columns` / `rescue_shipped` corpus vocabulary — the closed `InducedRescue` outcome set is unchanged.
- Floor exemption is `FLOOR_MIN_CARDS` only, behind the dominance predicate; thresholds are named constants (0.8 share, 2 rows). Adversarial review (executed probes) forced both this narrowing and the two-column restriction; accepted-and-noted residuals: first-cell `rowspan` continuation rows drop (inherent to the columns rule, rare), and the 80% share is a hard boundary. Pre-existing, out of scope: the Downloads structure-rescued notice only fires for Notion jobs, not uploads.
- No Card Options toggle — trio-unanimous (undiscoverable at the moment of failure; not reachable on the MCP path at all).

## Testing

- Full server suite: 7090 passed, 2 failed = documented `getPageCount` pdfinfo environmental exception only.
- Web: vitest 3558 passed, typecheck clean, lint warnings pre-existing only.
- `pnpm format:check` clean; server `tsc -p .` clean; arch 0 errors.
- New: 8 DeckParser dominant-table cases (incl. incidental-table, cherry-pick, repeated-front floor, wide-matrix guards) + 8 `hasDominantTable` unit cases.

Browser check: not applicable — server-side parser and MCP schema change; the only web diff is the changelog JSON.

## Verification watch (post-deploy)

`conversion_rule_scores`: `columns rescue_shipped` share up, `unclassified shipped` (docx/html) down, other formats flat.

### DIFF
--- a/src/lib/parser/DeckParser.induction.test.ts
+++ b/src/lib/parser/DeckParser.induction.test.ts
@@ -388,4 +388,33 @@ describe('DeckParser dominant table documents', () => {
 
     expect(parser.inducedRule).toBeUndefined();
   });
+
+  it('fails honest on a dominant table with repeated fronts (floor guard)', () => {
+    const planner =
+      '<table><thead><tr><th>Day</th><th>Task</th></tr></thead><tbody>' +
+      '<tr><td>Monday</td><td>Study cardiology chapter one</td></tr>' +
+      '<tr><td>Monday</td><td>Review pharmacology flashcards</td></tr>' +
+      '<tr><td>Monday</td><td>Practice ECG interpretation</td></tr>' +
+      '<tr><td>Monday</td><td>Read renal physiology notes</td></tr>' +
+      '</tbody></table>';
+    const parser = parse(page(planner), new CardOption({}));
+
+    expect(parser.payload[0].cards).toHaveLength(0);
+    expect(parser.inducedRule).toMatchObject({ outcome: 'rescue_rejected' });
+  });
+
+  it('leaves a wide data matrix to the existing pipeline', () => {
+    const timetable =
+      '<table><thead><tr><th>Time</th><th>Mon</th><th>Tue</th><th>Wed</th><th>Thu</th></tr></thead><tbody>' +
+      '<tr><td>09:00</td><td>Maths</td><td>English</td><td>Biology</td><td>History</td></tr>' +
+      '<tr><td>10:00</td><td>Physics</td><td>Art</td><td>Maths</td><td>English</td></tr>' +
+      '</tbody></table>';
+    const parser = parse(page(timetable), new CardOption({}));
+
+    expect(parser.payload[0].cards).toHaveLength(0);
+    expect(
+      parser.inducedRule == null ||
+        parser.inducedRule.outcome === 'rescue_rejected'
+    ).toBe(true);
+  });
 });
--- a/src/lib/parser/DeckParser.induction.test.ts
+++ b/src/lib/parser/DeckParser.induction.test.ts
@@ -275,3 +275,117 @@ describe('DeckParser degenerate-yield rescue', () => {
     expect(fronts).toContain('Toggle question 1?');
   });
 });
+
+// The reported shape behind #4366: a document whose body is one two-column
+// Question/Answer table with a bullet list inside each answer cell. The row is
+// the card boundary the author already drew; the parser must never fan the
+// answer bullets out into per-bullet cards, whatever the row count.
+const qaTable = (rows: Array<[string, string[]]>): string => {
+  const body = rows
+    .map(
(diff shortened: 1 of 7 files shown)


aalemayhu (commented):
Ship review: security fork clean (no new regex/sink/exhaustion surface; MCP string static). Engineering fork round 1 flagged one blocker — the dominant-table branch in a separate zero-cards block let the disable-indented-bullets else-if append front-only list notes onto an induced table deck — fixed in 01643bcd by folding the branch into the single zero-cards block with a regression test; round 2 verified all four paths semantically identical to pre-PR outside the new branch. UX fork clean on changelog + MCP copy (two non-blocking nits noted). Adversarial pre-review earlier forced the non-count floor guard and the exactly-two-column predicate, both landed with red-green tests. Full server suite 7094 passed with only the documented pdfinfo environmental pair; web vitest 3558 passed; format/tsc/arch clean.

PR #4372, opened 2026-09-09
fix: group the return-rate cohort by output position

aalemayhu:
## What

`/api/ops/return-rate/metrics` has been returning an error payload on prod: `column "events.props" must appear in the GROUP BY clause or be used in an aggregate function`. The cohort query in `ReturnRateMetricsService.buildCohortQuery` bound `UNKNOWN_SOURCE` twice — once in the `SELECT` and once in `GROUP BY` — and Knex numbers each `?` independently (`$1` and `$4`), so Postgres saw two different expressions. This PR groups by output position (`group by 1, 2`) so the fallback binds once.

## Why

The ops owner hit the red banner on `/ops/growth` today. The bug dates from #4067 (2026-08-11), which rewrote this query; its SQL-shape test compared `toString()` output, where Knex inlines the literal on both sides and the text reads identically, so the test passed while the prepared statement failed. The new test reads `toSQL().toNative()` — the native SQL and the bindings array — and asserts three bindings with no placeholder inside `GROUP BY`.

Runtime entry point checked: `GET /api/ops/return-rate/metrics` → `GetReturnRateMetricsUseCase` → `ReturnRateMetricsService.getMetrics` → `buildCohortQuery`. Single construction site (`OpsRouter.ts`). Prod access log shows the route answering 200 with the error payload at 11:14 UTC today.

## Measuring success

None — internal ops tooling. The Return rate section on `/ops/growth` renders numbers instead of the banner after deploy.

## Testing

- `ReturnRateMetricsService.test.ts`: cohort SQL-shape expectation updated; new test asserts native SQL has `group by 1, 2`, no `$n` in the group clause, and exactly three bindings. 9/9 green. Server `tsc -p .` clean. `pnpm format:check` clean.

Browser check: not applicable — server-only query change, no rendered output.

## Changelog

No changelog entry — internal ops tooling, not user-visible.

## Sonar

## Risks

Low. Same result set, one fewer bind parameter. Rollback is a straight revert.

https://claude.ai/code/session_01KRx1TvKhs21ee4M6BcNtcd

### DIFF
--- a/src/services/ops/ReturnRateMetricsService.test.ts
+++ b/src/services/ops/ReturnRateMetricsService.test.ts
@@ -37,8 +37,15 @@ describe('ReturnRateMetricsService — generated SQL', () => {
         'MAX(created_at) as last_success_at from "events" ' +
         'where "name" = \'conversion_succeeded\' and "created_at" >= \'<CUTOFF>\' ' +
         'and COALESCE(user_id::text, anonymous_id) IS NOT NULL ' +
-        "group by COALESCE(user_id::text, anonymous_id), COALESCE(props->>'source', 'unknown')"
+        'group by 1, 2'
     );
+  });
+
+  it('groups the cohort by output position so the source expression binds once', () => {
+    const { sql, bindings } = service.buildCohortQuery(NOW).toSQL().toNative();
+    expect(sql).toContain('group by 1, 2');
+    expect(sql).not.toMatch(/group by .*\$/);
+    expect(bindings).toHaveLength(3);
   });
 
   it('builds the returns query as a single scan with a LEAD window, never a correlated subquery', () => {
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
**Verdict**: approve

Fixes the prod-visible return-rate error the ops owner reported today: the cohort query bound the unknown-source fallback in both SELECT and GROUP BY, Knex numbered them $1 and $4, and Postgres rejected them as different expressions. Grouping by output position binds once. Two-fork review (security / engineering): no user input reaches the raw fragment; ordinal grouping is valid Postgres for this select list; the LEAD/PARTITION BY in the sibling query is exempt from the grouping rule so it stays as is; the new native-SQL test guards against a placeholder re-entering the group clause. Server-only, out-clause attestation.

PR #4383, opened 2026-09-10
test: keep the answers slug lists in lockstep

aalemayhu:
Follow-up from https://github.com/2anki/server/pull/4381.

## What

`web/src/pages/AnswersPage/answersConfig.parity.test.ts` asserts, in both directions, that the three hand-maintained answers lists agree:

- `ANSWERS_PAGES` (the config map) ⇄ `ANSWERS_SLUGS` in `src/routes/knownRoutes.ts` (read as text, same pattern as `events.parity.test.ts`)
- `ANSWERS_PAGES` ⇄ `/answers/<slug>/` URLs in `web/public/sitemap.xml`

## Why

A slug missing from the server allowlist makes direct loads of the page return a 404 status to crawlers; a slug missing from the sitemap never gets discovered. Both lists are maintained by hand and nothing caught a miss until now. Verified the test fails when a sitemap entry is removed and passes on `main`.

## Checks

- Web: parity test 4/4; typecheck clean; lint clean; `pnpm format:check` clean.
- No changelog entry — test-only, no user-visible change.

Browser check: not applicable — test-only file under `web/src/`, no rendered output.

### DIFF
--- a/web/src/pages/AnswersPage/answersConfig.parity.test.ts
+++ b/web/src/pages/AnswersPage/answersConfig.parity.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { ANSWERS_PAGES } from './answersConfig';
+
+const REPO_ROOT = join(__dirname, '../../../..');
+
+function readServerAnswersSlugs(): Set<string> {
+  const source = readFileSync(
+    join(REPO_ROOT, 'src/routes/knownRoutes.ts'),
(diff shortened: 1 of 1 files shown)


aalemayhu (commented):
**Verdict**: approve

Test-only guard for the three hand-maintained answers lists, so a new SEO page cannot ship 404ing to crawlers or missing from the sitemap. Engineering fork: path depth verified, the allowlist regex throws when unmatched (no vacuous pass), sitemap regex matches all 17 existing entries, assertions are exact empty-array equality. Verified locally that removing a sitemap entry fails the test.

PR #4385, opened 2026-09-10
chore: drop the unused deck_shares public-listing columns

aalemayhu:
Closes #4349

**Hard rail: migration — waits for Alexander.** Please run `migration-reviewer` before merging.

## Summary

The public shared-deck library (#3827) was removed, but its schema stayed behind on `deck_shares`: `is_public`, `title`, `card_count`, and the partial index `deck_shares_public_listing_idx`. Nothing in `src/` or `web/src` reads or writes them (grep verified again today; only the kanel-generated `DeckShares.ts` mentions them). The trio recommendation on the issue was **drop**, so this PR drops them.

- `migrations/20261014000000_drop_public_listing_from_deck_shares.js` — drops the partial index first (`CONCURRENTLY`, so `transaction: false`, same as the migration that created it), then the three columns. `down` restores the columns with their original defaults and recreates the index.
- `src/data_layer/public/DeckShares.ts` — kanel regenerated against the applied migration (`pnpm dlx kanel`), only the file this migration changes is staged.
- `src/data_layer/dropPublicListingFromDeckSharesMigration.sql.test.ts` — DDL-shape test: index dropped before the columns it filters on, all three columns dropped, rollback restores defaults + index.

## Migration safety

- `DROP INDEX CONCURRENTLY` takes no exclusive lock on the table; `DROP COLUMN` on Postgres is a catalog change (no rewrite), so the `ACCESS EXCLUSIVE` lock is momentary. `deck_shares` is small.
- Not reversible in the data sense: any `is_public`/`title`/`card_count` values are lost. Prod never wrote them (no writer exists), so nothing to lose.
- No changelog entry — internal schema cleanup, no user-visible change.

## Test plan

- `pnpm test src/data_layer/ShareRepository.test.ts src/data_layer/dropPublicListingFromDeckSharesMigration.sql.test.ts src/usecases/share` — 4 suites, 11 tests pass
- `pnpm exec tsc -p . --noEmit` clean
- `pnpm format:check` clean
- Migration applied to the local Postgres and the resulting schema/index list verified before running kanel
- Sonar: not run locally; PR decoration will report

## Note on the pre-push hook

[...]

PR #4386, opened 2026-09-10
fix: stop the pre-push format hook rejecting kanel-only pushes

aalemayhu:
**Hard rail: `.claude/hooks/` — waits for Alexander.**

## Summary

`.claude/hooks/pre-push-format.py` refused the push for https://github.com/2anki/server/pull/4385 even though CI's `pnpm format:check` accepts it. The hook runs before the command, so with an uncommitted-only diff the formattable file list was just the kanel-regenerated `src/data_layer/public/DeckShares.ts`. That path is in `.oxfmtrc.json`'s `ignorePatterns`, and oxfmt exits **2** (not 0) with `Expected at least one target file. All matched files may have been excluded by ignore rules.` when every explicitly-passed file is ignored. The hook treated any non-zero exit as a format failure.

Fix: `run_oxfmt_check` returns clean when the output carries that sentence. A real format diff (exit 1 with offending files) still denies. Docstring updated to say so.

## Test plan

- `python3 .claude/hooks/pre-push-format.test.py` — 19 tests pass (2 new: all-files-ignored is clean; a real diff is still returned as offending output)
- Reproduced by hand on the #4385 branch: `pnpm exec oxfmt -c .oxfmtrc.json --threads 1 --check src/data_layer/public/DeckShares.ts` → exit 2 with that message; with a non-ignored file added to the list → exit 0
- No changelog entry — tooling only, no user-visible change

### DIFF
--- a/.claude/hooks/pre-push-format.py
+++ b/.claude/hooks/pre-push-format.py
@@ -23,7 +23,9 @@ Two gates:
 
 Skips pushes to main/master (safety.py owns those). oxfmt applies its own
 ignorePatterns from .oxfmtrc.json (templates, fixtures, generated, etc.), so we
-don't re-implement that filter here.
+don't re-implement that filter here — but when EVERY passed file is ignored,
+oxfmt exits 2 with "Expected at least one target file"; that is treated as
+clean, not as a format failure (kanel-regenerated data_layer/public files).
 
 Bypass: CLAUDE_SKIP_FORMAT_CHECK=1 git push ...   (env var, or inline in the
         command string — both honored, mirroring the typecheck hook).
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
**Verdict**: approve

Removes a false positive that blocked every kanel-regeneration push; keeps the gate for real format diffs. Security fork: no bypass semantics changed, argv-list subprocess unchanged; one low-confidence hardening (pin exit code 2 alongside the message) — applied. Engineering fork: probed oxfmt directly and confirmed a mixed list with a real failure exits 1 without the ignored-files sentence, so the new branch cannot mask a genuine failure; nits (realistic oxfmt output in the mock, reuse the existing mock import) — applied. 20 hook tests pass.

### Blocking
- none

### Nits
- none outstanding

Hard rail (.claude/hooks): waits for Alexander to merge from the UI.

PR #4388, opened 2026-09-10
feat: keep upload-deck card identity stable across re-uploads

### LINKED ISSUE #4346: Stable card identity for upload decks: re-converting an edited file must update, not duplicate
The last open thread from the 26-month inbox synthesis (the other four build-next items all shipped in the 2026-09-01/02 run).

## Problem
Re-converting an edited upload produces new cards instead of updating existing ones — Anki sees fresh GUIDs, the user gets duplicates and loses scheduling. Asked for in every corpus era; the Jul 2026 Notion export change turned it from a wish into breakage. Notion pages got block-id re-keying (#4255); upload-path decks (PDF/HTML/zip/AI) have nothing — #4340 dedups *suppressed* repeats but does not preserve identity across edits.

## Direction
Derive stable GUIDs for upload-path cards from content-anchored identity (e.g. source-position + normalized-front hashing with tolerance for edits) so a re-uploaded revision updates in place. Parser-core, high blast radius — spec first, adversarial review per the multi-agent policy.

### Metric
"duplicate cards after re-upload" support threads → 0; existing-deck update rate in analytics.

aalemayhu:
## What
Signed-in uploads of markdown, plain (non-Notion) HTML, CSV and xlsx now get a stable, content-derived card identity. Each card is keyed on `u:` + `guidFor(normalizedFront, cardType)` (deck name deliberately excluded), and the guid python honours unconditionally is computed in TS. Re-converting an edited file replays the guid it issued last time instead of adding a duplicate.

Closes #4346

## Why
The legacy content formula keyed on the deck/file name, so any rename forked the deck in Anki and orphaned review history. This makes re-uploading an edited file update the existing cards for logged-in users. Anonymous conversions, Notion HTML exports, PDF and AI decks are unchanged.

## How
- New `src/lib/anki/uploadCardIdentity.ts`: the `u:` identity key (front + card-type, locked vocabulary), the source-key hash, the pack/unpack of the `source_page_id` column, and the ambiguity guard.
- `DeckParser.applyUploadIdentityGuids` (sibling of `applyLedgerGuids`, run in `processPayload` and in `tryExperimental` so the CSV/FallbackParser path is covered) sets `card.guid` and `card.identityKey`, only for cards with `notionId == null`.
- **Ambiguity guard**: on a ledger hit, replay the stored guid when the back fingerprint matches (rename/reorder) OR the source-key matches (answer edit in the same file). When BOTH the filename and the answer changed, issue a fresh `guidFor(owner, identityKey, fingerprint)` — a visible duplicate beats silently overwriting an unrelated card that shares a question. Identical normalized fronts in one upload get `#2`/`#3` ordinals.
- Eligibility (markdown / HTML / CSV / xlsx, signed-in) is decided in `PrepareDeck` by only threading the `uploadIdentity` context to the parser; ineligible/anonymous conversions never load the ledger and are byte-identical.
- `CardGuidLedgerRepository.getUploadIdentityForOwner` reads the `u:` rows; entries are recorded first-sight-only through the existing insert-ignore `record` path.

## Measuring success
Server-side `upload_identity_replayed` event (added to `src/types/AnalyticsEvents.ts`) fires once per signed-in eligible upload with `{ replayed, issued, guarded }` summed across the upload's decks, read from `/api/ops/metrics`. `replayed > 0` climbing over time is the deduplication working; a rising `guarded` flags genuinely-ambiguous shared-question collisions.

## Testing
- Unit tests added:
  - `src/lib/anki/uploadCardIdentity.test.ts` — identity key, ordinals, source-key hashing, pack/unpack, and the guard's four branches (issue / replay-on-answer-edit / replay-on-rename / guard-when-both-changed).
  - `src/lib/parser/DeckParser.uploadIdentity.test.ts` — end-to-end through the parser: first-sight issues `guidFor(owner, identityKey)` and records it; answer edit and rename replay; both-changed forks; identical fronts get distinct guids; anonymous leaves guids to the content formula; a Notion HTML export stays on the block-id path (no identity applied).
  - `src/data_layer/CardGuidLedgerRepository.test.ts` — `getUploadIdentityForOwner` filters `u:` rows and returns guid + packed source.
  - `src/services/UploadService.test.ts` — the `upload_identity_replayed` event sums per-package counts and does not fire when no package ran the path.
- Full server suite: **7081 passed, 81 failed, 25 skipped, 7 todo** (all 81 failures are the documented worktree-environmental `PythonExitError` apkg-build suites + the `getPageCount` pdfinfo-missing test; zero assertion failures — verified 129 PythonExitError throws and 0 `Expected/Received` mismatches across the parser suites).
- Full web suite: **3616 passed** (403 files), web typecheck clean, web lint clean (pre-existing warnings only).
- `pnpm exec tsc -p . --noEmit` clean · `pnpm arch` 0 errors · `pnpm format:check` clean.

## Browser check
- [x] No runtime-visible web change beyond the What's New changelog entry — this is a server/parser change; there is no new UI, toggle, or notice.
Notes: The only web artifact is the changelog JSON; verified the web suite (which includes the changelog loader's id-uniqueness assertion and the analytics parity test) is green.

## Changelog
`Re-converting an edited markdown, HTML, CSV, or xlsx file updates your existing cards instead of adding duplicates when signed in`

Answers to the spec's three open questions plus every judgment call beyond the spec:

1. **Does the Notion HTML file-upload path already produce a ledger hit end to end?** Yes. Notion HTML exports set `card.notionId` via `resolveToggleBlockId`, flow through `applyLedgerGuids` / `collectIssuedGuids` (sidecar keyed by `notionId`), and already have end-to-end identity. The new path therefore gates on `notionId == null` and never touches Notion cards — locked in by the "keeps a Notion HTML export on the block-id path" test.
2. **Column fit for the back fingerprint in `card_guids`.** There is **no** `fingerprint` column and only one free nullable string column (`source_page_id`, varchar 255). It fits **without a migration** by packing `sha256(sourceKey):fingerprint` (129 chars) into `source_page_id`. I deviated from the spec's literal `source_page_id = sourceKey` on two points, both deliberate: (a) I store `sha256(sourceKey)` rather than the raw filename — the guard only needs equality, and raw filenames carry user PII (support-confidentiality), and the hash also guarantees the 255-char fit for any filename; (b) I co-locate the fingerprint in the same column with a `:` separator (colon can't appear in the two hex halves and, unlike the `` I first tried, is safe in a Postgres text column).
3. **Exact normalizedFront rules for CSV cells with markup.** By the time identity runs, CSV cells are already `Note`s; `normalizeCardText` (now exported from `ClaudeService`) strips all HTML/markup, lowercases, trims and collapses whitespace, so cell markup never reaches the key. CSV reaches identity through `tryExperimental` (FallbackParser), which bypasses `processPayload` — so I call `applyUploadIdentityGuids` there too.

Other judgment calls:
- **Eligible formats.** In-scope per the spec: markdown, plain HTML, CSV, xlsx. I excluded docx (not listed) and PDF/PPT/image/AI (non-deterministic fronts). Eligibility is decided by file type in `PrepareDeck`, and "plain vs Notion HTML" is handled by the card-level `notionId == null` gate.
- **Fresh guarded guid is deterministic and unrecorded** (`guidFor(owner, identityKey, fingerprint)`), so re-uploading the duplicate is stable while the original card keeps owning the base guid via its first-sight ledger row.
- **Ordinals are per-file (per DeckParser).** For a single-file upload that is the whole upload; for a zip, two separate files that share an identical normalized front are not cross-deduplicated in v1 (Anki dedupes identical notes on import anyway). Noted as a v1 limitation.
- **Metric noise gate.** The parser always holds a zeroed stats object; `PrepareDeck` only surfaces it when the eligible path actually ran, so a signed-in PDF or Notion-export upload does not emit an all-zero `upload_identity_replayed`.
- **Question-text edits are a documented limitation** (asserted): editing a front changes the identity key, so the edited card is a new card and the old guid is orphaned. Out of scope per the spec.

## Changes from the adversarial review round (2026-09-10)
Four reviewers (security, parser engineering, an execution-verified skeptic that ran 11 probes in the worktree, voice). One blocker and one serious concern, both fixed on the branch:

- **Blocker — stale guard state after a replay.** Rows were written on first sight only, so "edit an answer, then rename the file" (two ordinary re-imports) hit the guard against the original fingerprint and duplicated the card. Now replayed cards are written back too, and `UploadService` routes every `u:` entry through `CardGuidLedgerRepository.reissue` (upsert) while Notion entries keep the insert-ignore `record`. Test: "keeps the guid through an answer edit followed by a rename".
- **Concern — identical-front cards followed position, not content.** With ordinals assigned by document order, reordering three same-front/different-back cards replayed all three (same file) and re-attached each card's review history to a different fact. The guard is now resolved per front (`resolveUploadIdentityGroup`): a card claims the stored row whose fingerprint matches; a changed answer replays only when it is the sole card with that front, one row is stored, and the file matches; anything else landing on a stored row forks. Consequence (documented, asserted): editing one answer inside an identical-front group forks that card instead of guessing — a visible duplicate over a wrong overwrite, the designer's rule. Tests: "keeps identical-front cards attached to their own answers when reordered", "forks rather than guesses when one of several identical-front answers is edited".
- **Perf nit:** `getAllForOwner` (the Notion ledger read) now excludes `u:` rows, so a signed-in upload no longer loads the upload ledger twice. The upload read still loads the owner's whole `u:` ledger (owner-scoped, no per-key `IN`); bounded by the account's own history, left for a follow-up if a heavy uploader shows up in the metric.
- **Cleanup:** dropped the never-read `Note.identityKey`; added parser-level tests for reorder and question edit; changelog line shortened per the voice review (drops "xlsx", moves the signed-in qualifier).
- **Locked-vocabulary note:** the `#n` ordinal suffix is kept as specified, but `#` is inside the guid alphabet, so keys are never parsed back — the group resolver computes candidate keys forward (`keyForOrdinal`) instead.

## Risks
- Blast radius is parser core. Mitigations: the whole path is inert unless `uploadIdentity` is threaded (signed-in + eligible format), gated on `notionId == null`, and the anonymous/Notion/PDF paths are proven byte-identical by tests. A lost ledger row re-issues the same guid while front+type are unchanged, so a ledger read failure degrades to today's behavior, not worse.
- The `source_page_id` packing is additive; no schema change, no `card_guids` migration.
- Rollback: revert the PR — the ledger rows already written stay valid (they only replay when their key recurs), and no column was added.

## Goal alignment
Retention lever: re-converting an edited study file updating the existing deck (instead of duplicating it) removes a real reason paid learners abandon the round-trip. Funnel/revenue metric to watch: `upload_identity_replayed.replayed` climbing (dedup working) at `/api/ops/metrics`, read week-1 post-deploy. Signed-in only, so it targets the retained/paid cohort, not acquisition.


aalemayhu (commented):
**Verdict**: approve

Retention lever: a signed-in re-upload of an edited markdown/HTML/CSV file updates the existing Anki notes instead of duplicating them. Four reviewers on the draft. Security: ledger reads and writes owner-scoped from res.locals, filename only ever stored as a hash, anonymous path provably inert, event carries counts only. Execution-verified skeptic ran 11 probes: anonymous/Notion/PDF/AI paths byte-identical, cross-owner isolation, rename/answer-edit/both-changed guard branches, idempotent re-upload — all pass. Parser engineering found one blocker (replayed rows were never refreshed, so answer-edit-then-rename duplicated) and the skeptic one serious concern (identical-front cards followed position, so a reorder re-attached review history to the wrong answer). Both fixed in the review-round commit: replayed rows now write back through the upsert path, and the guard resolves per front by fingerprint, forking rather than guessing inside identical-front groups. Nits applied: Notion ledger read excludes upload rows, unread field dropped, missing parser-level tests added, changelog line trimmed. Full server suite 7091 passed with only the documented environmental failures; identity suites 35 passed.

### Blocking
- none outstanding

### Nits
- Upload ledger read is owner-scoped but not per-key; bounded by the account's own history, follow-up if the metric shows heavy uploaders.

aalemayhu (commented):
**Verdict**: approve (re-affirmed for the Sonar follow-up commit)

Only change since the reviewed head: the three SonarCloud findings on that head are cleared — an unused import, two replace→replaceAll swaps, and the eighth parameter on buildAllInOneSlot folded into one object. No behavior change; identity, parser, and zip suites pass (43 tests).

aalemayhu (commented):
**Verdict**: approve (re-affirmed for the node:crypto import fix)

Only change since the previous head: the last SonarCloud finding (S7772, prefer node:crypto) is cleared. No behavior change; identity suite passes.

PR #4389, opened 2026-09-10
fix: show one App Store badge on the app page

aalemayhu:
## Summary

On https://2anki.net/app/ both App Store badges (the black light-theme one and the white dark-theme one) rendered side by side in the light theme. Cause: `.badge img { display: block }` has higher specificity (0,1,1) than `.badgeDark { display: none }` (0,1,0), so the "hide the other variant" rule lost. Dark mode only worked because its `[data-theme='dark'] .badgeLight` selector happened to be more specific.

Fix: the shared `.badge img` rule only sizes; visibility lives on `.badge .badgeLight` / `.badge .badgeDark` with the theme overrides one attribute selector above them. `NativeAppPage.badge.test.ts` locks the specificity by reading the stylesheet (vitest does not apply CSS modules); it fails on `main`'s CSS and passes on this branch.

Simpler/more beautiful: one badge, matching the theme, on an acquisition page for the native app. Metric: none beyond the page looking right; `native_app_store_clicked` is unchanged.

## Test plan

- `vitest src/pages/NativeAppPage` — 3 files, 15 tests pass (new: 3 stylesheet assertions, verified red against `main`'s CSS)
- `pnpm format:check` clean

## Browser check

- [x] Golden path on localhost:3000 — loaded `/app/` at 375px on `pnpm dev`: light theme renders only `/badges/app-store.svg`, `data-theme="dark"` renders only `/badges/app-store-white.svg`, and switching back restores the light badge — in both CTA placements (hero and bottom).
- [x] No console errors at 375px from this change (the only error is the pre-existing signed-out 401 on `/api/users/me/preferences`).

### DIFF
--- a/web/src/pages/NativeAppPage/NativeAppPage.badge.test.ts
+++ b/web/src/pages/NativeAppPage/NativeAppPage.badge.test.ts
@@ -1,0 +1,39 @@
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+// The two App Store badges are theme variants; exactly one may render. Vitest
+// does not apply CSS modules, so this locks the stylesheet's selector
+// specificity instead: a bare `.badge img` rule that sets `display` outranks
+// `.badgeDark { display: none }` and shows both badges side by side.
+const css = readFileSync(
+  resolve(__dirname, 'NativeAppPage.module.css'),
+  'utf8'
+).replaceAll(/\/\*[\s\S]*?\*\//g, '');
+
+const ruleBodies = (selector: string): string[] =>
+  [...css.matchAll(/([^{}]+)\{([^}]*)\}/g)]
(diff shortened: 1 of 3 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler and more beautiful: the native-app page shows one App Store badge, matched to the theme, instead of both variants side by side. Security fork: static CSS and a read-only test on a fixed path, nothing else. Engineering fork: computed the specificities (shared rule (0,1,1) no longer sets display; visibility on (0,2,0) selectors; theme overrides at (0,3,0)), confirmed gold/purple/hotpink correctly keep the black badge on their near-white backgrounds, confirmed the inline-flex anchor blockifies the img so dropping display:block changes nothing, and confirmed the stylesheet test fails on main for the right reasons. Voice fork: changelog nits (lead on the noun, drop the apology clause) applied. Browser check at 375px in the PR body.

### Blocking
- none

### Nits
- none outstanding

PR #4394, opened 2026-09-11
feat: happy score in ops and one-tap reasons on deck feedback

aalemayhu:
## Summary

Deck ratings skewed negative and every low rating was free text: 29 of the last 45 deck-done ratings were low, and each one had to be read before it could be triaged. There was also no number that said whether finished decks made people happy over time.

This PR ships the measurement half of the Happy Score plan (the eligibility engine that decides *when* to ask is a follow-up):

- **One-tap reasons on the negative path.** After "Something was off", four chips: Images missing, Wrong cards, Too few cards, Something else. The pick is stored as a `[tag]` prefix on the comment; `CustomerSignalsService` surfaces it in the label (`Deck-ready rating 1 · images missing`) so Ops groups defects without reading prose. No new table, no migration.
- **`happy_score_ask_shown` event**, fired once per show, in both event allowlists. Without an "asked" denominator the response rate is unbackable.
- **Longer suppression**: dismiss = 90 days, answered = 180 days (was 14). Still localStorage; server-side per-user suppression is the follow-up.
- **Ops: Happy score section** on Business with 7d / 30d / 90d windows: positive ÷ (positive + negative) among answered ratings, hidden under 10 answers, plus asked count and response rate.
- **Today row `happy_score_90d_pct`**, lever health, target ≥ 60%, labeled "Happy score (users who rated a finished deck)".

Simpler/faster/more beautiful: one tap instead of a sentence when a deck is wrong. Scale: defects become triaged signals, which is what raises conversion success. Metric: 90d happy score on Today (target ≥ 60% by 2026-11-10); defect chips triaged ≥ 3/week.

## Decisions

- Score is share-of-positive, not mean rating (the widget only emits 1/2/5, a mean is meaningless).
- Headline window is 90d, not 30d: at n ≥ 10 the 30d row would grey out most weeks on today's volume.
- Label says who is being scored. PM flagged the gated ask as inflating the ratio by construction; the honesty lives in the label and the fixed gate, not in a hidden denominator.
- Testimonial comp on the positive path: cut (PM). One tap, done.
- Return-visit trigger: cut (Designer). The follow-up gates on the second successful uncapped download only.
- Prompt copy unchanged: the existing "Did this deck come out right?" / "Yes, it worked" / "Something was off" already matches the designer's two-button recommendation in all 10 locales; rewriting it would have cost 10 locale edits for no user-visible gain.

## Test plan

- `src/lib/happyScore.test.ts` (7): love/low counting, sample floor, response rate, rounding.
- `BusinessMetricsService.test.ts` (+2): three windows with asks from the events repo; null asks without a repo.
- `TodaySnapshotService.test.ts` (+2, row count 10→11): 90d row reads the score; unscored under the sample floor.
- `CustomerSignalsService.test.ts` (+1): reason tag in the label.
- `DeckFeedbackPrompt.test.tsx` (+6): ask event once, chip tags the comment, chip-only skip, no chips on positive path, 90/180-day suppression, `tagComment` trims and caps.
- Locale parity: `feedback.*` has 19 keys in all 10 `downloadsx.json` files.
- Full server suite 7192 pass (2 fail in `getPageCount`, the documented pdfinfo environmental failure). Full web suite 3624 pass. `tsc`, web typecheck, oxlint, `pnpm format:check`, `pnpm arch` clean.

## Browser check

- [x] Golden path on localhost:3000 at 375px: `/upload` → markdown file → "Convert anyway" → 2-card `.apkg` downloaded, both signed out and after registering a local account; `/downloads` renders for the signed-in user.
- [x] No console errors at 375px from this change (the only error is the pre-existing signed-out 401 on `/api/users/me/preferences`).
- Not rendered locally: the deck-done prompt itself. Local dev has no S3 credentials, so a signed-in deck never persists (`StorageHandler.uploadFile` fails in `persistSyncDeck`) and `/downloads` shows "No decks yet", which is the gate for the prompt. The chip row is covered by the 17 component tests in `DeckFeedbackPrompt.test.tsx`, and the UX reviewer checked the `flex-wrap` CSS at 375px.

### DIFF
--- a/src/lib/happyScore.test.ts
+++ b/src/lib/happyScore.test.ts
@@ -1,0 +1,91 @@
+import {
+  computeHappyScore,
+  HAPPY_SCORE_ASK_EVENT,
+  MIN_HAPPY_SCORE_SAMPLE,
+} from './happyScore';
+
+describe('computeHappyScore', () => {
+  it('counts love as rating 5 and low as ratings 1 and 2', () => {
+    const window = computeHappyScore(
+      '30d',
+      [
+        { rating: 1, count: 4 },
+        { rating: 2, count: 2 },
+        { rating: 5, count: 14 },
+      ],
+      null
+    );
+    expect(window).toEqual({
+      window: '30d',
+      love: 14,
+      low: 6,
+      n: 20,
+      score_pct: 70,
+      asks: null,
+      response_rate_pct: null,
+    });
+  });
+
+  it('hides the score below the minimum sample but keeps the counts', () => {
+    const window = computeHappyScore(
+      '7d',
+      [
+        { rating: 1, count: 3 },
+        { rating: 5, count: MIN_HAPPY_SCORE_SAMPLE - 4 },
+      ],
+      12
+    );
+    expect(window).toMatchObject({
+      love: MIN_HAPPY_SCORE_SAMPLE - 4,
+      low: 3,
+      n: MIN_HAPPY_SCORE_SAMPLE - 1,
+      score_pct: null,
+    });
+  });
+
+  it('ignores ratings outside the three widget values', () => {
+    const window = computeHappyScore(
+      '90d',
+      [
+        { rating: 3, count: 50 },
+        { rating: 4, count: 50 },
+        { rating: 5, count: 10 },
+      ],
+      null
+    );
+    expect(window).toMatchObject({ love: 10, low: 0, n: 10, score_pct: 100 });
+  });
+
+  it('derives the response rate from asks when they are known', () => {
+    const window = computeHappyScore(
+      '30d',
+      [
+        { rating: 1, count: 5 },
+        { rating: 5, count: 5 },
+      ],
+      200
+    );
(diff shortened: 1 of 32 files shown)


aalemayhu (commented):
Review fan-out (security, engineering, UX/voice) on the diff: clean, no blocking findings. Security: the reason tag is a fixed four-value client enum, the server still trims and caps the comment, Ops renders it as JSX text, the label regex is anchored and linear, the new events-repo wiring reaches only owner-gated ops routes. Engineering: skip-without-chip still sends no comment, the ask event fires once in production, the happy_score DB task nulls only its own key on failure, Today row assertions updated to 11, tests assert exact payloads. UX: chip labels present and in register in all 10 locales, changelog title in user voice, chips wrap at 375px; the two Ops-copy nits (plain-voice hint, Score header) were applied in the follow-up commit. Browser check attested in the PR body with the honest limitation that the prompt cannot render in local dev without S3.

Issue #4052, opened 2026-08-11
feat: accept folder uploads on /upload (macOS auto-unzip friction)


aalemayhu:
## User signal

A user on `/upload` (feedback in the ops voice-of-user inbox, in German) reports: macOS/Safari automatically unpacks downloaded zips, so to upload their material they have to re-compress the folder first instead of just uploading the folder.

This is default macOS Safari behavior ("Open safe files after downloading") — a meaningful slice of Mac users get a folder where our upload flow expects a zip.

## Proposal

Accept folders on `/upload`:

- Support folder drag-and-drop (`webkitdirectory` / DataTransfer items with `webkitGetAsEntry`) and zip the contents client-side (or upload files individually into one job — whichever fits the existing upload pipeline better).
- Copy nudge: if a drop contains a directory, don't reject it — handle it or explain in one sentence.

## Why it matters

Upload → download is the funnel's biggest drop-off; "my file won't upload" friction on the very first step is exactly the acquisition-lane leak. Notion exports (our highest-intent source) arrive as zips, i.e. as auto-unzipped **folders** on many Macs.

Funnel/metric to move: upload_started → conversion_succeeded rate on `/upload`.

PR #4396, opened 2026-09-11
fix: point the unpacked-export warning at the folder drop

### LINKED ISSUE #4395: Unbundled HTML warning should lead with the folder-drop fix
A paying subscriber (user 12791) wrote in today saying the site "doesn't really work anymore" for Notion HTML exports. Their prod rows show what happened: on 2026-09-09 they uploaded a bare `.html`, the unbundled-HTML warning fired twice, they continued, and the deck came out at 3–49 cards. The same page through Notion live convert the same day gave 50–499 cards. The converter is fine; the browser had unpacked Notion's export into a folder and the warning did not tell them what to do with it.

## The two warnings sell the wrong fix

`web/src/pages/UploadPage/components/UploadForm/hooks/useFileValidation.ts`:

- Multiple HTML files: "Re-download the zip from Notion in a different browser, or find the original zip in Downloads and upload that."
- Single HTML file: "If this came from Notion, download the zip export instead."

Folder upload shipped 2026-08 (#4052): dropping the unpacked export folder onto `/upload` repacks it with images. Neither warning mentions it, so the one-step fix is invisible at the exact moment the user needs it. Both strings are also hardcoded English while the rest of the upload page is localized in 10 locales; this user's locale is `de-DE`.

## Change

- Lead both warnings with the folder path: "Drop the whole export folder here instead, images included." Keep the zip advice as the second sentence.
- Add a third option for signed-in users with Notion linked: convert the page directly on `/notion`.
- Move the two strings into the `upload` namespace and translate in all 10 locales per `.claude/docs/i18n.md`.
- Keep the `unbundled_html` code and the `unbundled_html_warning_shown` event unchanged.

## Measure

`unbundled_html_warning_shown` (178 actors / 30d) against `folder_upload_packaged` (33 actors / 30d) before and 30 days after. Target: folder packaged ÷ warning shown doubles.

aalemayhu:
## Summary

Both unbundled-HTML upload warnings told users to re-download the zip in a different browser. Neither mentioned that dropping the unpacked export folder onto `/upload` repacks it with the images, which has worked since #4052. A paying subscriber followed the old advice this week, uploaded the bare HTML, got a thin deck, and wrote in to say the site had stopped working. The strings were also hardcoded English on a page that is otherwise localized in 10 locales.

Change:
- Multiple HTML files: "Drop the whole folder to include the images" plus a body that names the folder drop as the fix.
- Single HTML file: "Images won't be included" plus a body that names the folder drop or the zip.
- Both moved to `upload.validation.unbundledMany.*` / `unbundledSingle.*` in all 10 `common.json` files. `code: 'unbundled_html'` and the `unbundled_html_warning_shown` event are unchanged, so the before/after read in #4395 works.

Simpler: the warning now names the one-step fix at the moment it is needed. Metric: `folder_upload_packaged` ÷ `unbundled_html_warning_shown` (33 vs 178 actors / 30d today), read 30 days after merge; target doubles.

Closes #4395

## Decisions

- Designer cut the "or convert the page straight from Notion" sentence from both dialogs: the component cannot render a link, and unlinked prose pointing at another workflow is a dead end mid-flow. Promote the Notion-direct path elsewhere as a real link.
- PM's guard: the multi-file warning fires for Safari-unpacked exports, so the promised folder drop must work in Safari. `collectDroppedEntries` uses `webkitGetAsEntry`, supported in Safari since 11.1.
- No button, no browser-specific variant, no toggle (PM).

## Test plan

- `useFileValidation.test.ts`: the two html cases now assert the full translated result shape through the translator (previously asserted English substrings).
- `web/src/lib/i18n/locales/parity.test.ts` passes: identical key set across the 9 non-en locales, no placeholders.
- Web vitest for the hook and i18n: 640 tests pass. Web typecheck, oxlint, `pnpm format:check` clean.

## Browser check

Browser check: not applicable — the change is copy resolution through the existing `translate(t, key)` path already exercised by the markdown warnings; the dialog component, layout, and buttons are untouched. The real call site passes react-i18next's `t` (`UploadForm.tsx`), so no raw key can render.


aalemayhu (commented):
Review fan-out (security, engineering, UX/voice) on the diff: clean, no blocking findings. Security: only locale JSON, the validation hook, its test and a changelog changed; the strings render as JSX text with no markup or interpolation. Engineering: the real call site always passes react-i18next's t, so the raw-key fallback is test-only; the replaced tests assert the full result shape for both the single-file and multi-file branches; parity test holds across the 9 non-en locales. UX: register, protected names, and continue labels consistent in all 10 locales; the Portuguese article nit was applied in the follow-up commit. Trio decisions (cut the Notion-direct sentence, Safari folder-drop confirmed via webkitGetAsEntry, no button) are in the PR body.

aalemayhu (commented):
Same verdict for the head after the Portuguese article nit: clean, no blocking findings (security, engineering, UX/voice fan-out; details in the previous review comment).

PR #4411, opened 2026-09-12
fix: reject an Anki package uploaded as .zip before any conversion

### LINKED ISSUE #4409: fix: an apkg uploaded as .zip triggers N×N conversions and a 1 GB log burst
## What happened

On 2026-09-11 between 11:28 and 11:35 UTC the blue process wrote about 1 GB of out-log in six minutes (four rotated files of 149–372 MB) and ran roughly 800,000 `convertFile` calls for one upload, twice. The user got an empty deck both times. pm2-logrotate's 10 MB cap could not keep up. Disk was fine (40% used) but the same input at 5× the size would not be.

## Cause

The upload was an Anki package renamed or wrapped as `.zip`: an 11.6 MB archive whose entries are named `0` … `897` (apkg media entries are numbered) plus `collection.anki2` and `media`. Three checks combine badly:

- `isAnkiDeckFile` in `src/usecases/uploads/worker.ts` only matches the `.apkg` extension, so a `.zip` bypasses the "already an Anki deck" advice.
- `isZipContentFileSupported` (`src/usecases/uploads/isZipContentFileSupported.ts`) accepts any extensionless name as content, so all 898 numbered entries become deck candidates.
- `getRelevantFiles` (`src/usecases/uploads/getRelevantFiles.ts`) treats any extensionless root entry as an embeddable image because `isImageFileEmbedable` is extension-agnostic, so each of the 898 candidates gets all 898 entries as its file list.

Result: 897 `PrepareDeck` runs × 898 files, every one converting nothing (`mimetype: '0'`), and `[PrepareDeck] received` / `start` logging the full 898-name array on every run.

Only this one day shows the pattern in the retained logs, so it is a bad-input shape, not a regression.

## Fix

1. Sniff the archive: if a zip contains `collection.anki2`, `collection.anki21` or `collection.anki21b`, treat it as an Anki deck and return the existing advice (or route to the apkg-import surface). Cheap and stops this class entirely.
2. Stop extensionless root entries counting as images in `getRelevantFiles`; only files that pass a real image check belong there.
3. Bound the two PrepareDeck log lines to a count plus the first few names. Logging the whole file list per run is the amplifier that turned a slow request into a 1 GB write.
4. Optional: cap the candidate × relevant-file product per upload and fail with a clear message above it.

Test: fixture zip with `collection.anki2`, `media`, and 20 numbered entries; expect the deck-file advice, zero `PrepareDeck` runs.

## Evidence

pm2 out logs on the box, 2026-09-11 11:28–11:35 UTC, aggregated by line signature. No identifiers recorded here.

aalemayhu:
## What

A zip that is really an Anki package (root or folder-wrapped `collection.anki2` / `anki21` / `anki21b`) is rejected before any conversion with the "already an Anki deck" message the upload form already renders as its dedicated block. The async failure-reason resolver passes that message through so job-mode uploads see it too. Root-level zip siblings now count as images only by extension, so an extensionless entry can never fan out as a "relevant" file.

## Why

Closes #4409. On 2026-09-11 one 11 MB upload of this shape ran about 800,000 no-op conversions twice, wrote 1 GB of log in six minutes, and gave the user an empty deck. Simpler: the user gets a real answer instead of an empty deck. Metric: `conversion_failed{reason: empty_deck, input_format: zip}` in the ops funnel, read at the next weekly check.

- Reject vs auto-route to apkg import: chose reject (pm). The phrase "already an Anki deck" already fires the upload form's dedicated block with the apkg pointers, so no new surface; auto-routing would guess intent. Assumption: a double-wrapped apkg is as often accidental as deliberate. If wrong, route to the apkg-import surface in `getPackagesFromZip` instead of throwing.
- Copy: "This zip contains an Anki package, so it's already an Anki deck. 2anki converts source files like Notion HTML exports, not existing decks." (designer). Second sentence is verbatim from the `.apkg` rejection.
- Scope: detection at any depth (engineer), root-sibling image allowlist widened to webp/avif so flat Notion exports keep those images. Deliberately did NOT change `isZipContentFileSupported`'s extensionless rule (a `README`-style entry is still a candidate) nor add a global candidate cap.

## Tests

- `isAnkiPackageZip.test.ts` (new), `getPackagesFromZip.test.ts` (rejects before PrepareDeck), `getRelevantFiles.test.ts` (extensionless root entries excluded, webp kept), `jobFailureReason.test.ts` (message pass-through). All three behaviour tests fail on `main` and pass here.
- Full server suite: 732 suites passed, 7192 tests; the only failure is the documented `getPageCount` pdfinfo environment case. Web suite: 404 files, 3624 tests green. `pnpm format:check` and `tsc -p . --noEmit` clean.

Browser check: not applicable — server-side rejection plus a changelog JSON; the upload form block it triggers already exists.

### DIFF
--- a/src/usecases/jobs/jobFailureReason.test.ts
+++ b/src/usecases/jobs/jobFailureReason.test.ts
@@ -452,3 +452,13 @@ describe('jobFailureReasonFromError on worker-flattened errors', () => {
     expect(reason).toMatch(/Job ID job-2/);
   });
 });
+
+describe('jobFailureReasonFromError — Anki deck uploads', () => {
+  it('passes the already-an-Anki-deck message through to the user', () => {
+    const message =
+      "This zip contains an Anki package, so it's already an Anki deck. 2anki converts source files like Notion HTML exports, not existing decks.";
+    expect(jobFailureReasonFromError(new Error(message), 'job-1')).toBe(
+      message
+    );
+  });
+});
--- a/src/usecases/jobs/jobFailureReason.ts
+++ b/src/usecases/jobs/jobFailureReason.ts
@@ -187,6 +187,9 @@ export function jobFailureReasonFromError(
   ) {
     return (error as Error).message;
   }
+  if (error instanceof Error && /already an Anki deck/.test(error.message)) {
+    return error.message;
+  }
   if (
     error instanceof EmptyContentError ||
     hasName(error, 'EmptyContentError')
--- a/src/usecases/uploads/getPackagesFromZip.test.ts
+++ b/src/usecases/uploads/getPackagesFromZip.test.ts
@@ -763,3 +763,29 @@ describe('getPackagesFromZip — encrypted PDFs', () => {
     );
   });
 });
+
+describe('getPackagesFromZip — Anki package renamed as zip', () => {
+  it('rejects the upload before any conversion runs', async () => {
+    const fileNames = ['collection.anki2', 'media', '0', '1', '2'];
+    mockZipHandlerClass.mockImplementation(() => ({
+      build: jest.fn().mockResolvedValue(undefined),
+      getFileNames: jest.fn().mockReturnValue(fileNames),
+      files: fileNames.map((name) => ({ name, contents: 'x' })),
+    }));
+
+    const settings = new CardOption({});
+    const workspace = { location: FAKE_WORKSPACE_LOCATION } as Workspace;
+
+    await expect(
+      getPackagesFromZip(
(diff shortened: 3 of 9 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler: a renamed .apkg gets the existing deck-file answer instead of an empty deck and a runaway conversion loop. Security, engineering, and voice forks all clean.

### Blocking
- none

### Nits
- getPackagesFromZip.test.ts rides the file's existing internal mocks (ZipHandler, PrepareDeck); pre-existing pattern, a real-zip fixture would be the outside-in version.

PR #4412, opened 2026-09-12
fix: convert .tsv uploads as tab-separated CSV instead of an empty deck

### LINKED ISSUE #4399: fix: .tsv uploads always produce an empty deck
## Symptom

Every `.tsv` upload to the one-shot converter ends in `conversion_failed` with reason `empty_deck`. Prod events, 30 days to 2026-09-11: 26 TSV attempts, 0 successes. CSV in the same window: 271 successes, 0 empty decks.

## Cause

The string `tsv` does not appear anywhere under `src/`. The upload path has no route for the extension, so the file falls through to the generic handler and yields no cards. #1153 (closed) asked for TSV support in 2022; the CSV path shipped, the TSV alias never did.

## Fix

Treat `.tsv` as a tab-delimited CSV in the upload validation and in the fallback parser routing (`src/lib/parser/experimental/FallbackParser.ts`, `src/lib/csv/`). Anki's own CSV importer sniffs the delimiter, so 2anki should too, or force tab for the `.tsv` extension. Add a fixture test with a two-column tab file. Accept `.tsv` in the client file picker if it is filtered there.

## Evidence

Aggregate query on `events` for `conversion_succeeded` / `conversion_failed` grouped by `props->>'input_format'`, 2026-09-11.

aalemayhu:
## What

`.tsv` uploads convert through the existing CSV reader (which already sniffs the delimiter), and the upload file picker accepts `.tsv`. One regex in `isCSVFile`, one entry in the accept list, a changelog entry, and tests.

## Why

Closes #4399. Prod events, 30 days to 2026-09-11: 26 `.tsv` attempts, 0 successes, all `empty_deck`; `.csv` in the same window had 271 successes. Simpler: a common export format stops dead-ending. Metric: `conversion_failed{input_format: tsv}` in the ops funnel at the next weekly check.

## Tests

- `FallbackParser.test.ts`: a tab-separated `.tsv` with a header yields two cards with the right front/back (fails on `main`, passes here).
- `checks.test.ts`: `isCSVFile` accepts `.csv`/`.tsv` any case, rejects `.txt` and bare names.
- Full server suite green apart from the documented `getPageCount` pdfinfo environment case. Server `tsc`, tree-wide `format:check`, web typecheck and lint clean (two pre-existing lint warnings in MindmapEditor untouched).

## Browser check
Notes: `pnpm --filter 2anki-web test:golden-path` 4/4 green on this branch (the machine-backed form of the attestation). The only web change is one extension added to the file picker's accept list.

### DIFF
--- a/src/lib/parser/experimental/FallbackParser.test.ts
+++ b/src/lib/parser/experimental/FallbackParser.test.ts
@@ -356,3 +356,19 @@ describe('FallbackParser deck naming', () => {
     expect(decks[0].name).toBe('vocab');
   });
 });
+
+describe('FallbackParser TSV uploads', () => {
+  it('reads a tab-separated .tsv file as cards', () => {
+    const tsv = 'front\tback\nHola\tHello\nAdiós\tGoodbye';
+    const parser = new FallbackParser([
+      { name: 'vocab.tsv', contents: Buffer.from(tsv) },
+    ]);
+    const decks = parser.run({} as any);
+    expect(decks).toHaveLength(1);
+    expect(decks[0].cards).toHaveLength(2);
+    expect(decks[0].cards[0].name).toBe('Hola');
+    expect(decks[0].cards[0].back).toBe('Hello');
+    expect(decks[0].cards[1].name).toBe('Adiós');
+    expect(decks[0].cards[1].back).toBe('Goodbye');
+  });
+});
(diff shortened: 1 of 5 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler: tab-separated exports stop dead-ending as empty decks; one regex plus the picker accept entry, delimiter sniffing already existed. Security, engineering, and voice forks clean.

### Blocking
- none

### Nits
- FallbackParser.test.ts uses `{} as any` like the 28 existing cases in the file; `new CardOption({})` would be the tidy form.

PR #4414, opened 2026-09-12
fix: keep raw <img> tags in markdown as images instead of escaped text

### LINKED ISSUE #4403: fix: inline base64 image markup arrives HTML-escaped in recent decks
## Symptom

Some recent decks carry an inline image as HTML-escaped text: the Back field contains `&lt;img src=&quot;data:image/png;base64,…&quot;&gt;` as literal characters. Anki shows a wall of base64 instead of the picture. Seen in 4 of 60, then 1 of 60, page conversions from August to September 2026 in a prod sample on 2026-09-11. 0 of 58 decks from April to July. Payloads are mostly under 10 KB, PNG and JPEG.

## What is known

- The docx path is not the source: `convertDocxToHTML.ts` writes images to files via mammoth's `imgElement` converter.
- `src/lib/markdown.ts` builds markdown-it with `html: false`, which escapes any raw `` in markdown or text input. A document with an inline base64 image in raw HTML would produce exactly this shape.
- Anki never unescapes fields, and its media check only extracts data URIs from real `` attributes (`rslib/src/media/check.rs:534`).

## Fix

1. Reproduce: markdown input containing ``, run the upload path, inspect the field.
2. Either allow the image tag through and convert the data URI to a media file on the embed path (`src/lib/parser/exporters/embedFile.ts` currently rejects `data:image`), or strip it. Rendering it as text is the one wrong answer.
3. Add the fixture to the golden set.

aalemayhu:
## What

`markdownToHTML` and `markdownToInlineHTML` restore an escaped `` after rendering, rebuilt from `src` and `alt` only. Allowed sources: `data:image/…`, `http(s)://`, and scheme-less relative paths. Anything else (a `javascript:` source, an image without `src`) stays escaped exactly as before, and attributes such as `onerror` never come back.

## Why

Closes #4403. A prod sample on 2026-09-11 found recent page decks whose Back field held `&lt;img src=&quot;data:image/png;base64,…` as literal text, so the card showed base64 instead of the picture. Cause: markdown-it runs with `html: false`, and the existing card-safe-tag restore only covers bare formatting tags. More beautiful: the picture the user put in the document is the picture on the card. Metric: none beyond the golden fixtures; the shape is not instrumented.

## Tests

- `markdown.test.ts`: five new cases (base64 image restored, relative path restored, handlers dropped, script scheme stays escaped, inline render). Four fail on `main`, pass here.
- Full server suite green apart from the documented `getPageCount` pdfinfo environment case. `tsc`, format, lint clean.

Browser check: not applicable — server-side markdown rendering plus a changelog JSON, no web code change.

### DIFF
--- a/src/lib/markdown.test.ts
+++ b/src/lib/markdown.test.ts
@@ -225,4 +225,37 @@ describe('markdownToHTML', () => {
       expect(result).toContain('&lt;blockquote&gt;');
     });
   });
+
+  describe('raw <img> tags in markdown (regression: #4403)', () => {
+    it('restores an inline base64 image instead of escaping it into text', () => {
+      const src = 'data:image/png;base64,iVBORw0KGgo=';
+      const result = markdownToHTML(`Look: <img src="${src}" alt="cell">`);
+      expect(result).toContain(`<img src="${src}" alt="cell">`);
+      expect(result).not.toContain('&lt;img');
+    });
+
+    it('restores a relative image path so the zip embed step can pick it up', () => {
+      const result = markdownToHTML("<img src='images/heart.png'>");
+      expect(result).toContain('<img src="images/heart.png">');
+    });
+
+    it('keeps only src and alt, dropping event handler attributes', () => {
+      const result = markdownToHTML(
+        '<img src="x.png" onerror="alert(1)" alt="a">'
+      );
+      expect(result).toContain('<img src="x.png" alt="a">');
+      expect(result).not.toContain('onerror');
+    });
+
+    it('leaves an image with a script scheme escaped', () => {
+      const result = markdownToHTML('<img src="javascript:alert(1)">');
+      expect(result).not.toContain('<img');
+      expect(result).toContain('&lt;img');
+    });
+
+    it('restores images in the inline (front) render too', () => {
+      const result = markdownToInlineHTML('<img src="a.png">');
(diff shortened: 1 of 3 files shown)


aalemayhu (commented):
**Verdict**: approve

More beautiful: the image a user put in a markdown document is the image on the card instead of a wall of base64. Security fork executed bypass payloads (quote injection, javascript: and data:text/html schemes, entity-encoded colons, 60k-char backtracking probe): all rejected or linear. Engineering and voice forks clean.

### Blocking
- none

### Nits
- protocol-relative `//host/x.png` passes the scheme-less branch of the allowlist; harmless in Anki (remote image), noted for a follow-up if it ever matters.

aalemayhu (commented):
**Verdict**: approve

Round two after the SonarCloud S7781 fix: replaceAll with fixed strings, behaviour identical. Security fork re-ran the quote-injection and double-decode payloads against the new head; engineering confirmed tests still cover the decode path.

### Blocking
- none

### Nits
- none

aalemayhu (commented):
**Verdict**: approve

Round three after SonarCloud S7780: the two attribute regexes are hoisted constants, matching identical (word boundary, case-insensitive, either quote), no global flag, quote-injection payload still lands inside the attribute, 33/33 tests on the head.

### Blocking
- none

### Nits
- none

PR #4415, opened 2026-09-12
fix: warn when basic-mode cards carry cloze syntax

### LINKED ISSUE #4402: fix: cloze syntax in non-cloze conversions ships as literal text
## Symptom

Notes converted in basic mode sometimes contain literal Anki cloze syntax (`{{c1::…}}`) in a field. Anki renders it as raw text. In a prod sample of 185 generated decks (11,363 notes) on 2026-09-11: 14 such notes across 4 files, 13 of them in the Back field, none inside `<code>`.

## Cause

Users type cloze syntax directly in Notion or a document, but the conversion runs with cloze off. The converter never inspects content for existing markers, so the syntax survives as text on a basic note.

## Fix

When a card is not in cloze mode but a field matches `{{c\d+::.+?}}`, either emit it as a cloze note (Anki treats any field marker on a cloze notetype as a card) or surface it in the conversion report as "cloze syntax found, cloze mode was off". Emitting as cloze is closer to the user's intent. Test: basic-mode fixture with a marker in the back, expect a cloze note.

Note: Anki's cloze engine treats `::` inside a cloze body as a hint delimiter (`rslib/src/cloze.rs:199`), so content with `::` needs care when auto-wrapping.

aalemayhu:
## What

When cloze mode is off and a card's front or back contains `{{cN::…}}`, DeckParser counts it. PrepareDeck emits one coded warning per deck (`stray-cloze:<n>`, outranked by the existing markdown-heuristic warning), and UploadService turns the summed count into the `X-Warning` text the upload form already shows. Cards stay basic.

## Why

Closes #4402. Prod sample on 2026-09-11: 14 basic-mode notes across 4 decks carried literal cloze syntax, 13 in the Back field. Simpler: the user learns why the braces show and what to flip, instead of discovering it in Anki. Metric: none instrumented; the report can grow an event once the notice has a click.

- Warn, do not auto-convert (pm). 13 of 14 markers sit in the answer; a cloze note's face is the Text field, so auto-converting would promote the answer to the front and demote the question. Assumption: users want their question kept as the front. If wrong, the flip lives at `DeckParser.transformCard` next to the existing back-to-name cloze move.
- Copy (designer): "N cards contain cloze syntax like {{c1::text}} but cloze mode is off, so the braces show on the card. Turn on cloze mode and convert again." Singular form for one card. Swap if you would word it differently.
- Scope: coded warning through the existing single-warning channel; the markdown-heuristic warning keeps priority because it questions the whole deck. Deliberately did NOT build a per-card list, an auto-convert path, or a new header.

## Review round 2

The security fork caught that a Unicode ellipsis in the header value would make Node throw `ERR_INVALID_CHAR` on every upload that tripped the warning. The example is now plain ASCII, verified with `ServerResponse.setHeader`. The marker regex also dropped its lazy any-character body for a no-brace body, so an unterminated marker stays linear.

## Tests

- `hasClozeMarkup.test.ts` (new helper), `DeckParser.strayCloze.test.ts` (basic card with a marker in the answer is counted and stays basic; cloze mode on or no marker counts zero; fails on `main`), `PrepareDeck.test.ts` (`parserWarning` priority), `UploadService.test.ts` (`resolveUploadWarning` copy, summing, singular, priority).

Browser check: not applicable — server-side warning text through the existing `X-Warning` header the form already renders, plus a changelog JSON.

### DIFF
--- a/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
+++ b/src/infrastracture/adapters/fileConversion/PrepareDeck.test.ts
@@ -1,4 +1,4 @@
-import { PrepareDeck, prepareDeckInfoOnly } from './PrepareDeck';
+import { PrepareDeck, parserWarning, prepareDeckInfoOnly } from './PrepareDeck';
 import CardOption from '../../../lib/parser/Settings/CardOption';
 
 jest.mock('../../../lib/claude/ClaudeService', () => {
@@ -998,5 +998,25 @@ describe('assembleParserFiles — both build paths share one file set', () => {
     const all = assembleParserFiles([original], []);
 
     expect(all).toEqual([original]);
+  });
+});
+
+describe('parserWarning', () => {
+  it('reports stray cloze markup as a coded count', () => {
+    expect(parserWarning({ usedHeuristic: false, strayClozeCount: 3 })).toBe(
+      'stray-cloze:3'
+    );
+  });
+
+  it('lets the markdown heuristic outrank stray cloze markup', () => {
+    expect(parserWarning({ usedHeuristic: true, strayClozeCount: 3 })).toBe(
+      'markdown-heuristic'
+    );
+  });
+
+  it('is silent when nothing happened', () => {
+    expect(
+      parserWarning({ usedHeuristic: false, strayClozeCount: 0 })
+    ).toBeUndefined();
   });
 });
--- a/src/infrastracture/adapters/fileConversion/PrepareDeck.ts
+++ b/src/infrastracture/adapters/fileConversion/PrepareDeck.ts
@@ -172,6 +172,19 @@ function shippedInducedRule(
   return induced;
 }
 
+// One coded warning per deck; UploadService turns the code into copy. The
+// markdown heuristic outranks stray cloze markup because it questions the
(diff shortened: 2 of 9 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler: users who typed cloze syntax with cloze mode off get told why the braces show and what to flip. Three review rounds: round one caught a Unicode header value that would have thrown at response time, round two the quadratic marker scan; the head is ASCII-safe and linear (1 ms on 300 KB adversarial input). Engineering and voice forks clean.

### Blocking
- none

### Nits
- none

aalemayhu (commented):
**Verdict**: approve

Round four after SonarCloud S3776: the stray-cloze check is its own method, same guard and count, same call point; stray-cloze and helper tests 10/10 on the head.

### Blocking
- none

### Nits
- none

aalemayhu (commented):
**Verdict**: approve

Rebased onto main after #4416; the only conflict was two appended test blocks, kept both. Re-review confirmed the diff is stray-cloze only, #4416's guard intact, 105/105 on the head.

### Blocking
- none

### Nits
- none

PR #4416, opened 2026-09-12
fix: answer photo drops without a vision path before parsing

### LINKED ISSUE #4400: fix: images dropped on the converter fail as empty deck instead of routing to photo-to-deck
## Symptom

Users drop a `.jpg` or `.png` onto the main converter and get an empty-deck failure. Prod events, 30 days to 2026-09-11: 42 image uploads failed with `empty_deck`, 2 succeeded. That is a 93% failure rate on a format users clearly expect to work, because the site advertises photo-to-deck.

## Cause

The one-shot upload path has no image handler. Photo-to-deck is a separate surface, and nothing routes an image dropped on the upload zone to it.

## Fix

Pick one:
- Route image uploads on the main converter into the photo-to-deck use case and return its deck.
- Or reject images at validation with a message that links to the photo-to-deck page.

The first is simpler for the user and keeps the "drop something in, get a deck back" promise. Either way the `empty_deck` failure for images should disappear from the funnel.

## Evidence

`conversion_failed` grouped by `input_format` and `reason`, 30 days to 2026-09-11.

aalemayhu:
## What

`handleUpload` answers any image-only drop that the signed-in single-photo vision path does not take (anonymous users, or more than one photo) with the existing `image_only_no_text` response before the parser runs. The empty-deck catch that used to produce that response now calls the same helper.

## Why

Closes #4400. Prod events, 30 days to 2026-09-11: 42 jpg/png drops on the converter recorded `conversion_failed{reason: empty_deck}`. The users did get the Photo to Deck pointer, but only after a full parse, and the funnel counted each as a failed conversion. Faster: the answer arrives without a parse. Metric: `conversion_failed{input_format: jpg|png}` should drop to zero while `image_only_no_text_shown` carries the count, read at the next weekly check.

- Reject with the existing pointer, do not run the vision pipeline for anonymous users (pm). Photo to Deck meters cost by owner; an anonymous AI path would need a new spend cap, which is quota code and off limits unattended. Assumption: a free sign-in is acceptable friction for a photo. If wrong, the anonymous vision route is scoped in the engineer's notes on the issue (pass the anonymous id as owner, drop the `owner != null` gate).
- Copy: kept the shipped `IMAGE_ONLY_NO_TEXT_MESSAGE` and the client block it already triggers. Designer offered a sign-in-first variant ("Sign in to turn it into cards with Photo to Deck. It's free."); not adopted because the client renders this response by code with its own copy and a change would need the ten-locale pass. Swap if you want the sign-in nudge.
- Scope: one early return plus a shared helper. Deliberately did NOT add an anonymous vision path, a multi-photo batch, or new copy.

## Tests

- `UploadService.test.ts`: anonymous single photo and signed-in two-photo drops return `image_only_no_text`, never call the package use case, fire `image_only_no_text_shown`, and never fire `conversion_failed`. Both fail on `main`.
- Existing image-only test (0-card parse path) still passes through the shared helper.
- Full server suite green apart from the documented `getPageCount` pdfinfo environment case; `tsc`, format clean; lint warning count in `UploadService.ts` unchanged at the pre-existing 17.

No changelog entry: the response and the client block are unchanged; only the timing and the funnel label move.

### DIFF
--- a/src/services/UploadService.test.ts
+++ b/src/services/UploadService.test.ts
@@ -3971,3 +3971,75 @@ describe('UploadService.handleSyncUpload — persisted copy for a signed-in owne
     );
   });
 });
+
+describe('UploadService.handleUpload — image drops without a vision path (#4400)', () => {
+  const imageFile = (name: string) =>
+    ({
+      originalname: name,
+      mimetype: 'image/png',
+      size: 2048,
+      path: `/tmp/${name}`,
+    }) as unknown as Express.Multer.File;
+
+  it('answers an anonymous photo drop with the Photo to Deck pointer before any parse', async () => {
+    const execute = jest.fn().mockResolvedValue({ packages: [] });
+    MockGeneratePackagesUseCase.mockImplementation(
+      () =>
+        ({ execute }) as unknown as InstanceType<typeof GeneratePackagesUseCase>
+    );
+    const service = new UploadService(
+      buildRepository(),
+      {} as JobRepository,
+      buildUsersRepo(),
+      ...fakeUploadServiceDeps()
+    );
+    const req = buildRequest({
+      files: [imageFile('lecture-page.png')],
+      cookies: { anon_id: 'anon-photo' },
+    } as Partial<express.Request>);
+    const { res, capturedStatus, capturedJson } = buildResponse();
+
+    await service.handleUpload(req, res);
+
+    expect(capturedStatus()).toBe(400);
+    expect((capturedJson() as { code: string }).code).toBe(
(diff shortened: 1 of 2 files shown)


aalemayhu (commented):
**Verdict**: approve

Faster and honest funnel: photo drops with no vision path get the existing Photo to Deck pointer before any parse, and stop counting as failed conversions. Security confirmed the early return sits after upload validation and skips no cap; engineering confirmed both tests fail without the guard. Round two only relocated the helper under its own comment.

### Blocking
- none

### Nits
- none

PR #4417, opened 2026-09-12
feat: return a next step when an MCP deck hits the card cap

### LINKED ISSUE #4393: MCP: return the pass link when a tool result hits the card cap
Agent-shippable. The assistant channel converts poorly because the cap is invisible from inside the assistant: of 184 accounts created via `/mcp` in the last 90 days, 5 subscribed and 4 bought a pass, while only 16 ever opened the web app where the pricing page lives. When `create_deck` or `convert_to_deck` hits the anonymous or monthly cap, the tool result should carry the pass link, so the assistant can relay it in the same turn.

## Change

- When the MCP tool result is truncated by `anonymous_cap` or `monthly_limit`, include a short `next_step` field in the structured result and one sentence in the text: how many cards were held back and the direct checkout link for the Day Pass (`https://2anki.net/pricing`, or the pass checkout route if a signed-in deep link exists).
- Fire `paywall_shown` with `surface: 'mcp'` so the channel's paywall funnel is readable next to the web ones.
- Do not change the caps themselves.

## Measure

`paywall_shown{surface:mcp}` → `checkout_completed` for `/mcp`-origin accounts, read 30 days after merge against the baseline above (9 paid of 184).

Pairs with the three listing issues; listings without this are traffic with no checkout.

aalemayhu:
## What

When `create_deck` or `convert_to_deck` is blocked by the monthly card cap, the error result carries `next_step: { upgrade_url, cards_held_back? }` in `structuredContent`, and the message gains one sentence with the held-back count when it is known (the subdeck path and `create_deck` know it; the `convert_to_deck` upload path blocks before counting). The same branch fires `paywall_shown{surface: 'mcp', kind: 'card_count'}` once per blocked call. Caps unchanged.

## Why

Closes #4393. The assistant channel made 184 accounts in 90 days and 9 paid; only 16 ever reached the web app where the cap is explained. Simpler: the way past the cap is in the same turn. Metric: `paywall_shown{surface: mcp}` → `checkout_completed` for `/mcp`-origin accounts, read 30 days after merge against the 9-of-184 baseline in the issue.

- The issue assumed MCP truncates a deck; it does not, the tool refuses the whole deck at the cap (pm and engineer both confirmed). So `cards_held_back` is the candidate count that was refused, and the sentence reads "N cards were held back — none were created." Assumption: that phrasing is honest for a refusal. If you would rather MCP deliver a partial deck under the cap, that is a cap-semantics change for its own spec.
- Link: the pricing page with `?from=mcp`, already in the message. No Day Pass checkout deep link: it would need a Stripe session or a new checkout route, which is payment code. Swap once a no-auth pass link exists.
- Event fires from the tools service, not the router, so the subdeck branch (which never reached the upload service's `paywall_shown`) is covered too. User id resolves through the same owner-string rule Photo to Deck uses; a non-numeric owner records null.
- Scope: deliberately did NOT add an anonymous-cap branch (MCP is OAuth-only), partial delivery, or a typed props union.

## Tests

- `McpToolsService.test.ts`: the upload-redirect block returns `next_step` with the link and fires one `paywall_shown{surface: mcp}`; the subdeck block returns the count, the held-back sentence, and the event. Both cases were extended from the existing monthly-limit tests.
- `McpServerFactory.test.ts` unchanged and green; `next_step` passes through `errorResult` into `structuredContent`.

Hard rails: none touched. `src/lib/isPaying.ts` is read, not edited; the pricing URL is a string literal.

Browser check: not applicable — MCP tool result and a changelog JSON, no web code change.

### DIFF
--- a/src/services/mcp/McpServerFactory.ts
+++ b/src/services/mcp/McpServerFactory.ts
@@ -2,6 +2,7 @@ import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { z } from 'zod';
 
 import { McpToolsService } from './McpToolsService';
+import type { McpNextStep } from './McpToolsService';
 import {
   McpConvertOptions,
   McpCreateDeckOptions,
@@ -171,10 +172,16 @@ const DECK_RESULT_SHAPE = {
   message: z.string().optional(),
 };
 
-function errorResult(message: string, code = 'error'): ToolResult {
+function errorResult(
+  message: string,
+  code = 'error',
+  nextStep?: McpNextStep
+): ToolResult {
+  const structuredContent: Record<string, unknown> = { code, message };
+  if (nextStep) structuredContent.next_step = { ...nextStep };
   return {
     content: [{ type: 'text', text: message }],
-    structuredContent: { code, message },
+    structuredContent,
     isError: true,
   };
 }
@@ -462,7 +469,11 @@ export function buildMcpServer(context: McpRequestContext): McpServer {
           result.code,
           inputProps
         );
-        return errorResult(result.message, result.code ?? 'convert_failed');
+        return errorResult(
+          result.message,
+          result.code ?? 'convert_failed',
+          result.next_step
+        );
       }
       context.recordToolResult('convert_to_deck', true, undefined, inputProps);
       const raw = result as unknown as Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler for assistant users: the cap answer now carries the way past it in the same turn, and the funnel gets a paywall_shown keyed on both source and surface. Security confirmed only bounded integers and a literal URL reach the result and no rail file moves; engineering's round-two asks (source prop, create_deck redirect test) landed; voice clean.

### Blocking
- none

### Nits
- none

Issue #4401, opened 2026-09-12
chore: prune 147 upload rows whose bucket object the old sweep deleted


aalemayhu:
## Symptom

147 rows in `uploads` reference bucket keys that no longer exist. 141 of them were created in April 2026 across 19 owners, 88 are Ankify sync decks. `GET /api/upload/mine` returns rows without checking the object, so those owners see decks in their list that 404 on download (`src/controllers/Upload/UploadController.ts`, `src/controllers/DownloadController.ts`).

## Cause

The orphan sweep deleted objects that upload rows referenced until https://github.com/2anki/server/pull/3881 (merged 2026-07-27) restricted it to unreferenced keys. The fix stopped the deletions but never cleaned up the rows left behind.

## Fix

1. One-off: delete `uploads` rows whose key has no bucket object (list the bucket, `comm` against the table, delete the difference). Aggregates only in the PR body.
2. Guard: the listing endpoint, or a nightly check, drops or flags rows whose object is missing so this cannot silently recur.

## Evidence

Bucket listing (3533 objects) reconciled against `uploads` (3629 rows) on 2026-09-11. 51 objects have no row; all are `chat-attachments/` or `io-drafts/`, which own their own lifecycle.

PR #4419, opened 2026-09-12
chore: add a prune-dead-uploads ops command with dry-run

aalemayhu:
## What

Adds an owner-locked ops command `prune-dead-uploads` (route → controller → use case → repository) plus its button in the ops Commands tab. It lists the storage bucket once, loads every `uploads` row, and finds rows whose key has no bucket object and is not a reserved-prefix key (`mindmaps/`, `io-drafts/`, `assets/`). Dry-run by default: returns `{ dryRun: true, missingRows, sampleKeys }` (up to five keys) and deletes nothing. A live run (`dryRun: false` in the POST body) deletes exactly those rows through the upload repository and returns the deleted count. The button mirrors the delete-inactive-accounts check-then-confirm flow — the prune button unlocks only after a check in the same session finds rows.

Scope is item 1 of #4401 (the one-off cleanup). Item 2 (a listing-endpoint guard / nightly check so this cannot silently recur) is intentionally left for a follow-up.

## Why

The orphan sweep deleted objects that `uploads` rows referenced until #3881 (merged 2026-07-27) restricted it to unreferenced keys. The fix stopped the deletions but never cleaned up the rows left behind. Roughly 147 rows across 19 owners — most created in April 2026 — now point at objects that no longer exist, so those owners see decks in their list that 404 on download. This gives us a safe, previewable way to clear them.

## How

`PruneDeadUploadsUseCase` takes narrow interfaces for the bucket listing (`getContents`) and the upload repository (`getAllUploadReferences` + `deleteUpload`), mirroring `deleteDanglingUploadsInBucket` in reverse — one bucket listing, no HeadObject per row. Reserved-prefix filtering reuses `isReservedKey` from the sweep's helper so the two paths agree on what owns its own lifecycle. `getAllUploadReferences` is a new query-builder method on `UploadRepository`; it is not added to the shared `IUploadRepository` interface (the use case consumes a narrow local interface, matching the `IInactiveUserDeleter` pattern) so no existing repository mock breaks.

- **Ops command with dry-run, not the alternatives.** Chosen over a HeadObject-per-listing reconciliation (one bucket `ListObjectsV2` pass instead of thousands of HEAD calls) and over auto-deleting the rows inside the daily sweep (a one-off cleanup that quietly deletes rows on every deploy is the wrong shape; a manual, previewable button is safer for a destructive DB operation).
- **Assumption: the missing keys are genuinely absent**, not a transient listing gap or a bucket/env misconfiguration. The dry-run guards this — you see the count and sample keys before anything is deleted, and the live button only unlocks after a dry-run finds rows.
- **No prod data run tonight.** This PR ships the command only. Alexander runs the button (dry-run first, then live) against prod when he chooses. Nothing here touches production data.

## Measuring success

The command returns the deleted-row count in the response and the dry-run count is visible in the ops UI. After a live run, `GET /api/upload/mine` for an affected owner no longer lists the dead decks. A re-run dry-run should report `missingRows: 0` once the backlog is cleared.

## Testing

- Unit test `PruneDeadUploadsUseCase.test.ts`: fake storage handler (known key set) + fake upload repository — dry-run reports the count and deletes nothing; sample cap of five; live run deletes exactly the missing-key rows, keeps existing keys, skips reserved prefixes (asserts on the `deleteUpload` call args).
- `UploadRespository.test.ts`: new `getAllUploadReferences` filters null/empty keys and maps to `{ id, key, owner }`.
- Web: `pruneDeadUploads.test.ts` (fetch wrapper) and a `CommandsTab.test.tsx` case for the check-then-confirm gate. Route↔button parity test stays green.

Full-suite counts:
- Server `pnpm test`: 726 passed / 9 failed / 2 skipped suites; 7130 passed / 81 failed / 25 skipped / 7 todo tests. All 81 failures are the documented environmental suites — `getPageCount` (pdfinfo) and the Python-venv `DeckParser*` / `conversionGolden` / canary suites. Every touched suite passes.
- Web `pnpm --filter 2anki-web test`: 405 files, 3631 tests, all passing.
- `npx tsc -p . --noEmit`, `pnpm format:check`, web typecheck, and web lint: clean (lint shows only pre-existing warnings, none in the new files).

## Browser check

Notes: `pnpm --filter 2anki-web test:golden-path` (Playwright `golden-path.spec.ts`) — 4/4 passed, including "landing renders without console errors at 375px".

## Risks

Deleting the wrong rows if a key's object exists but the listing missed it (e.g. pagination cut-off). Mitigated: `getContents` paginates up to 100k keys and the dry-run count is inspected before any live run. Reserved-prefix keys are never touched. Rollback: the command is manual — simply don't run the live pass; the code path is inert until the button is pressed.

## Goal alignment

Serves **beautiful** indirectly: it stops affected owners seeing broken decks that 404 on download, so their deck list reflects what actually exists. Internal ops surface — no user-facing metric moves; the signal is the dry-run `missingRows` count trending to 0 after the cleanup, read from the ops Commands tab.

## Changelog

No entry — internal-only ops command (owner-locked), no user-visible change.

## Review round 2

Both reviewers asked for a fail-safe: an empty or paging-capped bucket listing would have made every row look dead. The use case now refuses (`UnsafeBucketListingError`) before judging any row, on dry-run and live alike, and the route answers 409 with the reason. Two tests cover it.

no changelog entry — owner-only ops command, nothing a user would notice.

### DIFF
--- a/src/controllers/OpsController.ts
+++ b/src/controllers/OpsController.ts
@@ -32,6 +32,7 @@ import { GrantUnclaimedPassUseCase } from '../usecases/passes/GrantUnclaimedPass
 import { SetBlockIdIdentityUseCase } from '../usecases/ops/SetBlockIdIdentityUseCase';
 import { ChangeUserEmailUseCase } from '../usecases/ops/ChangeUserEmailUseCase';
 import { GetTodaySnapshotUseCase } from '../usecases/ops/GetTodaySnapshotUseCase';
+import { PruneDeadUploadsUseCase } from '../usecases/ops/PruneDeadUploadsUseCase';
 
 class OpsController {
   constructor(
@@ -63,7 +64,8 @@ class OpsController {
     private readonly setBlockIdIdentityUseCase?: SetBlockIdIdentityUseCase,
     private readonly changeUserEmailUseCase?: ChangeUserEmailUseCase,
     private readonly getEmailDeliveryMetricsUseCase?: GetEmailDeliveryMetricsUseCase,
-    private readonly getTodaySnapshotUseCase?: GetTodaySnapshotUseCase
+    private readonly getTodaySnapshotUseCase?: GetTodaySnapshotUseCase,
+    private readonly pruneDeadUploadsUseCase?: PruneDeadUploadsUseCase
   ) {}
 
   async changeUserEmail(req: express.Request, res: express.Response) {
@@ -419,6 +421,22 @@ class OpsController {
     } catch (error) {
       console.error('[ops] deleteInactiveUsers failed', error);
       res.status(500).json({ message: 'Failed to run inactive user deletion' });
+    }
+  }
+
+  async pruneDeadUploads(req: express.Request, res: express.Response) {
+    if (this.pruneDeadUploadsUseCase == null) {
+      res.status(500).json({ message: 'Dead upload prune not configured' });
+      return;
+    }
+    try {
+      const body = req.body as { dryRun?: unknown };
+      const dryRun = body?.dryRun !== false;
+      const result = await this.pruneDeadUploadsUseCase.execute(dryRun);
+      res.status(200).json(result);
+    } catch (error) {
+      console.error('[ops] pruneDeadUploads failed', error);
+      res.status(500).json({ message: 'Failed to prune dead uploads' });
     }
   }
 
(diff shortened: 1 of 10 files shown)


aalemayhu (commented):
**Verdict**: approve

Owner-only ops command with a dry-run default, the same auth middleware as its neighbours, per-row bounded deletes, and a fail-safe that refuses an empty or paging-capped bucket listing (round two, both reviewers asked for it). Security, engineering, and voice forks clean; the button mirrors the delete-inactive-accounts pair with a confirm before the live run.

### Blocking
- none

### Nits
- none

Issue #4408, opened 2026-09-12
feat: pre-flight import warnings in the conversion report


aalemayhu:
## Idea

Anki reports import problems after the fact in a log most users dismiss. 2anki can compute the same checks before download and show them in the conversion report, using Anki's own vocabulary (added / updated / skipped / failed).

## Checks, with prod prevalence from a 185-deck sample on 2026-09-11

- Package over 100 MB: 23 stored decks exceed AnkiWeb's per-file media and collection limits, largest 878 MB. Warn that AnkiWeb sync will refuse it.
- Image-only front with empty back: 34 notes in one live sync deck.
- Cloze syntax in a non-cloze deck: 14 notes (https://github.com/2anki/server/issues/4402).
- Media in the package that no note references: 24 files in sync decks. Anki drops them silently; wasted bytes.
- Duplicate guid inside one package: 5 decks.
- Deck nesting deeper than 10 levels, empty deck path components, which Anki renames to `blank`.

## Not in scope

Anything that needs the user's collection (notetype conflicts, already-present notes). Those stay in Anki's log.

Feature work, so trio first.

aalemayhu:
Checked against current main: 3 of the 6 checks this issue asked for already shipped as separate PRs — package-over-100MB (#4421), duplicate-guid-inside-package as shared-question (#4423), and cloze-syntax-in-non-cloze-deck (#4415). None of those PRs closed this issue, so it stayed open with stale scope.

Still missing: image-only front with empty back, media in the package no note references, and deck nesting deeper than 10 levels. Leaving this open, scoped down to those three. Feature work, so still needs a trio pass before implementation per the issue's own note.

aalemayhu:
Investigated tonight before implementing. Three of the six proposed checks are already shipped or overlap with existing coverage, so the issue's scope has shrunk:

- **Package >100 MB**: shipped in #4421 (2026-09-12, `warn-when-a-deck-is-too-big-for-ankiweb-sync`).
- **Cloze syntax in a non-cloze deck**: shipped separately, already tracked as #4402 and shipped 2026-09-12.
- **Image-only front with empty back**: overlaps with the existing `empty_back` reason_code in `buildConversionReport.ts` (any card with a blank back already gets flagged, including image-only fronts) — a distinct notice here would be redundant.

That leaves three genuinely new candidates: unreferenced media, duplicate guid inside one package, and deck nesting depth / empty path components.

[...]

PR #4421, opened 2026-09-12
feat: warn when a built deck is too big for AnkiWeb to sync

aalemayhu:
## What

The sync upload path measures the built `.apkg` and, above 100 MB, adds `apkg-over-100mb:<mb>` to the deck's warnings. `resolveUploadWarning` turns it into one sentence in the existing `X-Warning` banner: "This deck is 210 MB. AnkiWeb won't sync packages over 100 MB, so split it into smaller decks before syncing." First slice of #4408; the issue stays open for the other checks.

## Why

Refs #4408. A prod sample on 2026-09-11 found 23 stored decks over AnkiWeb's 100 MB limit, the largest 878 MB, with no warning anywhere. Simpler: the user learns before syncing fails silently. Metric: none instrumented; the banner can grow an event once it exists.

- Slice: the size warning only (pm). The other pre-flight checks in the issue (unreferenced media, duplicate guids, deck nesting) need the package opened and are not a one-night change.
- Priority: ranked ahead of the password, markdown-heuristic and stray-cloze warnings in the single-warning channel (engineer's call, flagged as a product decision). Assumption: "this will not sync" beats "some cards were skipped". If wrong, move the branch below the password check in `resolveUploadWarning`.
- Scope: sync single-package path only, where the buffer is already in hand. The async job path has no warning channel today and the batch (multi-deck) response would need per-package sizes; both left out on purpose.
- Copy (pm): the size in whole MB and the one action, no exclamation.

## Tests

- `apkgOversizeWarning.test.ts` (new): silent at and under the limit, coded rounded MB above it, the rendered sentence.
- `UploadService.test.ts`: the coded size outranks the markdown-heuristic warning.

Browser check: not applicable — server-side warning text through the existing `X-Warning` banner, plus a changelog JSON.

### DIFF
--- a/src/lib/upload/apkgOversizeWarning.test.ts
+++ b/src/lib/upload/apkgOversizeWarning.test.ts
@@ -9,16 +9,21 @@ describe('apkgOversizeWarning', () => {
     expect(apkgOversizeWarning(12_345)).toBeNull();
   });
 
-  it('codes the rounded size in megabytes above the limit', () => {
-    expect(apkgOversizeWarning(105 * 1024 * 1024)).toBe('apkg-over-100mb:105');
+  it('codes the size to a tenth of a megabyte, rounded up', () => {
+    expect(apkgOversizeWarning(105 * 1024 * 1024)).toBe(
+      'apkg-over-100mb:105.0'
+    );
     expect(apkgOversizeWarning(100 * 1024 * 1024 + 1)).toBe(
-      'apkg-over-100mb:100'
+      'apkg-over-100mb:100.1'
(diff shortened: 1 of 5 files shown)


aalemayhu (commented):
**Verdict**: approve

Simpler: a deck too big for AnkiWeb says so before the user finds out at sync time. Security verified the header value is ASCII and integer-only; engineering and voice asked for a one-decimal, rounded-up size so the boundary never contradicts the limit, and round two confirmed it.

### Blocking
- none

### Nits
- none

Issue #4406, opened 2026-09-12
feat: mark the Image Occlusion notetype as Anki stock kind so the mask editor accepts it


aalemayhu:
## Symptom

Image Occlusion decks from 2anki import and review fine, but Anki's mask editor refuses them: `io_notetype_if_valid` errors with "Not an image occlusion notetype" unless `originalStockKind == IMAGE_OCCLUSION` (`rslib/src/image_occlusion/notetype.rs:124`). Users cannot edit or add masks on 2anki IO notes inside Anki. Restore to Default is also unavailable. The notetype is named `Image Occlusion`, the same as Anki's stock one, so users who already have the stock notetype get a `+`-suffixed duplicate.

## Change

In `create_deck/helpers/get_model.py` and `src/templates/n2a-io.json`:
- `originalStockKind: 6`
- field `tag` 0–4 in order Occlusion, Image, Header, Back Extra, Comments; `preventDeletion: true` on the first four
- a distinct name such as `2anki Image Occlusion`
- keep the existing template and occlusion string format, which already match Anki's

genanki's `to_json` does not emit `originalStockKind`, so a small subclass or post-processing step is needed for that one key.

## Effect on existing users

Name is not part of Anki's schema comparison, and the five fields keep their count and order, so a returning user's existing IO notetype is updated in place and the mask editor starts working on notes they imported earlier. Restore to Default then swaps in Anki's stock IO templates, which have the same five fields, so it is not destructive.

Blocked by https://github.com/2anki/server/issues/4404.

PR #4422, opened 2026-09-12
feat: pin stable field and template ids on every 2anki notetype

### LINKED ISSUE #4405: feat: pin stable field and template ids on every 2anki notetype

aalemayhu:
## What

Every 2anki notetype now ships a stable integer `id` on each field and template, plus `originalId` set to the notetype id. Field ids for the shipped notetypes (`src/templates/n2a-*.json`: basic, cloze, hierarchy, input, io, mcq) are explicit, frozen literals. Template ids, `originalId`, and ids for dynamically built notetypes (custom/transform, image occlusion) are emitted by a small genanki `Model` subclass (`N2AModel`) in `create_deck/helpers/n2a_model.py`. No field name, order, or template count changed; the Image Occlusion stock-kind metadata is untouched (that's #4406).

## Why

Anki 2.1.67+ reconciles an incoming notetype with the collection copy by matching each field and template by `id` when present, and by name only when an id is missing (`rslib/src/notetype/merge.rs` `is_match`). 2anki emitted no ids (0 of 185 sampled prod decks), so any rename of a field or template forks the notetype on a returning user's next import and drops every one of their notes into "conflicting". About 22% of deck owners re-upload, so this is a recurring loss of review history, not a corner case. Closes #4405.

- **Id derivation — frozen JSON literals, not live derivation.** The brief suggested deriving each id from a hash of `<notetype name>:<field/template name>` at generation time. I used that derivation to *seed* the values, but wrote the shipped field ids as **frozen literals** in `src/templates/n2a-*.json` rather than recomputing them live. Reason: if the id were recomputed from the live name every build, a future field rename would change the id and refork the notetype — defeating the whole point (an id only adds value over name-matching precisely when the name changes). Frozen literals make "renames go through ids, never through name changes" actually hold: a future rename edits the JSON `name` and keeps the `id`. The derivation is `int(sha1("<notetype>:<field>").hexdigest(), 16) % 10**13` (documented in `n2a_model.py`); the literals are reproducible from it.
- **Where `originalId` and template/dynamic ids are emitted.** genanki's `Model.to_json` passes field/template ids through untouched but never emits notetype-level keys, so `N2AModel.to_json` sets `originalId = model_id` and fills a `setdefault` id on any field/template lacking one (templates always; custom/transform notetype fields, which have no JSON to freeze — those derive from the name, which is correct since they mirror the user's own uploaded deck). `get_model` and `get_custom_model` now build `N2AModel` instead of `Model`, so all production paths (including Image Occlusion) are covered.
- **The returning-user test proves in-place update with no full sync.** Verified with Anki's real importer (`anki==26.8.1`): build the id-carrying apkg, strip the ids to simulate a notetype a user imported before this change, import into a fresh collection (30 new, 0 conflicting, field ids null), then import the id-carrying build (newer export time) into the same collection. Result: 0 new, 0 conflicting, **30 updated**, notetype count unchanged, and after reopening the collection the field/template ids and `originalId` are written in. A byte-identical re-import is all 30 duplicate. No `merge_notetypes` flag and no forced full sync were needed — this is the default importer path. (The test reopens the collection before asserting because `import_anki_package` does not invalidate the in-memory `models` cache — a subtle trap worth knowing.)

## Measuring success

Sample prod apkgs after deploy: `zipget` the notetype JSON and confirm `flds[].id` / `tmpls[].id` / `originalId` are present (was 0 of 185 before). Longer term, the signal is the absence of "conflicting" reports from returning users after a future template rename.

## Tests

- `create_deck` full pytest: **138 passed** (incl. 2 new: `test_returning_user_reimport_writes_field_ids_in_place`, `test_reimport_with_ids_is_all_duplicate`, and 5 new `helpers/test_n2a_model.py` unit tests).
- `pnpm test src/lib/templates`: **21 passed** (the TS export path reads only field names, unaffected by the new `id` key).
- `npx tsc -p . --noEmit`: clean.
- Full server `pnpm test`: **7222 passed**, 25 skipped, 7 todo, **2 failed** — both the documented `getPageCount` pdfinfo-ENOENT environmental failures (no pdfinfo binary in this env), unrelated to this change.
- Web `typecheck` clean; web `lint` exit 0 (pre-existing warnings only); `WhatsNewPage`/changelog vitest **7 passed**.
- `create_deck` pylint: **10.00/10** (CI `fail-under=10`).

## Changelog

`web/src/pages/WhatsNewPage/changelog/2026-09-12-reuploading-a-deck-keeps-updating-your-existing-anki-cards.json` — "Re-uploading a deck keeps updating your existing cards in Anki even after 2anki changes a card layout"

Browser check: not applicable — apkg generation and a changelog JSON, no web code change.

## Goal alignment

Serves **simpler**: re-uploading a deck keeps updating a returning user's existing cards instead of forking a duplicate notetype and stranding their review history. Metric: absence of "conflicting"/duplicate-notetype reports from re-uploaders, read from support inbox after the next template rename.


aalemayhu (commented):
**Verdict**: approve

Simpler for returning users: every 2anki notetype now carries frozen field and template ids plus originalId, so a future rename cannot fork the notetype and land their notes in Anki's conflicting bucket. Proven with Anki's own importer in the returning-user test (updated, not new; zero conflicting; notetype count unchanged). Security, engineering, and voice forks clean.

### Blocking
- none

### Nits
- none

PR #4423, opened 2026-09-13
feat: warn when cards in a built deck share a question

aalemayhu:
## What

After the package is built, count how many notes share a guid (python's `guids.json` sidecar, the real issued guids) and surface it on the upload warning channel as `duplicate-guid:<n>`. `resolveUploadWarning` turns the summed count into one sentence:

> 1 card repeats the question of another card in the same deck, so Anki keeps only the first and you import 1 card fewer than you see here. Give it a different question and convert again.

Priority: sync-size > locked PDF > duplicate guid > markdown heuristic > stray cloze.

## Why

Refs #4408, third of the pre-flight import checks. Anki imports the first note of a guid group and reports the rest as updated/skipped in a log most users dismiss, so a deck that shows N cards in 2anki lands with fewer and nobody is told. 5 of 185 sampled prod decks (2026-09-11) carried a duplicate guid. More beautiful: no silently missing cards. Metric: none instrumented, internal; the warning is visible in the upload response.

## Trio synthesis

- pm: de-dup anonymous guids with an ordinal (like #4388) instead of warning; warn only as a stopgap.
- designer: warn through the existing one-sentence channel, never say "guid", rank just under oversize.
- engineer: count from the sidecar, inject in the single-package branch mirroring the oversize warning, python-gated e2e.
- Conflict resolved: warn now. #4408 scopes this as a warning, de-dup touches the python guid formula fresh from #4388/#4422, and forking identical-back true duplicates turns a silent merge into visible clutter. De-dup gets a follow-up issue with pm's back-diff test on the colliding decks.

## Scope

- Single-package uploads only, the same scope the oversize warning took. Zip batches and Ankify sync builds do not warn.
- Image occlusion decks (`create_io_deck.py`) write no sidecar, so they never warn.
- Server-side English string like the other `X-Warning` sentences; not localized.

## Tests

- `countDuplicateGuids.test.ts`: distinct, grouped, missing guid, no file, malformed, non-array.
- `duplicateGuidWarning.test.ts`: code and singular/plural copy.
- `UploadService.test.ts`: priority, summing, and an end-to-end `X-Warning` on a sidecar with a collision.
- `DeckParser.stableGuids.test.ts`: real python builds two identical toggles onto one guid and the sidecar counts 1 (skips under `SKIP_CREATE_DECK`).
- Full server suite: 7246 passed, 2 failed (`getPageCount`, documented pdfinfo environmental), 25 skipped.
- Browser check: not applicable, server-only change plus a changelog JSON.

### DIFF
--- a/src/lib/anki/countDuplicateGuids.test.ts
+++ b/src/lib/anki/countDuplicateGuids.test.ts
@@ -1,0 +1,49 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { countDuplicateGuids } from './countDuplicateGuids';
+
+function workspaceWith(sidecar: string | object | undefined): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'dup-guids-'));
+  if (sidecar !== undefined) {
+    const raw = typeof sidecar === 'string' ? sidecar : JSON.stringify(sidecar);
+    fs.writeFileSync(path.join(dir, 'guids.json'), raw);
+  }
+  return dir;
+}
+
+describe('countDuplicateGuids', () => {
+  it('is zero when every note has its own guid', () => {
+    const dir = workspaceWith([
+      { notionId: null, guid: 'a' },
+      { notionId: null, guid: 'b' },
+    ]);
+    expect(countDuplicateGuids(dir)).toBe(0);
+  });
+
+  it('counts the notes Anki drops, one fewer than each group', () => {
+    const dir = workspaceWith([
+      { notionId: null, guid: 'a' },
+      { notionId: null, guid: 'a' },
+      { notionId: null, guid: 'a' },
+      { notionId: 'block', guid: 'b' },
+      { notionId: 'block', guid: 'b' },
+      { notionId: null, guid: 'c' },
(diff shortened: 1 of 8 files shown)


aalemayhu (commented):
**Verdict**: approve

More beautiful: a deck no longer lands in Anki silently short of the cards 2anki showed. Security fork: no issues (sidecar read is server-owned workspace, parse is try/catch + shape-checked, the code goes through a strict regex before copy). Engineering fork: clean on layers, types, colocated tests, python-gated e2e; one nit (test restore only on the happy path) fixed in 1080da2. UX fork: copy matches the sibling what/why/action shape, no banned words, no implementation leaks; one nit (run-on sentence) fixed in 1080da2.

### Blocking
- none

### Nits
- both fixed on the branch

aalemayhu (commented):
Re-marking the same reviewed diff: the head moved by an empty commit that only retriggers SonarCloud after its outage. No source change since the review in 1080da2.

aalemayhu (commented):
Re-marking the same reviewed diff: second empty commit to retrigger SonarCloud after its outage. No source change since the review in 1080da2.

Issue #4425, opened 2026-09-13
feat: prepaid AI credits per plan, enforced by actual cost, soft limit with non-AI fallback


aalemayhu:
## Problem

AI spend is not bounded by what a plan pays. The only guard is a rolling $50 per user per 24h breaker plus a $25 per 30d email alert, the same for every plan. Last week one $6 24h pass ran 184 comprehensive conversions in 24 hours and cost $31 in Claude usage, 42% of the week's AI spend. Historically the tail is small (8 of 336 day passes ever used AI, p99 $1.46 per pass; 28 of 753 subscribers used AI in 30 days, p99 $4.25) but a single session can erase a month of margin, and nothing stops it before $50.

## Decision

Prepaid AI credits, enforced by our own ledger. Stripe collects money only.

- Stripe Billing Credits were checked and rejected: they apply only to metered subscription items and only at invoice finalization ("Customers can exceed their balance during the cycle"), cannot apply to one-off pass payments, and may not be used as stored value. Metronome offers real-time drawdown but is a separate platform with limited Checkout support. Our `ai_usage_recorded` events already are the meter.
- Credits are an abstract unit backed by actual recorded cost (1 credit = $0.01 of AI cost, internal only). Balance = grants in window minus consumed cost. Decrement follows real cost, so many small cheap conversions do not drain a plan and one giant PDF cannot blow it. Counting conversions was rejected: 6 conversions cost one user $61 in August, and another averaged $0.125 per call.
- Soft limit, never a blocked deck. Conversion at zero runs the standard parser and ships the deck with one warning on the result ("You're out of AI credits, so this deck was built without AI. Add credits to use AI on the next upload."). Chat and photo-to-deck have no non-AI path and stop with a calm add-credits notice. No upgrade wall.
- Balance checked once at conversion start with an estimate from input size; a conversion that starts with AI finishes with AI (overshoot bounded to one conversion). Per-call guard stays for chat and photo.
- Prepaid only. No overage, nothing billed after usage. At zero the user buys a pack before continuing.
- Balance is visible only where AI is on: the AI badge on the upload page ("312 credits left", warning color and "Add credits" link under ~10%) and one line on the account page. Not in nav, not for users with AI off, not for anonymous users.

## Allowances (margin rule: about 50% of net after Stripe fees)

| plan | credits | period |
|---|---|---|
| 24h pass | 300 | pass window |
| 7d pass | 500 | pass window |
| 120d pass | 1 500 | pass window |
[...]

PR #4426, opened 2026-09-13
feat: prepaid AI credits with a non-AI fallback at zero

aalemayhu:
## What

Part 1 of prepaid AI credits: replaces the flat `$50/day` AI spend breaker with per-plan **credit allowances** (cost-backed, 1 credit = $0.01 of recorded AI cost). `assertAiBudget` now runs before every Claude call at all 7 sites. At zero, an upload silently falls back to the standard parser with a warning (no 4xx), while chat / photo / tag / note-type return a calm coded stop. Every allowance holder sees their balance on the upload AI badge and the account plan line via a new `GET /api/ai/credits`.

Trio synthesis + spec: see the spec commit c947ab25eba0 (`Documentation/specs/ai-credits-1.md`, removed on this branch before graduation).

## Why

Refs #4425. A `$6` 24h pass ran 184 comprehensive conversions in 24h for `$31.33` — 42% of that week's AI spend — and the only guard was a flat `$50/24h` breaker that ended in an opaque 429. This caps worst-case single-account AI spend at the plan allowance while keeping decks building via the parser fallback. Margin protection, not growth.

## How

- **Allowance resolver** (`lib/claude/aiCredits/allowance.ts`, pure, unit-tested per plan shape): passes 300/500/1500 over the pass window; Stripe Unlimited 300 over the item-level `current_period_*` from `subscriptions.payload` with a rolling-30-day fallback when stale/absent; legacy `$2/€2` subs (`unit_amount <= 200`) get 100; lifetime `patreon` / `ankify_access` comps 300 over the calendar month.
- **Balance** = `allowance + unexpired grants − recorded spend / 0.01`, derived on every check (no cron). Grants (`ai_credit_grants` table, kanel type regenerated in this PR) stay empty until part 2 writes packs.
- **Guard** (`aiSpendGuard.ts`): `assertAiBudget` throws `AiCreditsExhaustedError` (402, `ai_credits_exhausted`) at zero and runs before **every** Claude call — the conversion entry, every chunk, the PDF vision entry, every page, and the interactive surfaces — so overshoot is bounded to one in-flight call. The `ai_credits_exhausted` event fires once per window from the balance reader (`getAiBudgetStatus`) the moment it observes `<= 0`. Fails open on a read error; skips anonymous.
- **Upload fallback**: `PrepareDeck` pre-checks one question — is the balance `<= 0` — and, if so, builds with the standard parser plus an `ai-credits-exhausted` warning (mapped by `resolveUploadWarning`, ranked under oversize). A mid-run guard trip ships the cards produced so far with the same warning.
- **Deleted**: `AI_SPEND_DAILY_CAP_USD`, `AiSpendCapError`, the 429 path, `ai_spend_cap_tripped`. The `$25` ops alert stays.
- **Web**: `useAiCredits` hook over the typed endpoint; badge readout + account line; strings in 10 locales; reduced-motion-aware fade-in.

**Overshoot / concurrency**: eventually consistent, no reservation. The guard re-reads the balance before every Claude call, so overshoot is bounded to one in-flight call. The interactive surfaces (chat/photo/tag/note-type) surface the coded stop; the single-file/zip upload path degrades to the parser, and a mid-run trip ships the cards produced so far.

**Cost/latency**: no new Claude calls, no prompt/model change. Adds a few small indexed DB reads per conversion start and per interactive call (replacing the two the old breaker already did); prompt caching unaffected.

## Measuring success

- `ai_credits_exhausted` distinct users at day 7 (analytics `events` table). At T+30d re-read the 90-day backfill query: does it still only stop subscriber-months whose 30-day AI cost already exceeds what they pay?
- Guardrail: conversion success rate and deck downloads stay flat (the fallback keeps decks building).
- AI cost per plan cohort vs net revenue in `/ops` AI usage, weekly.

## Deviations from spec

**Round 7 — the round-6 concurrency/allowance fixes, applied where they were missed (8 changes).** Six confirmed bugs plus two defensive one-liners, each mirroring an already-correct sibling in this PR, all TDD-first except the two behaviour-neutral/defensive ones (noted below): (1) `convertWithClaude` — the concurrent `file_conversion` document call, fanned out at `FILE_CONVERSION_CONCURRENCY` — now routes through `withAiBudget` instead of the plain `assertAiBudget`, closing the same overshoot race round 6 fixed for chunks/pages. (2) `runClaudeRestart` now pushes `AiCreditsTrippedWithSalvage.salvagedDecks` on a mid-loop trip instead of discarding the cards produced before the trip, so a single-file restart no longer fails the whole job at zero decks. (3) `conversionInvokesAi`'s PDF branch now also requires `processPDFs !== false`, matching `convertFile`'s real gate, so a PDF upload with PDF processing off and zero credits gets no false out-of-credits notice. (4) the Apple `unlimited` pass labels its calendar-month window `resets:'month'` (not `'period'`) so the account line formats the reset date in UTC and never renders a day early west of UTC. (5) `IUserPassRepository` now declares `findActivePasses`, the method `getPlanInputs` actually calls against the concrete class. (6) `passAllowanceOf` uses `betterAllowance` instead of `??` so the pass granting more credits wins — restoring the file's documented greater-of invariant. **Behaviour-neutral at today's credit constants** (a single anonymous pass is 300 and the unlimited pass is 300, so anonymous is always >= unlimited and `betterAllowance` breaks the tie toward it, exactly as `??` did); applied as an invariant-correctness change, so it has no red-first test. Plus two defensive one-liners: `getOwnerId('')` returns `null` instead of a synthetic valid-looking `0`, and `subscriptionMonthlyWindow` clones `now` instead of aliasing the caller's Date into `windowStart` (an unreachable-today clamp branch, so also no red-first test).

**Round 6 — real concurrency control replaces "no reservation" (findings 1, 2, 11).** The earlier "eventually consistent, no reservation; overshoot bounded to one in-flight call" claim was false under concurrency: `runChunks` fans out every chunk at once and the PDF path runs 4 pages in parallel, so N concurrent guard checks all read the same pre-call balance and could each spend, overshooting N-fold. This batch adds an **in-process per-user reservation ledger** (`aiCredits/inflightReservations.ts`): `withAiBudget` admits a metered call only when the recorded balance minus outstanding reservations is still positive, reserves a conservative slice (`RESERVED_CREDITS_PER_INFLIGHT_CALL = 10`) for the call's duration, and releases it on settle. Check-and-reserve is synchronous after the balance read, so concurrent admissions coordinate through the ledger. It wraps every fan-out chunk (default, heading-driven, floor-v1) and every PDF vision page. **Chosen over full per-user serialization** (option 1's strict form) because serializing all of a user's chunks would multiply comprehensive-conversion latency; the reservation preserves concurrency when the user has headroom and throttles only near the limit. **Known residual limitation:** the ledger lives in one process/worker-thread — a second concurrent conversion for the same user in a different worker thread does not share it, so overshoot is bounded per process, not cluster-wide. Also: a call's reservation releases when its Claude call returns, and there is a brief window before its recorded spend is queryable, so the bound is tight but not a hard zero. This is proportionate for a margin-protection feature; a DB-level atomic reservation would close the multi-process gap but is more invasive and was deferred.

**Round 6 — allowance is max(subscription, pass), not subscription-always-wins (finding 6).** Supersedes the "subscription wins over passes" note below: precedence let a legacy \$2 subscriber (100 credits) who buys a 1500-credit pass get only 100. `resolveAllowance` now returns whichever of the best active-subscription allowance and the best pass allowance grants more credits (ties → subscription). This does **not** re-introduce the batch-5 summing/window-transition bug: it is `max()` of two independently-computed allowances, never a sum of spend windows. Lifetime remains the floor. Relatedly (finding 7), the credits repository now evaluates **every** active `subscriptions` row (aligned with `SubscriptionService.getUserActiveSubscriptions`) and keeps the best, instead of cherry-picking one row by `updated_at` where a drifted legacy row could mask the real plan.

**Round 6 — a mid-run trip that produces zero AI cards degrades to the parser (findings 3, 4).** The spec's "no standard-parser recovery of the untouched files" holds for a *partial* AI deck (cards + warning ship, no engine mixing). But a single-file upload whose only chunk trips produces **zero** AI cards, and handing the Python exporter an empty deck throws `PythonZeroCardsError` and fails the whole upload. With no AI cards there is no engine-mixing concern, so this case now falls back to the standard parser carrying the credits warning — "never a blocked deck" wins. Mid-run trips also now carry the cards produced so far (`AiCreditsTrippedWithSalvage`) through `runChunks`, floor-v1, and the PDF vision path, so a partial trip ships those cards with the warning instead of masking the trip as a silent partial success.

**No start-of-conversion cost estimate (4th review).** Earlier batches carried a token/page cost estimator feeding a `decideConversionBudget` pre-check. Four review rounds each found new bugs in it (two estimators disagreeing, a 4× text-PDF over-price, a restart-path copy, stray `pdfinfo` logs). Batch 5 **deletes the estimate entirely** — the pre-check is one question, is the balance already `<= 0`. The always-on per-call guard is the only safety net, so no estimate is needed. This drops the spec's "start-of-conversion estimate" line; everything else the user locked (cost-backed credits, prepaid, never a blocked deck, balance on the badge/account) stands.

**"A started conversion finishes with AI" is not implemented — it cannot be made safe.** The spec said a conversion that begins with AI should complete with AI. Three consecutive high-severity bug-review rounds each found the same class of hole in every mechanism that tried to honour it (`budgetPreChecked`): any path that disables the per-call guard lets an un-estimated job run unbounded on one credit. The final design removes the bypass entirely — **the per-call `assertAiBudget` guard runs before every Claude call (entry, every chunk, every page) and is never disabled.** The start-of-conversion pre-check only decides the *early* fallback. If the guard trips mid-run (a concurrent conversion drained the balance), the conversion stops issuing Claude calls, **ships the cards produced so far**, and attaches the credits warning — **no standard-parser recovery of the untouched files** (a partial AI deck with a warning beats mixing two engines in one deck), and the job never fails. **Overshoot is bounded to one in-flight Claude call.**

**Subscription wins over passes — not summed (4th review, finding 5).** An earlier batch summed a subscriber's subscription allowance with an active pass. Batch 5 makes it precedence: an active subscription wins over any pass, a pass wins over the lifetime comp, lifetime is the floor. A subscriber who also holds a pass draws the subscription allowance only. Multiple *passes* still each carry their own credits, summed over the rows whose window contains now.

Two smaller, flagged deviations:
- **Async upload notice (3rd review, finding 3):** an async AI upload becomes an `uploads` row (the job is deleted on promotion) and that table has no per-upload notice column; the `jobs.conversion_report` channel that reaches the Downloads page goes with the deleted job. So a fallback warning cannot reach Downloads without a new `uploads` notice column. Per standing instruction I did **not** add a second migration; the async path logs the fallback, and the Downloads notice is deferred to a follow-up that adds the column.
- **Pass window anchor (cut list):** `user_passes` has no `created_at`, so each active row's window is `[expiresAt − duration, expiresAt]` from its own expiry — no stored purchase timestamp, no cross-row stacking arithmetic.

## Known limitations for part 2

- **Grant spend is tracked only within the current allowance window** (`aiCredits/balance.ts`). Balance is `allowance + unexpired grants - spend`, and spend is counted only from the current window start, so a grant is re-counted as fresh balance every time the allowance window resets — spend against a grant does not carry across the reset. Nothing in part 1 writes `ai_credit_grants`, so this is dormant now, but **part 2's grant-consumption design must account for it** (track spend against the grant itself, not against the rolling window) before credit packs ship. Written down here so it is not rediscovered mid-build.

## Testing

**Earlier review rounds (1-3, 30+ findings, all fixed TDD-first).** Prepaid cost-backed credits, the guard-before-every-metered-Claude-call design, the calm SSE/inline stop for chat/photo/tag, the billing-day allowance windows, durable `ai_usage_recorded` events drained on shutdown, and the 10-locale copy all landed across the first three rounds. Batch 5 subtracts the parts those rounds kept patching (see the round-4 entry below and Deviations); the round-1..3 test files it removed (`conversionCostEstimate.test`, `passWindow.test`) are gone with their sources.

**Fourth bug-review round (Batch 5 — 17 confirmed, a subtraction):** deleted the cost estimate (`conversionCostEstimate`, `estimateConversionCostUsd`, `ai-credits-short`, `decideConversionBudget`, `hasAiCreditsForConversion`), restored the per-chunk/per-page guards, dropped the mid-run parser recovery (`PrepareDeck` ships the produced cards + warning; the vision/quiz branches catch the trip and skip the file), rewrote passes to per-row windows with no stacking arithmetic (`allowance.test` covers summed active passes, earliest-start anchor, and a not-yet-active pass ignored), made subscription win over passes, centralised the `ai_credits_exhausted` fire in the balance reader (`aiSpendGuard.test`), passed a single-image drop's `e.code` through, moved fr/ru copy to the informal register, and added colocated tests for `createAiCreditReaders`, `getAiCredits`, and `useAiCredits`. **The net diff vs `origin/main` shrank from +2945 to +2633.**

**Gates (all green):**
- Full server suite: **739 passed suites / 7223 tests pass**. The 9 failed suites are the documented worktree environmental failures only — `PythonExitError` (no `create_deck` venv, apkg build) and `getPageCount` (`pdfinfo` missing). No logic/assertion failures.
- Full web suite: **410/410 files, 3678 tests pass**.
- `pnpm format:check` clean, `pnpm lint` clean (warnings only, no errors), `tsc -p . --noEmit` + web `typecheck` clean.

**Known scope note (pass anchor):** `user_passes` has no `created_at`, so each active pass row's window is `[expiresAt − duration, expiresAt]` from its own expiry — no stored purchase timestamp and no cross-row stacking arithmetic, which closes the stacking/mixed-kind exploit without a payments-table migration. **Cleanup note (subscription query):** `AiCreditsRepository` keeps its own subscription-by-email query because it needs the `payload` the auth lookups don't return; a literal reuse would make a `data_layer` repository import the `services` layer.

## Browser check

Notes: not run in this environment — the worktree has no dev server and I was instructed not to start `pnpm dev`. The web changes are covered by unit tests and reduced-motion CSS mirrors existing patterns. Please verify the badge/account line on `localhost:3000` before merge (this is a hard-rail PR you review anyway).

## Changelog

`web/src/pages/WhatsNewPage/changelog/2026-09-13-ai-credits.json` — "Every plan now includes AI credits, and your deck still builds without AI if you run out"

## Kanel

`ai_credit_grants` migration applied to the local DB, `pnpm dlx kanel -c ./.kanelrc.js` run against the live schema, and `src/data_layer/public/AiCreditGrants.ts` committed in this PR. No hand-writing.

## Sonar

## Risks

- Legacy `$2/€2` detection keys on `unit_amount <= 200`; a stored `subscriptions.payload` without item-level price defaults to the safer 300, not 100.
- The Apple `unlimited` pass kind is absent from the spec's allowance table; I mapped it to the Unlimited row (300, rolling 30 days) so paying Apple subscribers keep AI. Flagged as a resolved gap.
- Rollback: the change is additive plus a breaker swap; reverting restores the old breaker. The migration is `dropTableIfExists`-safe.

## Goal alignment

Serves **simpler** (a soft, self-explaining landing instead of an opaque 429 "contact support") and protects margin so the product stays sustainable. Metric: `ai_credits_exhausted` distinct users and cohort AI-cost-vs-revenue, read in `/ops` AI usage at day 7 and T+30d.

## Hard rail

This PR touches `migrations/`, `src/data_layer/public/`, and subscription-reading code, so it is a hard-rail PR: it goes **ready** with the review verdict and waits for Alexander in the GitHub UI. Not merged by an agent.

## i18n native-speaker pass

New strings shipped in all 10 locales. Flag for a native pass: the "AI credits" concept term (KI-Credits / créditos de IA / crédits IA / кредиты ИИ / AIクレジット), "Add credits", "built without AI", the ru/pl plural forms, and each locale's word for "deck".

## Reviews before ready

- **Security review** (`/security-review` on the PR diff): no high-confidence findings. `GET /api/ai/credits` sits behind `RequireAuthentication` and reads `owner` from `res.locals` only; every repository query keys on the server-supplied user id with bound parameters; no code path in this PR writes `ai_credit_grants`; the client-visible `aiCreditsExhausted` input is overwritten server-side; the response is a mapped `{ credits, allowance, windowEnd, resets }` shape; the exhausted error message is static; the new log line uses the hashed file label. Observation, not a finding: with the $50/day breaker gone, callers whose allowance resolves to null stay unmetered, which is only free users on the 5/month photo quota and anonymous pass holders, both unmetered before this PR as well.
- **Migration review**: ship it. `CREATE TABLE` only, no lock on existing tables; `(user_id, expires_at)` index covers the balance read; nullable unique `stripe_session_id` allows manual comps; FK cascade matches the GDPR sweep convention; `down` drops the table; kanel type matches column for column. Two notes: `source` has no CHECK constraint (validated in the app layer when part 2 adds the writer), and only the reader landed here by design, the writer arrives with the credit pack in part 2.
- **Designer review**: clean after hiding the add-credits links until part 2.


aalemayhu:
## `/code-review 4426 high` — round 6 (post "stop patching, subtract" simplification, head `a96c8550`)

Root architectural issue survives the batch-5 simplification: the guard checks balance **before** a Claude call, but concurrent calls all read the same pre-call balance since spend is only recorded after each call completes. Deleting the estimator removed the false-precision bugs but did not add a reservation/lock, so the "bounded to one call" overshoot claim is false whenever calls run concurrently.

1. **`src/lib/claude/ClaudeService.ts:1371`** — `runChunks()` fires every chunk concurrently (`Promise.allSettled`, no cap). Each chunk's `assertAiBudget` (line 920) reads the same stale balance before any chunk's cost lands, so a large comprehensive-mode doc can overshoot by N chunks' worth of spend, not one call.
2. **`src/lib/claude/generateDeckInfoFromPdfImages.ts:241`** — `runPagesWithConcurrency` runs up to `PDF_PAGE_VISION_CONCURRENCY` (4) page-vision workers concurrently, same stale-balance race — up to 4x overshoot on one PDF.
3. **`generateDeckInfoFromPdfImages.ts:232`** — when the guard trips for *some* pages mid-run, `AiCreditsExhaustedError` is swallowed into the generic per-page failures array. Since `failures.length < images.length`, the all-failed throw is skipped and a partial deck ships as a normal success — no `AI_CREDITS_EXHAUSTED_WARNING_CODE`, contradicting "ships cards produced so far with the same warning."
4. **`PrepareDeck.ts:798`** — the zero-card safety check only fires when `crossFileDedup` is set (2+ files). A single-file upload whose balance drains mid-run (e.g. concurrent conversion on the same account) falls through to `exporter.save()` with an empty deck → `PythonZeroCardsError`, failing the whole upload instead of degrading gracefully.
5. **`PrepareDeck.ts:849`** — `usesAiSettings()`/`resolveAiConversionGate()` gate on the user's *persisted* CardOption toggles, not on whether this upload's files actually need AI. A user with an AI toggle on but zero credits gets a false "built without AI, out of credits" notice on a plain markdown/HTML conversion that never touches Claude.
6. **`src/lib/claude/aiCredits/allowance.ts:236`** — `resolveAllowance()` lets any active subscription win over any pass unconditionally. A legacy $2 subscriber (100 credits) who buys a 1500-credit Semester Pass still gets only 100 credits — the "subscription already covers the paying user" comment is only true for the 300-credit standard tier.
7. **`src/data_layer/AiCreditsRepository.ts:71`** — `buildActiveSubscriptionQuery`/`getPlanInputs` picks one row via `orderBy('updated_at','desc').first()` with no product/price filter, diverging from the canonical `SubscriptionService.getUserActiveSubscriptions` (evaluates all active rows, as `hasAnkifyAccess` does). Given this repo's own documented risk of drifted `subscriptions` rows (`STRIPE_SYNC_ON_STARTUP=false`), a stale legacy row with a newer `updated_at` than the real Auto-Sync row makes the credits balance disagree with what the rest of the app tells the user their plan grants.
8. **`src/usecases/uploads/worker.ts:403`** — `getEventsSink().flushDurable()` is only awaited on the success path; the `catch` block returns `{ok:false}` without draining pending durable `ai_usage_recorded` writes. A conversion that makes billable calls and then throws later loses that recorded spend if the worker thread is recycled.
9. **`src/controllers/UsersControllers.ts:663`** — `getAiCredits` uses `if (!owner)`, the falsy-ID anti-pattern `security.md` explicitly forbids (CWE-754). Same pre-existing `Number(getOwner(res)) || null` pattern at `TemplatesController.ts:162/195` now feeds `AINoteTypeUseCase`'s new `assertAiBudget(userId ?? null)` — `userId=null` is treated as anonymous and the guard is silently skipped for that account.
10. **`web/src/pages/AccountPage/components/AiCreditsAccountLine.tsx:17`** — forces UTC formatting on `windowEnd` assuming it's always a UTC-midnight boundary, but `resets:'period'`/`resets:'pass'` carry the exact non-midnight Stripe timestamp — can render one calendar day off for users west of UTC.
11. **`src/lib/claude/aiSpendGuard.ts:162`** — `assertAiBudget`/`getAiBudgetStatus` re-resolve the full plan (user, passes, subscription, grants, spend-to-date) from scratch on every call with no per-conversion snapshot — the direct enabling condition for findings 1 and 2.
12. **`aiSpendGuard.ts:162`** (structural) — `assertAiBudget` is manually re-inserted at 10 independent call sites across 7 files with no parity test / lint rule enforcing every Claude-calling surface includes it. This repo already hit this exact shape twice (`events.parity.test.ts`, the ops-command/button parity test) — a future AI surface (part 2, or any new Claude call) can silently ship with no guard.
13. **`src/data_layer/createAiCreditReaders.test.ts:13`** — mocks internal repositories (`jest.spyOn` on `AiCreditsRepository.prototype.getPlanInputs`, `AiCreditGrantsRepository.prototype.sumActiveCredits`, `AiUsageMetricsRepository.prototype.userCostSince`) instead of exercising them for real — violates `testing.md`'s internal-collaborators-run-for-real rule, and would pass even with finding 7's bug present.
14. **`PrepareDeck.ts:903`** (dup, not new bug) — the gate/warning-derivation logic is duplicated verbatim between the sync entry point and `prepareDeckInfoOnly` (~1047-1061); a future fix to one (e.g. finding 5) needs to land in both.

Findings 1–9 are correctness bugs at high confidence (all quote-verified against the worktree source). 10 is a real but narrower correctness bug. 11–14 are structural/quality — no fix required to ship, but 12 in particular is worth addressing before part 2 adds more Claude call sites.

aalemayhu:
## Round 6 — findings resolved (head `426f8b4`)

Fixed findings 1–10 and 13; 11/12/14 assessed below.

**Concurrency (1, 2, 11)** — added an in-process per-user reservation ledger (`aiCredits/inflightReservations.ts`). `withAiBudget` admits a metered call only when recorded balance − outstanding reservations is positive, reserves a conservative slice for the call, releases on settle; check-and-reserve is synchronous after the balance read so concurrent chunks/pages coordinate. Wired around every fan-out chunk (default, heading-driven, floor-v1) and every PDF vision page. Chose the reservation over full per-user serialization to avoid multiplying comprehensive-conversion latency. Residual limit: single process/worker-thread scope, and a small release-vs-record-visibility window — bounded per process, documented in the PR body.

**Mid-run trip masking (3)** — `AiCreditsTrippedWithSalvage` carries the decks produced before the trip. `runChunks`, `runFloorV1`, and the PDF vision path throw it on a partial trip; `runClaudeConversion` ships the salvage and flags the credits warning instead of masking the trip.

**Zero-card crash (4)** — the empty-deck guard now fires for the single-file path too: an empty AI deck (single file or a fully-tripped run) falls back to the standard parser (shared `buildParserResult`) with the credits warning instead of crashing on `PythonZeroCardsError`. The multi-file "covered by earlier files" drop is preserved for the non-tripped case.

**False credits notice (5)** — `conversionInvokesAi(settings, files)` gates the pre-check on whether this upload's files actually reach Claude (claudeAIFlashcards always; the vision toggles only with a matching file type), so a plain markdown/HTML upload with a PDF/image toggle left on no longer gets a false "built without AI" notice.

**Allowance max (6)** — `resolveAllowance` returns max(best subscription, best pass); a legacy $2 subscriber who buys a 1500-credit pass now gets 1500. `max()` of two independently-computed allowances, not a sum of spend windows.

**All active subscription rows (7)** — `getPlanInputs` evaluates every active row (aligned with `SubscriptionService.getUserActiveSubscriptions`) and the resolver keeps the best, instead of cherry-picking by `updated_at`.

**Worker spend drain (8)** — `flushDurable()` now runs in `finally`, draining recorded `ai_usage_recorded` on the failure path too.

**Falsy owner id (9)** — `getOwnerId()` preserves a valid id of `0`; `getAiCredits` uses `== null`; the templates/retry-pdf call sites feeding `assertAiBudget` no longer treat owner `0` as anonymous.

**Account-line timezone (10)** — force UTC only for the calendar-month window; `period`/`pass` windows carry the exact timestamp and render in local time.

**Real repos in test (13)** — `createAiCreditReaders.test.ts` exercises the real repositories against an in-memory sqlite DB (getPlanInputs over multiple subscription rows, sumActiveCredits), so it would now catch a finding-7 regression. `userCostSince` stays over a stubbed knex builder because its jsonb query is Postgres-only (covered by the generated-SQL test).

**11 / 12 / 14** — 11's enabling condition is closed by the reservation ledger; per-call plan caching not added (measured as a few indexed reads, unchanged from the prior per-chunk guard). 12 (a guard-parity lint) and 14 (the two-occurrence gate/warning duplication) left as-is: 14 is only two occurrences (below the dedup threshold) and the gate logic is already shared via `resolveAiConversionGate`.

Gates: server tsc clean, web typecheck clean, full web suite 3681 pass, server suite 7254 pass (only the documented Python/pdfinfo environmental suites fail), `pnpm format:check` and both lints clean. Still **draft**; hard-rail, not merged.

aalemayhu:
## `/code-review 4426 high` — round 7 (post batch-6 concurrency fix, head `9800a292`)

Batch 6 fixed the concurrency-overshoot race with a per-user in-process reservation ledger (`withAiBudget`). Round 7 found the fix wasn't applied everywhere it needed to be, plus two independent bugs of the same "fixed one call site, missed the sibling" shape from round 6's own findings 5 and 10.

**Confirmed, fixing in batch 7:**
1. `claudeFileConversion.ts:66` — `convertWithClaude` still calls plain `assertAiBudget`, not `withAiBudget`, even though `PrepareDeck.ts` fans it out at `FILE_CONVERSION_CONCURRENCY=4` — the exact race class batch 6 closed elsewhere, missed here.
2. `UploadService.ts:875` — `runClaudeRestart`'s catch treats `AiCreditsTrippedWithSalvage` like a plain exhaustion error, never reads `salvagedDecks` — cards already produced before the trip are silently dropped. `PrepareDeck.ts`'s equivalent path does this correctly.
3. `allowance.ts:219` — `passAllowanceOf` uses `??` instead of `betterAllowance()` for anonymous-vs-unlimited pass, so an anonymous pass always masks a larger unlimited pass. Every other tier comparison in the file uses `betterAllowance()`.
4. `PrepareDeck.ts:970` — `conversionInvokesAi` checks `vertexAIPDFQuestions` but not `processPDFs`, unlike the real gate in `convertFile`. Round 6's finding-5 fix (no false "out of credits" notice on non-AI conversions) is incomplete for `processPDFs=false`.
5. `allowance.ts:90` — `unlimitedPassAllowance` labels its UTC-midnight calendar-month window `resets:'period'` instead of `'month'`; the frontend only forces UTC for `'month'`, so unlimited-pass users west of UTC see the reset date one day early — round 6's finding-10 fix, incomplete for this tier.
6. `UserPassRepository.ts:21` — `IUserPassRepository` doesn't declare `findActivePasses`, the method `AiCreditsRepository.getPlanInputs` actually calls against the concrete class.
7. `getOwner.ts:6` — `getOwnerId('')` returns `0` (`Number('')` is finite), not `null` — defensive fix.
8. `allowance.ts:174` — `subscriptionMonthlyWindow` aliases the caller's `now` Date into `windowStart` instead of cloning — latent shared-mutable-Date hazard.

**Confirmed but deliberately deferred (not fixed this batch):**
- `balance.ts:34` — spend against a grant is tracked only within the current allowance window, so a grant gets re-counted as fresh balance every window reset. **Currently dormant** — nothing in this PR writes to `ai_credit_grants` yet (that's part 2, credit packs). Documented in the PR body as a known limitation part 2's grant-consumption design must account for, not fixed here.
- Quality/efficiency only, no fix needed: `PrepareDeck.ts:291` (wasted round-trip on a doomed post-trip call, functionally caught), `ClaudeService.ts:1422` (3x duplicated salvage-throw, below rule-of-three), `AiCreditsRepository.ts:75` (duplicate subscription query — already an accepted round-6 deviation), `EventsSink.ts:73` (per-row durable insert — already an accepted round-6 durability trade-off), `PrepareDeck.ts:887` (empty-deck invariant enforced by caller branching rather than exporter boundary — architecture note).

One candidate refuted: `UploadService.ts`'s `e.code ?? 'image_conversion_failed'` — traced client-side, this is actually correct (the client already expects the more specific `no_ready_made_questions` code that was previously dead on arrival).

aalemayhu:
## `/code-review 4426 high` — round 8 (post round-7 batch + Sonar cleanup, head `e5274fb3`)

**Fixing in batch 8 (final planned batch):**
1. `web/src/lib/backend/templates.ts:190` — `aiError()` reads `data.error`, but the server's coded-error path (`TemplatesController.ts` → `ErrorHandler.tsx`) sends `{code, message}`, never `error`. The real "out of AI credits" message is discarded for a generic fallback ("The change couldn't be applied. Try rephrasing your request.") — actively misleading. `templates.test.ts` still asserts the old `{error: string}` contract.
2. `src/usecases/chat/ChatUseCase.ts:554` and sibling interactive surfaces (photo, tag, note-type) — use plain `assertAiBudget` (no reservation) unlike every conversion call site. Two concurrent requests from the same user (e.g. two browser tabs) can both pass the balance check and both spend — the same overshoot class round 6's reservation ledger closed for conversions, not extended to interactive surfaces.
3. `src/lib/claude/aiSpendGuard.ts:215` — `withAiBudget` (used by `convertWithClaude` for image-quiz/vertex-PDF vision) never calls `maybeNotifyAlert`. The pre-existing $25/30-day ops spend alert silently stops firing for any upload whose only metered calls go through this path — a real regression vs. the old `guardAiSpend`, which ran the alert check on every metered call.
4. `aiSpendGuard.ts:120` — `getAiBudgetStatus` only fires the `ai_credits_exhausted` analytics event when raw `balance.credits <= 0`, but `withAiBudget`'s reservation-aware check can throw earlier (`balance.credits - reserved <= 0`). Reservation-throttled stops never record the event, undercounting the PR's own stated success metric for exactly the concurrency-throttled population.
5. `web/src/pages/PhotoToFlashcardsPage/PhotoToFlashcardsPage.tsx:289` — falls into the generic `!res.ok` branch for a 402 credits-exhausted response instead of a dedicated handler like `ChatPanel.tsx` has — shows raw unlocalized server text, doesn't refresh the credits badge.

**Confirmed but explicitly not fixing (accepted debt, documented pattern from prior rounds):**
- `allowance.ts:178` — `windowEnd` returned as a direct alias of the caller's Date (not cloned like `windowStart` was in round 7) — latent, no current caller mutates it.
- `balance.ts:37` — float division by `0.01` before rounding can flip direction at an exact half-credit boundary — sub-cent edge case, cosmetic.
- 7 dedup/architecture findings (`conversionInvokesAi` drift, `AiConversionState` mutable side-channel, `getPlanInputs` query ordering, `withAiBudget` wrapping-shape duplication across 4+ call sites, duplicate date formatters, duplicate `PrepareDeckResult` object literals) — same category of accepted DRY/architecture debt as prior rounds' deviations, not correctness bugs.

One claim investigated and refuted: `TemplatesController.ts`'s `throw error` for `HttpCodedError` does NOT crash the process — Express 5's router correctly forwards a rejected promise to `ErrorHandler` middleware. The bug is purely client-side (finding 1 above): the server delivers the right message, `aiError()` reads the wrong field.

This is the last planned fix batch for this review cycle — after it lands, verification is CI green + Sonar clean, not another full 8-angle review round.

aalemayhu (commented):
Security, engineering, and UX/voice forks all clean on the final head — no blocking findings, no high-confidence nits beyond a pre-existing untested-file gap unrelated to this PR (RetryPdfWithCredentialUseCase.ts, not introduced here). This closes 8 rounds of adversarial /code-review on top of the fan-out: concurrency-overshoot race, allowance/subscription-row bugs, salvage-on-trip handling, error-message delivery, and spend-alert coverage all confirmed-fixed and re-verified against CI green + SonarCloud zero findings. Hard-rail (migration + subscription-adjacent code) — already queued for merge by Alexander.

PR #4429, opened 2026-09-16
feat: credit pack purchase

aalemayhu:
## What
Adds a one-time **$5 / 250-credit** AI credit pack (90-day expiry) that a signed-in user can buy from three surfaces: the "built without AI" conversion-result notice, the upload-page AI-credits badge (at ≤25 or zero credits), and the account-page credits line (any active allowance). Completes the zero-balance dead end shipped in #4426 — instead of waiting weeks for the allowance to reset, the user can top up.

## Why
Closes the recourse gap from the AI-credits launch: one day after #4426, a paying subscriber hit zero credits with ~three weeks to reset and asked support how to get more. Today the answer is "wait." User-visible outcome: a spent balance now has a buy action rather than a stop. This is pure revenue upside — #4426 already capped the AI cost leak.

## How
- **Backend** mirrors `CreatePassCheckoutUseCase` minus the anonymous branch: `POST /api/checkout/credit-pack` (`RequireAuthentication`) → `CreateCreditPackCheckoutUseCase` → Stripe Checkout (`mode: payment`, `credit_pack` metadata). Success/cancel URLs are resolved **server-side** from a `source` enum (`credits_conversion`/`credits_badge`/`credits_account`) → path map — never a client-supplied redirect URL.
- **Webhook** grant branch in `checkout.session.completed`: derives credits from the server-side SKU constant (`CREDIT_PACK_CREDITS = 250`), sanity-checks `amount_total >= CREDIT_PACK_PRICE_CENTS`, and inserts into `ai_credit_grants` (`source: 'pack'`, 90-day expiry) via `insert().onConflict('stripe_session_id').ignore()` for idempotency. `credits_purchase_completed` fires only when a row is actually inserted (not on redelivery).
- `AiCreditGrantsRepository` gains a writer (`insertPackGrant` + `IAiCreditGrantsWriter`); the reader path (`sumActiveCredits`) is untouched.
- **Frontend**: one shared `BuyCreditsButton` (owns pending/error state) wired into the three surfaces, a `startCreditPackCheckout` client helper, and `?credits=added` return handlers on Upload/Account that show a calm confirmation and invalidate the AI-credits query so the balance updates in place.
- SKU price id read from `process.env.CREDIT_PACK_PRICE_ID` (env-var name only in `src/env.example`, mirroring `PASS_*_PRICE_ID`); nothing hardcoded.

## Measuring success
- `credits_purchase_completed` (fires on the idempotent insert) and `credits_buy_clicked` (fires per surface click) — the `credits_buy_clicked → credits_purchase_completed` funnel, read on the AI-usage ops page / `/api/ops/business/metrics`.
- Structured webhook log `credits.granted` (`user_id`, `amount_credits`, `expires_at`, `inserted`) confirms grants land; `credits.webhook.amount_mismatch` / `credits.webhook.missing_metadata` surface malformed sessions.
- **Day-7 prod check**: grants writing correctly, no webhook errors, at least one real purchase or a clear zero.

## Testing
- Unit tests added: `CreateCreditPackCheckoutUseCase` (redirect allowlist per source, SKU metadata, customer email/id, GA attribution), `CreditPackCheckoutController` (source validation drops arbitrary strings, 401 without owner), `AiCreditGrantsRepository` (generated-SQL test for the `on conflict … do nothing` shape), `WebhookRouter` (grant, purchase event, **idempotent redelivery inserts once / counts once**, missing metadata, amount-below-price, non-pack lifetime session), `BuyCreditsButton`, `startCreditPackCheckout`, `formatCreditPackExpiry`, plus the three surface components and the Account `?credits=added` confirmation.
- Full server Jest suite: **green**. Full web Vitest suite: **413 files / 3694 tests green**. `pnpm format:check` (tree-wide) clean. Server `tsc -p .` (incl. tests) green; web typecheck green.
- Both `KNOWN_EVENTS` allowlists updated in the same PR; `events.parity.test.ts` passes.

## Browser check

Notes: No live Stripe price was configured in this environment and Netlify previews redirect to prod (decorative), so the checkout redirect + return flow was verified through the unit/integration suites (redirect-URL assertions, `?credits=added` handler tests) rather than a live click-through. Alexander can exercise the live golden path once `CREDIT_PACK_PRICE_ID` is set on prod (see **Ops** below).

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-16-buy-more-ai-credits-when-you-run-out.json` — "Buy more AI credits when you run out, instead of waiting for your allowance to reset"

## Ops — one manual step for merge
A live Stripe Price already exists for this pack. Before/at merge, add to prod's `.env` (I do not and will not write to the prod host):

```
CREDIT_PACK_PRICE_ID=price_1UGHShB4lekI0uHmlHBo7uXB
```

- Product: `prod_VGp6HTzqrnUFpN` ("2anki.net - AI Credit Pack")
- Price: `price_1UGHShB4lekI0uHmlHBo7uXB` ($5.00 USD, one-time, "AI Credit Pack $5 / 250 credits")

This is the **only** manual step beyond clicking merge. Until it is set, `POST /api/checkout/credit-pack` returns 503 and the buy actions stay hidden — the safe default.

## Deviations from spec
- **Open question #2 (which conversion UI renders):** both render today — the non-blocking "built without AI" success notice (single-deck `X-Warning` header + batch `warning` field) and, only when zero cards result, the blocking `ai_credits_exhausted` error. The spec locks Trigger A to the **built-without-AI** outcome, so the buy link attaches there. To gate it reliably (rather than string-matching the warning text), the server now also emits the warning **code**: an `X-Warning-Code` header (single deck) and a `warningCode` field (batch). This is the small server response addition the spec anticipated. The blocking `ai_credits_exhausted` error is left without a buy link per the spec's acceptance criterion; it's a candidate fast-follow alongside ChatPanel.
- **Open question #1 (ChatPanel 4th placement):** not in this PR; filed as a follow-up so scope stays to the three locked surfaces.
- **Open question #3 (expiry model):** shipped the simpler rolling window — `sumActiveCredits` already sums every unexpired grant, and each pack carries its own `expires_at`, so two packs simply both count until each expires; no per-pack tracking UI needed.

## Known limitation (documented, not fixed here)
Pack credits count toward a user's balance only while they have an active plan/pass allowance: `computeAiCreditBalance` returns `null` (ignoring grants) when `resolveAllowance` is null. A user who buys a pack and then lets their subscription/pass lapse before using it loses visibility into those credits until they have an active plan again (the row survives up to 90 days; it is not deleted). Fixing this is a deliberate read-side change to `balance.ts` — a separate decision, out of scope here.

## Risks
- **Open redirect** — mitigated: redirect URLs come from a server-side enum→path map, never client input.
- **Double grant on webhook redelivery** — mitigated: unique `stripe_session_id` + `onConflict.ignore()`; the purchase event fires only on a real insert (tested).
- **Spoofed `?credits=added`** — cosmetic only: the banner is param-driven, but no credits are granted client-side; the real balance (unchanged) shows on refetch. Same shape as the existing `?subscribed=1` banner.
- **Webhook race** — the success redirect can beat the async webhook, so the refetched balance may briefly not include the new grant; it lands on the next natural refetch. The confirmation banner is shown regardless, so the user gets immediate positive feedback.
- Rollback: unset `CREDIT_PACK_PRICE_ID` on prod → route 503s and the buy actions disappear; already-granted rows remain valid.

## Goal alignment
**Faster** — removes a forced multi-week wait for a paying user mid-study. Metric: `credits_buy_clicked → credits_purchase_completed` attach rate among zero-balance users, read on the AI-usage ops page; T+30d adoption review verdict bar is ≥5% attach (keep/remove). Also revenue-positive, which is upside, not the goal test.

## Hard rail
Checkout + webhook + Stripe → hard-rail. Ships to **ready** with the security-review verdict posted; **Alexander merges from the GitHub UI** (like #4426). Not merged by an agent.


aalemayhu:
## Security review — credit pack purchase (payments/webhook, hard rail)

**Verdict: PASS — no blocking findings.** Focused on the two load-bearing points the spec named (checkout redirect allowlist, idempotent grant) plus the standard payments/webhook surface.

### Load-bearing correctness
- **Open redirect (CWE-601):** `success_url`/`cancel_url` are built by `resolveCreditPackRedirect(source, appUrl)` from a **server-side** `source` enum → fixed path map (`/upload`, `/account`). The client value passes through `parseCreditPackSource`, which returns only an allowlisted enum member or `undefined` (falls back to `/upload`). No client string reaches the URL. A controller test asserts `source: 'https://evil.example.com'` is dropped to `undefined`.
- **Idempotent grant (double-spend):** `insertPackGrant` uses `insert().onConflict('stripe_session_id').ignore().returning('id')` against the existing unique constraint; a redelivered webhook inserts zero rows and returns `false`. `credits_purchase_completed` fires **only** on a real insert. Covered by a redelivery test (inserts once, counts once) and a generated-SQL test asserting `on conflict ("stripe_session_id") do nothing`.

### Broader checklist
- **Credits/amount not trusted from client (CWE-20):** credits derive from the server constant `CREDIT_PACK_CREDITS`; `amount_total >= CREDIT_PACK_PRICE_CENTS` is a defense-in-depth sanity check. The trust anchor is server-set metadata (`credit_pack: '1'`, `user_id`).
- **Webhook signature (CWE-345):** unchanged — the branch runs inside the existing `constructEvent` HMAC-verified handler.
- **AuthZ (CWE-862):** route is behind `RequireAuthentication`; controller re-checks `res.locals.owner` and 401s without it. Signed-in only.
- **SQL injection (CWE-89):** query-builder bindings only; no raw interpolation.
- **Sensitive logging (CWE-532):** `credits.granted` logs `user_id` (internal numeric), `amount_credits`, `expires_at`, `inserted` — no session id, customer id, or token. Error path logs the error object, not the payload.
- **SSRF/new HTTP (CWE-918):** none — Stripe SDK wrapper only; client posts to a fixed relative path; `location.href` is set to the Stripe session URL our server created (not user input).
- **XSS (CWE-79):** confirmation copy is i18n text with a client-computed formatted date; no user HTML.
- **`?credits=added` spoof:** cosmetic only — the banner is param-driven but grants nothing client-side; the real (unchanged) balance shows on refetch. Same shape as the existing `?subscribed=1` banner. Not a vulnerability.

aalemayhu (commented):
Re-reviewed after rebase onto main (resolved a 10-locale + 1-test conflict against #4430/#4433, purely additive merges). Security and engineering forks both clean: redirect allowlist verified closed-union/no-client-URL, webhook grant verified idempotent via unique-constraint onConflict.ignore, credits/amount verified server-derived with an underpayment guard against session.amount_total, no hardcoded price id. Rebase conflict resolution independently verified correct (all 10 locale files valid with both key sets intact, test file has all 5 relevant tests with no drops). CI green, Sonar 0 findings on this exact SHA. Hard-rail (checkout+webhook+Stripe) — waiting for Alexander to merge from the UI.

PR #4430, opened 2026-09-16
feat: AI-credit usage visibility (account line + sync upload result)

aalemayhu:
## What
Two small additions that let paying users track AI-credit spend. On the account page, once spend exists this period the credits line reads "88 AI credits used this period. 212 left, valid through 7 October 2026." On a sync upload result, a deck that used AI now shows "This deck used N AI credits." A conversion with no AI use shows nothing extra.

## Why
Answers real user feedback from day 1 of the AI-credits system (#4426): a subscriber asked for a way to keep track of how many AI credits they used each time, so they'd know how much they have left next month. Scoped by a pm/designer/engineer trio as these two additions, not a historical usage-log dashboard.

## How
- `computeAiCreditBalance` already derived spend internally; it now exposes a rounded, floored `spent` field. `GetAiCreditsUseCase` returns it as `used`, and both the server and web `AiCreditsResponse` shapes gain `used` in the same change so it round-trips.
- `AiCreditsAccountLine` composes two independently-pluralized i18next fragments (used-this-period + remaining) only when `used > 0`; otherwise it falls through to the unchanged balance line.
- `recordClaudeUsage` now persists `request_id` on the `ai_usage_recorded` event (verified it survives `track()`'s PII stripper — it matches no PII key). The id is threaded through the PDF-vision and file-conversion Claude call sites, which previously carried only `userId`.
- New `AiUsageMetricsRepository.costByRequestId(userId, requestId, since)`, bounded by all three (the `since` bound limits cross-attribution because `requestId` is client-suppliable). The sync upload path sets an `X-Credits-Used` header from it; the web result reads it and renders the line.

## Measuring success
- Prod log line `[claude-usage] … request=<id> … cost_usd=<x>` already carries the request id; the new `request_id` prop lands on `ai_usage_recorded` rows, queryable via `select props->>'request_id', props->>'cost_usd' from events where name = 'ai_usage_recorded'`.
- `X-Credits-Used` header presence on `/api/upload/file` responses confirms the end-to-end read; absence when no AI ran confirms the gate.
- No new analytics event: this extends the already-tracked AI-credits feature rather than introducing a new surface, so no T+30d adoption issue is warranted (see Surface lifecycle note below).

## Testing
- Unit tests added: `balance.spent` (value + floor-at-zero); `GetAiCreditsUseCase.used` round-trip; web `getAiCredits`/`useAiCredits` `used` round-trip; `AiCreditsAccountLine` used>0, per-count pluralization, no-date fallback, and unchanged used===0 line; `recordClaudeUsage` request_id present/omitted; `AiUsageMetricsRepository.costByRequestId` generated-SQL (real pg-dialect knex); `UploadService` X-Credits-Used set / omitted-at-zero / fail-open-on-reject; `uploadResponse` header read; `UploadForm` renders / hides the credits line.
- Full server suite: only the documented environmental failures (Python-venv `PythonExitError` suites + `getPageCount` pdfinfo) fail in this worktree; every touched suite passes. Full web suite green. `/check` green; `pnpm format:check` clean tree-wide.

## Browser check
Notes: Verified via Playwright at 375px against a running `pnpm dev` — /upload and /account both load with zero console errors. The two new lines are additive renders gated on a numeric value being present and > 0 (already exercised by DOM-level component tests: `AiCreditsAccountLine.test.tsx`, `UploadForm.test.tsx`); actually triggering nonzero AI spend requires real backend usage data not available in this dev session.

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-16-see-how-many-ai-credits-a-deck-used-and-how-many-are-left.json` — "See how many AI credits a deck used, and how many you have left this period"

## Risks
- The displayed number is a **display estimate**, not a billing figure: `requestId` is client-suppliable (the middleware honors a UUID-shaped inbound `X-Request-Id`). Balance/spend enforcement stays entirely with the untouched `assertAiBudget`/`withAiBudget` guard; the `costByRequestId` lookup is bounded by `user_id` + a recent `created_at`, so the worst case is "your own number looks off if you deliberately reuse an id."
- The header read depends on the worker having flushed durable `ai_usage_recorded` rows before the sync result returns — `worker.ts`'s `finally` awaits `flushDurable()` before the worker resolves, so the rows are queryable by the time the main thread reads. If a future refactor moves the sync conversion off the worker, revisit.
- Fail-open by design: a metrics-read failure omits the header and never fails the conversion. Rollback is a straight revert; no migration, no schema change.

## Scope (explicitly not built)
- No historical per-conversion usage dashboard (trio-rejected as premature).
- No per-conversion cost on the async Downloads page — blocked by the pre-existing `uploads`-table notice-column gap documented in `src/lib/claude/FEATURE.md`; no migration added here.
- No changes to chat/photo-to-deck/tag-cards/note-type surfaces; no pre-conversion estimate.

## Goal alignment
Serves **simpler**: a paying user can answer "how many credits did that cost, and how many are left" without guessing. Metric: it should reduce credit-related support questions on the AI-credits feature; read from the support inbox over the coming weeks. This extends the already-tracked AI-credits capability on existing surfaces (account line, upload result), so it is not a new "surface" per the surface-lifecycle rule and does not warrant its own T+30d adoption issue.

## Decisions (trio synthesis)
- **PM**: ship the two small additions answering the real ask; do not build a usage-log dashboard (premature, speculative).
- **Designer**: compose the account line as two independently-pluralized fragments so both counts pluralize correctly; render nothing extra at `used === 0`; on the upload result show the line only when a deck actually used AI (no "0 credits used" noise, and the existing exhaustion warning takes precedence). Flagged that the server and web `AiCreditsResponse` are two independent hand-maintained shapes — edit both in the same commit or ship a silent `undefined`.
- **Engineer**: `spent` already exists inside `computeAiCreditBalance`; the honest per-conversion number requires threading `request_id` into the two Claude call sites that dropped it and a `since`-bounded lookup; fail open on the metrics read.
- **Agreement**: small, additive, no new surface; both response shapes edited together; fail-open.
- **Conflict / resolution**: none material. The only tension — whether the upload-result number is authoritative — resolved by framing it as a display estimate and leaving enforcement with the existing guard.

Closes the follow-up from #4426's day-1 feedback.


aalemayhu (commented):
Reviewed against security.md, code-quality.md, testing.md, and VOICE.md. The client-suppliable request_id risk this PR itself documents is correctly mitigated: costByRequestId scopes by user_id, event name, AND a created_at bound (conversionStart), not request_id alone (confirmed in the generated-SQL test). Both hand-maintained AiCreditsResponse interfaces (server + web) gained the used field together. New copy (account-page used+left line, sync-result deck-credits line) is VOICE-compliant with correct i18next _one/_other pluralization, gated on used/creditsUsed > 0 so a non-AI conversion shows nothing extra. Changelog entry is user-facing with no implementation-detail leaks. No blocking findings.

PR #4433, opened 2026-09-16
fix: use the house term Stapel instead of Deck in German credits copy

aalemayhu:
## What
The German AI-credits copy shipped in #4430 said "Deck" for a built deck; every other string in the same file uses the house term "Stapel." Fixes the one inconsistency.

## Why
Caught by the UX/voice review fork during #4430's ship — couldn't land in that PR because an unrelated agent had already enqueued/merged it before I could push the fix, so it ships as this tiny follow-up instead.

## Testing
JSON validated, `pnpm format:check` clean. No logic touched.

Browser check: not applicable — single-string locale copy fix, no rendered output changes beyond the string itself.


aalemayhu (commented):
Two-line German locale string fix, gender/term-consistency only (Deck to Stapel, matching every other string in this file). No logic, no security, no i18n structural change — the key/placeholder set is unchanged so parity.test.ts is unaffected. Reviewed directly against the diff, not delegated.

PR #4435, opened 2026-09-16
chore: self-heal upload rows whose bucket object is gone

### LINKED ISSUE #4401: chore: prune 147 upload rows whose bucket object the old sweep deleted

aalemayhu:
## What
Adds `deleteDeadUploadRowsInDatabase`, a recurring backstop that removes `uploads` rows whose bucket object no longer exists, wired into the existing daily `deleteOldUploads` sweep. It is the mirror image of `deleteDanglingUploadsInBucket` (which removes objects with no row); this handles the reverse case (a row with no object).

## Why
Closes #4401 (item 2 — the guard/nightly check follow-up to the #4419 prune ops command). The pre-#3881 orphan sweep deleted bucket objects without deleting their `uploads` rows, so affected owners still see those decks in `GET /api/upload/mine` even though the download 404s. #4419 gave Alexander a manual, on-demand prune command; this makes the class of bug self-heal on a schedule so it cannot silently recur. User-visible outcome: a deck whose file is gone stops appearing in your list (already handled reactively on download-click via the existing `deleteMissingFile` self-heal; this adds the proactive daily pass).

## How
Reuses what #4419/the storage-jobs layer already have rather than reinventing: one `storage.getContents(MAX_KEYS)` listing, `isReservedKey` (skips `mindmaps/`, `io-drafts/`, `assets/`), and the `assessDeletionVolume`/`raiseDeletionVolumeAlarm` anomaly guard shared with the sibling sweeps. It borrows the empty/paging-capped refusal from `PruneDeadUploadsUseCase`: if the listing came back empty or hit the 100k cap it deletes nothing (every row would look dead), and if the dead fraction is anomalous it alarms and aborts instead of mass-deleting. Runs inside the already-scheduled 24h `deleteOldUploads` job — no new timer, no per-request S3 on the hot `/api/upload/mine` path, no migration, no DB column.

Alternative considered: per-request object-existence checks in `GET /api/upload/mine`. Rejected — it adds S3 round trips (one HeadObject per row) to a hot listing endpoint on every page load. The recurring diff (the technique #4401 pointed at) is cheaper and cleans the DB rather than just hiding rows.

## Measuring success
Log line `Deleting dead upload row <id> (owner <owner>) — bucket object <key> is gone` in the daily sweep output confirms it fired and how many rows it cleaned (expect ~147 on the first prod run, then near-zero). A spurious mass-deletion is caught by the `[deletion-volume-alarm] deleteDeadUploadRowsInDatabase` error-event, and the empty/capped-listing refusal logs `refusing to judge` — grep these to confirm the guard behaved. Follow-up: `select count(*) from uploads` dead-row count trends to zero and stays there.

## Testing
- Unit tests added (`deleteDeadUploadRowsInDatabase.test.ts`, real in-memory sqlite, S3 mocked at the edge): deletes gone-object rows / keeps live + reserved-prefix rows; refuses on empty listing; refuses on paging-capped listing; alarms + deletes nothing on anomalous fraction; no alarm on normal volume.
- `deleteOldUploads.test.ts`: asserts the sweep invokes the new backstop with the shared db + storage.
- Full server suite: 743 suites / 7299 tests pass. The 9 failing suites (82 tests) are only the documented worktree-environmental failures — `PythonExitError` (create_deck venv absent) across the DeckParser/canary/golden suites, and `getPageCount` (`pdfinfo` binary absent). None touch storage/upload/sweep code.

## Browser check
Notes: Backend-only cleanup job; no UI surface. No browser-verifiable behavior (the effect is a scheduled DB cleanup). Nothing to render.

## Changelog
No changelog entry — internal-only. This is visibility hygiene: dead rows silently stop appearing (they should never have been visible), matching #4419 which shipped with no changelog for the same class of bug.

## Risks
It deletes `uploads` rows automatically. Mitigations: (1) it only deletes when the object is genuinely absent from a full, non-empty, non-capped listing; (2) empty/capped listings delete nothing; (3) an anomalous dead-fraction (Sonar-class thresholds: >=1000 rows or >=25% of the table) alarms and aborts; (4) the object is already gone, so no live download is broken. Rollback: revert the one-line addition in `deleteOldUploads.ts` (the helper is inert if never called).

## Goal alignment
Simpler — a broken deck that 404s no longer lingers in your list, so the deck list reflects what you can actually download. Metric: dead-row count in `uploads` (read via the #4419 prune dry-run / prod psql) trends to zero after the first daily sweep and stays there.


aalemayhu (commented):
Backend-only backstop sweep (no route/UI surface). Security fork: no knex.raw/interpolation, no path traversal (compares DB keys against an in-memory Set, never touches disk), no new env reads, no auth/webhook surface — clean. Engineering fork: tight 4-file diff, colocated test for the new function, no layer violations, no top-level timers, mirrors existing sibling sweep patterns (deleteDanglingUploadsInBucket) — clean. Serves simpler: a 404ing dead upload stops lingering in the user's list.

aalemayhu (commented):
Re-review on updated head (test-only Sonar fix since prior pass: .length).toBe(N) -> toHaveLength(N)). Security fork: knex builder bound queries only, no path traversal, no auth/env/webhook surface -- clean. Engineering fork: colocated tests, no layer violations, no top-level timers, mirrors existing sibling sweep pattern -- clean. Serves simpler: a 404ing dead upload stops lingering in the user's list.

PR #4436, opened 2026-09-16
feat: mark Image Occlusion notetype as Anki stock kind

### LINKED ISSUE #4406: feat: mark the Image Occlusion notetype as Anki stock kind so the mask editor accepts it

aalemayhu:
## What
Marks the Image Occlusion notetype that 2anki generates as Anki's built-in Image Occlusion stock kind (`originalStockKind: 6`), adds the stock field tags and `preventDeletion` flags, and renames it to `2anki Image Occlusion` so it no longer collides with Anki's own stock notetype.

## Why
Closes #4406. IO decks imported and reviewed fine, but Anki's mask editor refused to open them: its `io_notetype_if_valid` gate requires `originalStockKind == IMAGE_OCCLUSION` (6), which our notetype lacked — so users could not add/edit masks or use "Restore to Default". The notetype was also literally named `Image Occlusion`, so a user who already had Anki's stock IO notetype got a `+`-suffixed duplicate on import.

## How
- `src/templates/n2a-io.json`: `name` → `2anki Image Occlusion`, add `originalStockKind: 6`, and add `tag` 0–4 to the five fields with `preventDeletion: true` on the first four (Occlusion, Image, Header, Back Extra) — matching Anki's own stock IO field layout. Templates, CSS, and the occlusion-string format are untouched.
- `create_deck/helpers/n2a_model.py`: `N2AModel` gains an optional `original_stock_kind` and emits `originalStockKind` at the notetype level next to the existing `originalId` injection (genanki's `to_json` doesn't emit notetype-level extras). `tag`/`preventDeletion` ride through genanki unchanged, since `to_json` returns the field dicts as-is.
- `create_deck/helpers/get_model.py`: passes the JSON's `originalStockKind` through, so the JSON stays the single source of truth (non-IO notetypes have no such key and emit nothing).
- `create_deck/create_io_deck.py`: display name → `2anki Image Occlusion`, but the model id stays seeded from the legacy `Image Occlusion` name so a returning user's existing 2anki IO notetype is matched by id and updated in place on re-import (field count/order unchanged, so the schema stays compatible and cards are preserved).
- `src/services/__fixtures__/starters-baseline.json`: the IO starter's `noteType.name` tracks the JSON rename (the gallery display label stays "Image Occlusion").

Two valid seams existed for `originalStockKind` (an IO-only `Model` subclass vs. a data-driven key on the existing `N2AModel`). I chose the data-driven one: less code, keeps the "6" in one place (the JSON), and reuses the seam that already injects `originalId`.

## Trio note (lightweight, per CLAUDE.md)
This changes an existing feature, but it's a narrow, mechanical fix with no UI, copy, pricing, or flow change, so a full parallel trio dispatch would be ceremony. PM view: unblocks a broken promise (IO decks that can't be edited in Anki). Designer view: no surface change — the only user-visible artifact is the What's New line. Engineer view (this PR): safe because the rename isn't part of Anki's schema comparison and the field count/order is unchanged, so returning users' notetypes update in place rather than forking.

## Measuring success
- Durable regression guard: `create_deck/tests/test_anki_import.py` now imports a generated IO deck with the real Anki engine and asserts the imported notetype reports `originalStockKind == 6` with the stock field tags/preventDeletion, and coexists with Anki's stock `Image Occlusion` without a `+` duplicate. This runs in CI (`create_deck.yml`, Python 3.11/3.12) on every create_deck change.
- Post-deploy: IO deck generation should continue with no rise in `IO deck builder exited with code` errors in the conversion logs, and no new "can't edit masks in Anki" support reports. The mask editor opening is a client-side Anki action, so there is no server-side event that observes it directly — the CI real-Anki assertion is the authoritative signal that the shipped apkg is editable.

## Testing
- Unit tests added: `create_deck/tests/test_io_model.py` — asserts the generated model JSON carries `originalStockKind: 6`, the `2anki Image Occlusion` name, the legacy model id (in-place-update guarantee), field order, `tag` 0–4, and `preventDeletion` on the first four fields only; plus a guard that non-IO notetypes emit no `originalStockKind`.
- Real-Anki tests added: `create_deck/tests/test_anki_import.py` — IO deck builds via the production `create_io_deck.py` path, imports cleanly with real Anki, is editable (stock kind + tags), and does not collide with Anki's stock notetype.
- `src/services/starters.test.ts` — added an assertion that the IO starter's notetype is named distinctly from Anki stock; existing baseline-equality tests updated via the fixture.
- Full runs: create_deck pytest 146 passed (Python 3.12 + real Anki), pylint 10.00/10; server tsc clean; server Jest 7294 passed (the 82 failures are the documented worktree-environmental class — Python-bridge `PythonExitError` suites + `getPageCount` pdfinfo, no `create_deck/venv` in the worktree); web typecheck clean; web vitest 3702 passed; `format:check` clean tree-wide; oxlint clean (only pre-existing unrelated warnings).

## Browser check
Browser check: not applicable — the only `web/src/` change is a What's New changelog JSON; no component or rendered-UI change (the attestation hook's changelog-only bypass applies).

## Changelog
`web/src/pages/WhatsNewPage/changelog/2026-09-16-io-mask-editor.json` — "Image Occlusion decks open in Anki's built-in mask editor, so you can add and edit masks after importing"

## Risks
- Low. Field count/order is unchanged and the model id is preserved, so returning users' notetypes update in place rather than forking; cards map by ordinal, so no card breakage. The rename is cosmetic to Anki's schema comparison.
- Rollback: revert the PR. Any decks already generated with the new notetype remain valid Anki decks; reverting only stops setting `originalStockKind` on future exports.

## Goal alignment
Serves **more beautiful** (and a bit of **simpler**): IO decks made on 2anki.net now behave exactly like Anki's native ones — the mask editor and "Restore to Default" work — instead of being read-only imports. Metric: absence of new "can't edit masks" support reports and a steady/again-healthy IO deck generation rate in the conversion logs; the mask editor action itself is client-side in Anki and not server-observable.


aalemayhu (commented):
Python notetype-generation change plus one changelog entry, no HTTP/user-input surface. Security fork: no auth/secrets/HTML/path/knex touched — clean. Engineering fork: tight diff, colocated tests at both unit (test_io_model.py) and real-Anki-import level for every behavior (stock-kind emission, field tags, name distinctness, legacy-id continuity) — clean, two low-confidence style notes not worth blocking on. Voice fork: changelog entry sentence-case, no banned words, correct brand capitalization — clean. Serves more beautiful: Image Occlusion decks become editable in Anki's own mask editor after import, retroactively for existing users via the preserved legacy model id.

PR #4438, opened 2026-09-16
fix: show purchased AI credits when the plan has lapsed

aalemayhu:
## What
When a user's plan lapses, `computeAiCreditBalance` returned `null` the moment `resolveAllowance` found no active plan — so an unexpired credit-pack grant (`ai_credit_grants`) was invisible on the account page. It now builds a **grant-only balance** in that case (`allowance: 0`, credits/spend computed against the grant, the grant's own earliest-anchor→latest-expiry window, `resets: 'pass'`), and the account page renders a muted **paused** line pointing at resubscribe. Display-only: a lapsed user still can't invoke paid AI conversion.

## Why
Closes #4434. A user who bought a $5 / 250-credit pack and then let their plan lapse saw nothing on the account page — the credits looked lost. They aren't: they become spendable again automatically the moment the user resubscribes or buys a pass.

## How
- `computeAiCreditBalance` (`aiCredits/balance.ts`): on a null allowance, read `sumActiveCredits`; if `0`, still return null; if `> 0`, build a grant-only balance via a new `activeGrantWindow` reader.
- New `AiCreditGrantsRepository.activeGrantWindow(userId, now)` → `{ windowStart, windowEnd } | null` = `min(created_at)` / `max(expires_at)` across unexpired grants (generated-SQL test added). Wired through `AiCreditBalanceReaders` and `createAiCreditReaders`.
- `GET /api/ai/credits` now carries `usable` (`allowance > 0`) so the client tells the paused state from an active balance. `AiCreditsAccountLine` gains a paused branch (no `BuyCreditsButton` — selling more inert credits while none are usable is a trap); mounted in `PlanDetails`' free/lapsed branch, which already carries the "See plans" resubscribe link.
- **Did NOT touch** `src/lib/isPaying.ts` or any `isPaying`/`hasAnkifyAccess`-gated controller/use case. Not a hard-rail diff.
- Reader/field shapes landed: `GrantWindow { windowStart: Date; windowEnd: Date }` (balance.ts) / `AiCreditGrantWindow` (repo); response field `usable: boolean`; `resets` reuses `'pass'` (no new `CreditWindowReset` member, so no needless 10-locale chip label). The paused UI never falls through to the generic `resets` chip — it renders its own bespoke copy.

## Known behavior change (accepted, documented — do not "fix")
[...]

PR #4439, opened 2026-09-16
feat: show AI credit balance in the sidebar

aalemayhu:
## Summary
Subscribers, pass holders, and lapsed users with an unexpired credit pack could only see their AI credit balance on `/account` or `/upload`. This adds a compact line to the always-visible sidebar identity block (below the free-tier card-usage counter) so the balance is visible on every logged-in page.

States (mirrors `AiCreditsAccountLine`'s real predicate — `usable = allowance > 0`):
- Normal balance: `{count} AI credits left.`
- Low (≤25) or zero: warning color + a `Buy credits` link to `/account`.
- Paused (lapsed plan, unexpired pack): `{count} AI credits, paused. Resubscribe to use it before {date}.` — no buy link, matching the Account page's decision not to sell into a state that isn't actively usable.
- No allowance and no pack: nothing rendered (unchanged from today).
- Collapsed sidebar: hidden for free — `.identity` is already `display: none` when collapsed.

## Decisions
Ran the pm/designer/engineer trio (required — user-facing sidebar change):
- **Placement**: identity block, directly under the card-usage counter. Agreed by all three.
- **Reuse**: `useAiCredits` hook, existing `aicredits` i18n keys (all 10 locales already have `left`/`zero`/`paused`/`buyShort` — no new translation strings), existing `identityUsage*` CSS classes.
- **Conflict resolved**: PM's original metric wanted an embedded `BuyCreditsButton` (source: credits_sidebar), which touches `src/usecases/checkout/creditPack.ts` — hard-rail. Went with the designer's plain `Link` to `/account` instead: smaller blast radius, ships without a hard-rail wait, and matches the identity block's one existing CTA pattern (the free-tier "Upgrade for unlimited" link). New lightweight `credits_sidebar_link_clicked` event added to both `KNOWN_EVENTS` allowlists in this PR (parity test enforced).
- **Correction during implementation**: the designer's review ran against a stale local `main` and reported no `usable` field existed on the response. It does (added by #4438, already merged) — `usable = allowance > 0`. Used the real field and mirrored `AiCreditsAccountLine`'s exact paused/zero/normal branches rather than a simplified guess, so the sidebar and account page never disagree about state.
- First attempt relocated `formatLongDate` from `pages/AccountPage/utils/` to `lib/`, but that touched `SubscriptionManagement.tsx`/`useSubscriptionCancellation.ts` for a one-line import path, tripping the hard-rail `subscription` keyword for no real reason. Reverted — imports it cross-page into `Sidebar.tsx` instead, keeping the diff to only the files the feature actually needs.

Not a new product surface — AI credits already shipped (account line #4430, purchase #4429); this is a visibility placement change, so no T+30d adoption issue. Signal to watch: `credits_sidebar_link_clicked` → `/account` → `credits_purchase_completed` at T+7; if the sidebar source drives zero clicks, remove the line.

## Review fan-out (security / engineering / ux-voice)
Security: clean, no findings. Engineering and ux-voice each found one fixable issue, both fixed on-branch and re-verified:
- **ux-voice**: the card-usage line and the new AI-credits line read as one fused block when both render (free user with a pack). Fixed by wrapping each state in one outer element (mirroring `CardUsageCounter`'s existing shape) and adding `margin-top` via `.identityUsage + .identityAiCredits` so the gap only appears when the AI-credits line directly follows the card counter. Two smaller ux-voice notes (paused-state has no CTA; the reused warning color is under AA contrast for a full sentence, not just a count) were accepted as non-blocking follow-ups — both are pre-existing patterns this PR extends rather than introduces.
- **engineering**: the "no plan" test mocked `getAiCredits` resolving `null`, but a real free user never gets `null` from `GetAiCreditsUseCase` — they get `{credits:0, allowance:0, usable:false}`, which drives a different, previously-untested branch. Added a test with that exact shape.

## Also in this PR
Added `staleTime: 30_000` to `useAiCredits`'s query — the sidebar keeps this query mounted on every logged-in page, so without it a refetch-on-focus default would hit `/api/ai/credits` on every window refocus. Safe because every spend site already invalidates the query key directly after a spend.

## Test plan
- [x] Full web suite: 414/414 files, 3717/3717 tests passing
- [x] `tsc --noEmit` (web): clean
- [x] `tsc -p .` (server, for the AnalyticsEvents.ts allowlist add): clean
- [x] `pnpm --filter 2anki-web lint`: clean (no new warnings)
- [x] `pnpm format:check`: clean

## Browser check
Notes: Ran `pnpm --filter 2anki-web test:golden-path` — 4/4 passed at 375px, zero console/page errors, including the signed-in dashboard and both account-page states (which render the sidebar, and therefore the new `AiCreditsSidebarLine`, on every run). The mocked backend's catch-all response doesn't shape-match `AiCreditsResponse`, so these runs exercise the null-credits branch; the paused/low/zero/normal visual states are covered by `Sidebar.test.tsx`'s jsdom suite instead (7 cases, includes the new margin-separation markup and the realistic zero-allowance shape). No dedicated golden-path fixture for the other states — flagging per `.claude/docs/browser-attestation.md`'s honesty requirement rather than overclaiming coverage.


aalemayhu (commented):
Security/engineering/ux-voice fan-out review: clean. Two rounds of findings fixed on-branch — (1) trimmed scope to avoid an accidental hard-rail hit from an unrelated file move, (2) added margin separation between the card-usage and AI-credits lines in the sidebar identity block, and (3) added a test for the real free-user zero-allowance response shape (not the null shortcut). All three review dimensions re-verified clean against the final head SHA. Full web suite green (414/414 files, 3717/3717 tests), typecheck/lint/format clean, not hard-rail.

PR #4444, opened 2026-09-17
fix: prompt sign-in instead of a generic error on an expired session

### LINKED ISSUE #4442: Favorite toggle shows generic error instead of login prompt on 401
### Problem
Favoriting from `/rules/:id` or `/preview/:id` (reachable via the gear/settings link on `PreviewPage`) shows a generic "Failed to update favorite" error instead of a login prompt when the session is expired or invalid.

`Backend.addFavorite`/`deleteFavorite` POST `/api/favorite/create`/`remove`, gated by `RequireAuthentication`, which returns `401 {message:'Authentication required'}` when unauthenticated. Client-side, a 401 triggers `redirectToLogin()` but the function still returns `false`, identical to any generic failure. `toggleFavorite` then throws an untagged `Error` (no `.status`), which skips `reportClientError`'s 4xx filter and both toasts a generic message and logs a false "crash" to `/ops/errors`.

This is the same bug class already fixed for search (see the comment at `Backend.ts:242-245`), never applied to favorites. A second, likely-related occurrence: `CardOptionsForm`'s reset action (`cardOptions.resetError`) hits the same shape — an authenticated API call (`resetUserCardOptions()`) failing on a stale/expired session shows a generic "try again" message instead of prompting re-login.

Note: production error samples tag both occurrences `User: anonymous`, but that tag isn't reliable evidence the visitor lacked an account — `reportClientError` never sends a `userId`, so every web-originated error shows as anonymous regardless of true login state. The real trigger is most likely a session that expired between page load and the click, not a logged-out visitor clicking a favorite star.

### Severity
Low-moderate. Doesn't block any real flow (retry after re-login works), but pollutes the ops error dashboard with false crash reports and gives users a confusing message when the actual issue is "please log in again."

### Proposed direction
Tag the 401 with its status (mirroring the `Backend.ts:242-245` search fix) and surface a distinct "Log in to save favorites" notice instead of routing it through the generic crash-reporting path. Apply the same pattern to `CardOptionsForm`'s reset error if convenient while in the area.

aalemayhu:
## Problem

Favoriting a page (/rules/:id, /preview/:id) and resetting card options both hit RequireAuthentication's 401 on an expired/invalid session. Neither call site tagged the failure with its status, so:

- The client showed a generic "Failed to update favorite" / reset error instead of a login prompt.
- The untagged error skipped reportClientError's expected-client-fault filter and logged a false "crash" to /ops/errors.

Same bug class already fixed for search (Backend.ts searchFailure), never applied to favorites or the card-options reset.

## Fix

- Backend.addFavorite / deleteFavorite / resetUserCardOptions now throw a tagged status: 401 error (and call redirectToLogin()) instead of silently returning false / discarding the response status - mirroring searchFailure.
- CardOptionsForm's reset handler no longer discards the caught error; it passes a tagged 401 straight through (so classifyError renders "Session expired... Sign in again") and only falls back to the generic reset-failed message for a real failure.

## Browser check
Notes: Signed in, hit /card-options and clicked "Reset to defaults" (the exact `resetUserCardOptions()` path this PR changes) and loaded /favorites -- zero console errors on either. The 401-specific path itself needs an actually-expired session to trigger live; that branch is covered by the unit tests instead.

## Testing

- Scoped pnpm test runs green (Backend.test.ts, CardOptionsForm.test.tsx).
- Full /check (server tsc + arch, web typecheck, web vitest, web lint) green.

Fixes #4442


aalemayhu (commented):
**Verdict**: approve

Small, well-scoped mirror of the existing `searchFailure` pattern applied to favorites and card-options reset — serves "simpler/faster" by turning a silent generic-failure UX into a correct login prompt, and stops false crash reports in /ops/errors.

### Blocking
None.

### Nits
- `authRequiredError` is a nice small private helper; consider co-locating it right above `resetUserCardOptions` (its first use) rather than between `convert` and `addFavorite` — pure readability, not a merge blocker.

Security: no new HTTP surface, no user-controlled URL, no HTML rendering change. Clean.
Testing: new 401 branch and the non-auth fallback are both covered (Backend.test.ts, CardOptionsForm.test.tsx). Clean.
Voice: changelog entry is sentence case, no trailing period, no implementation-detail leak, under 120 chars. Clean.

aalemayhu (commented):
Reviewed: mirrors the existing searchFailure auth-tagging pattern, small and well-tested, no blocking findings.

aalemayhu (commented):
Reviewed against security.md, code-quality.md, testing.md, and VOICE.md (inline review — fork tooling unavailable this pass). Clean: no knex/HTTP/HTML/path-traversal surface touched, auth check stays server-side in RequireAuthentication middleware, the new 401 tagging mirrors the existing searchFailure pattern exactly, both new call sites get test coverage (Backend.test.ts, CardOptionsForm.test.tsx), and the changelog entry follows VOICE.md (sentence case, no trailing period, no implementation-detail leak). No blocking findings.

aalemayhu (commented):
Re-reviewed at current HEAD after two follow-up commits (optional-chaining nit fix, unrelated FEATURE.md line dropped). Diff is clean: small, well-tested mirror of the existing searchFailure pattern, no security/layer/voice issues.

aalemayhu (commented):
Security fork: no issues (fixed literal auth message, no injected/echoed error text). Engineering: no layer violations, no any/as any, comments explain why not what, tests cover the new tagging behavior at both call sites and the CardOptionsForm fallback branch, scope stays tight to the 401 handling fix. Voice: changelog entry is sentence case, no trailing period, no banned words, describes the new behavior directly. Clean.

PR #4446, opened 2026-09-17
fix: stop chat-deck saves racing and leaking a path on failure

### LINKED ISSUE #4443: Chat-deck save race lets users retry blind, ENOENT leaks internal path
### Problem
Four production `ENOENT: no such file or directory, open '<deck>.apkg'` errors, same user, same minute, four different filenames — `ChatDeckUseCase.execute` → `CustomExporter.save()` reading a generated `.apkg` that isn't there.

Root cause: `CustomExporter.save()` (`src/lib/parser/exporters/CustomExporter.ts:75-76`) does `apkgPath = await gen.run(); return fs.promises.readFile(apkgPath);` with zero existence check or retry — it fully trusts the path the Python subprocess printed on close.

The actual trigger is a UI race, not a filesystem flake: `commitSave()` in `web/src/pages/Chat/CardPreview.tsx:178-185` flips the UI to "Saved as X" **before** the fire-and-forget `onSave()` → `downloadDeck()` (`ChatPanel.tsx:935-940,155`) resolves, and immediately exposes a "Save again" button (`CardPreview.tsx:281-293`) with no loading/disabled gating. A user who doesn't see the download land clicks "Save again" repeatedly, firing independent generation attempts (each its own workspace/subprocess) — explaining the four isolated attempts in under a minute, one of which hit a transient tmp-fs read failure.

Confirmed NOT the cause: same-day PR #4423 (unrelated files, verified via `git show --stat`), a deploy restart (no deploy ran anywhere near that window), or the filesystem cleanup job (only sweeps `WORKSPACE_BASE`/`UPLOAD_BASE`, never the `chat-deck-<uuid>` dirs, and its age floor is 2 hours vs. this 1-minute window).

Secondary issue: `ENOENT` doesn't match `isInternalFault` in `ErrorHandler.tsx:159-172` (SQLSTATE/Knex/TypeError checks), so it falls through to the generic branch and returns a `400 {code:'unknown', message: err.message}` — leaking the raw internal tmp file path to the client and mislabeling a server-side bug as a client fault.

### Severity
Moderate. Not silent (retrying does work, since each attempt is independent), but the premature "saved" confirmation misleads users into retrying blind, and the error response leaks an internal filesystem path.

### Proposed direction
Don't flip to the "saved" state until `onSave()` actually resolves, and disable Save/Save-again while a save is in flight. Add an existence check (or typed retry) around the read in `CustomExporter.save()`. Route filesystem errors through the 500 branch in `ErrorHandler.tsx` instead of echoing `err.message` as a 400.

aalemayhu:
## Problem

Four production ENOENT errors, same user, same minute, four different filenames - ChatDeckUseCase.execute -> CustomExporter.save() reading a generated .apkg that wasn't there.

Root cause was a UI race, not a filesystem flake: CardPreview.tsx's commitSave() flipped the UI to "Saved as X" and exposed a "Save again" button BEFORE the actual save request (onSave -> downloadDeck -> POST /api/chat/deck) resolved. A user who didn't see the download land clicked "Save again" repeatedly, firing independent generation attempts - one of which hit a transient tmp-fs read failure.

Secondary issue: CustomExporter.save() trusted the Python subprocess's printed .apkg path with no existence check, and ErrorHandler's isInternalFault classifier didn't recognize real Node filesystem errors (ENOENT etc., which always carry `code` + `errno`), so the ENOENT fell through to the generic 400 branch and echoed the raw internal tmp path back to the client.

## Fix

- `CardPreview.tsx`: commitSave() is now async - it disables Save/Cancel (`aria-busy` + a CSS-only spinner on Save, no new copy/i18n) while the save is in flight, and only flips to "saved" after the request actually resolves. A rejected save keeps the naming row open instead of showing a false "saved" state; the existing `errors.deckGenerate` banner in ChatPanel already surfaces the failure.
- `ChatPanel.tsx`: `handleSaveAsDeck` now returns the save promise (and rethrows on failure) instead of firing it fire-and-forget, so CardPreview can await it.
- `CustomExporter.save()`: waits briefly (up to ~350ms) for the generated apkg to actually appear on disk before reading it, instead of trusting the printed path blindly.
- `ErrorHandler.tsx`: `isInternalFault` now recognizes real Node fs/OS errors (anything with both a string `code` and numeric `errno`) as our bug, not the client's, so they route through the 500 branch (generic message) instead of a 400 that echoes the raw error text.

- Scope: fixed the race (root cause) and hardened both backend layers it exposed (existence-check + error classification) in one PR since they're the same bug's three symptoms. Did NOT add click-throttling telemetry or a toast confirming download start - a pm review pass on this fix judged the existing `errors.deckGenerate` banner sufficient and flagged that as out of scope for a bug fix.
- UX: a designer review pass recommended visible in-flight feedback (not just a disabled button, which reads as "broken") - added `aria-busy` + a CSS-only spinner on the Save button, mirroring the existing `cardPreviewTagSkeleton` in-progress pattern in this file. No new copy, no i18n touch.
- Left the generic ENOENT->500 classification broad (any error with `code` + `errno`) rather than a narrow allowlist, since Node's own fs/net errors always carry both and no deliberate application error in this codebase sets `errno`.

## Testing

- New/updated tests: `CardPreview.test.tsx` (save-in-flight disabling, aria-busy, reject-keeps-naming-open), `CustomExporter.test.ts` (reads back the reported apkg, waits for a slow-to-appear file, rejects with a real ENOENT when it never appears), `ErrorHandler.test.ts` (fs ENOENT responds 500 without leaking the path).
- Full server Jest suite green except the documented environmental `getPageCount`/pdfinfo failures (no `pdfinfo` binary in this environment).
- Full web vitest suite green (414/414 files, 3719/3719 tests).
- `/check` (server tsc + arch, web typecheck, web lint) green.

## Browser check

Not run - no dev-server/browser access in this environment. Covered by unit tests exercising the real async state machine (RTL `fireEvent` + `waitFor`), including the in-flight disabled/aria-busy state and the reject path.

Fixes #4443


aalemayhu (commented):
**Verdict**: approve

Fixes the actual UI race (premature "saved" state) at the root, then hardens the two backend layers the race exposed (existence-check before read, real fs-error classification). Serves "faster/more trustworthy" - stops duplicate generation attempts and stops leaking an internal path.

### Blocking
- Found and fixed during this review pass: the changelog title used "no longer glitch," a prior-brokenness hedge phrase .claude/docs/changelog.md explicitly bans (mirrors its own "no longer broken" example). Rewritten to "Save again on a chat deck waits until the first download finishes" in a follow-up commit on this branch.

### Nits
None.

Security: no new HTTP surface, no path traversal (apkgPath is the CardGenerator subprocess's own reported path, not user input), no leaked internals in the client-facing 500 body. Clean.
Testing: race (in-flight disable + reject-path), CustomExporter.save (happy/slow/missing), and the fs-error ErrorHandler classification are all covered with real assertions, not just "did not throw". Clean.

aalemayhu (commented):
Reviewed: root-caused UI race fixed, backend hardening consistent with existing patterns, changelog voice issue caught and fixed in-branch, no blocking findings remain.

aalemayhu (commented):
Reviewed against security.md, code-quality.md, testing.md, and VOICE.md/changelog.md. No security issues — the fs read/retry path uses a path already derived from CardGenerator's own trusted output, and the ErrorHandler change is itself a defensive fix (stops leaking a raw internal path). Engineering: the new setTimeout lives inside an async per-call function (readGeneratedApkg), not module scope; CardGenerator mock is at the legitimate external-process boundary; comments explain WHY not WHAT; every new function has direct test coverage. One observation, not blocking: isInternalFault's new code+errno check is intentionally broad (any Node fs/OS error) rather than an ENOENT allowlist — documented in-code as deliberate, seems right since no legitimate app error sets errno. UX: changelog entry is clean (sentence case, no trailing period, no implementation leaks, no hedge phrasing, under 120 chars). The aria-busy + CSS spinner respects prefers-reduced-motion and keeps the button's accessible name intact while busy.

aalemayhu (commented):
Security: no knex/axios/URL/HTML-render/env/JWT/zip-path surface touched; the fs errno-classification change only affects which HTTP status an already-internal error maps to, never echoes attacker-controlled text. Engineering: race fix is scoped to the async save flow (isSaving guard mirrors the existing toggleFavorite pattern), retry helper is small and well-commented, new tests exercise both the guard and the reject path. Voice: no new user-visible copy beyond the changelog entry (sentence case, no hedge words, no implementation leak) and the disabled/aria-busy state carries no text needing i18n. Clean.

aalemayhu (commented):
Re-reviewed after the SKIP_CREATE_DECK CI-environment fix (the new save() tests short-circuited into the deck_info.json read path in CI where the flag is set tree-wide). Test-only change, isolates the flag for this describe block. No new findings.

Issue #4445, opened 2026-09-17
feat: note-level content-hash mod for signed-in re-imports (card_guids ledger)


aalemayhu:
Split from #4407. Option 3 (notetype mod constant) shipped separately.

## Problem
Re-uploading the same deck overwrites note edits users made inside
Anki, because every genanki note is stamped with export time and
Anki updates any guid-matched note whose incoming mod is newer.
22% of owners re-upload the same deck name -- a paying-user path.

## Scope (signed-in, eligible-format uploads only)
Anonymous uploads have no ledger and keep current always-newer
behaviour.

- Migration: add content_hash + content_changed_at to card_guids;
  run kanel in the same PR (hard rail -- needs migration-reviewer).
- Note mod = content_changed_at (time the card's content hash last
  changed), seeded to now the first time a guid is seen under the
  new scheme. First re-upload after deploy updates once, then stays
  quiet until content changes.
- Update CardGuidLedgerRepository + uploadCardIdentity.ts to write
  and read the hash.
- Fold in the duplicate-guid case: 5 of 185 sampled decks held a
  guid pair with identical front, different back; second note wins
  on re-import. Detect and disambiguate via content_hash.

## Riskiest assumption
Seeding content_changed_at correctly on the first post-deploy pass
does not skip parser-improvement updates for returning users.
Smallest test: re-import a fixture deck twice, assert edited notes
survive and content-changed notes update.

## Tier
Tier 3 draft PR, migration-reviewer required, kanel regen in-PR.

PR #4447, opened 2026-09-17
fix: pin notetype mod to a constant so re-uploads keep local edits

### LINKED ISSUE #4407: design: note and notetype mod policy for re-imports
## Problem

Anki updates a guid-matched note only when the incoming `mod` is newer than the local one (`rslib/src/import_export/package/apkg/import/notes.rs:521`). genanki stamps every note and notetype with export time (`genanki/note.py:159`, `genanki/model.py:116`), and a prod sample on 2026-09-11 confirmed a `mod` spread of 0 inside every file. Consequences:

- Every re-upload overwrites edits the user made to those notes inside Anki.
- Every re-upload overwrites the user's CSS or template edits on the notetype.
- Field/template ids and the Image Occlusion metadata only reach returning users because of this "always newer" behaviour, so changing the policy must not break that.

22% of owners re-upload the same deck name, so this is a paying-user path.

## Options considered

1. Note `mod` = Notion `last_edited_time`. Rejected for returning users: their local `mod` is a past export time, so an older page edit date makes Anki skip every note, and parser improvements stop reaching them.
2. Note `mod` = time the card's content hash last changed, seeded to `now` the first time a guid is seen under the new scheme. First re-upload after deploy updates once, then stays quiet until content changes. Needs `content_hash` and `content_changed_at` on `card_guids` (migration, hard rail). Anonymous uploads have no ledger and keep current behaviour.
3. Notetype `mod` = constant chosen at deploy time, bumped on every template change. A user who exported minutes before deploy keeps a newer `mod` until the next bump; acceptable.

Recommendation: 2 and 3, shipped after https://github.com/2anki/server/issues/4404, https://github.com/2anki/server/issues/4405 and https://github.com/2anki/server/issues/4406.

## Related

Duplicate guids inside one file: 5 of 185 sampled decks held one pair of notes with the same guid and different content (identical front, different back). Documented v1 limitation of the upload identity scheme; the second note wins on re-import. Worth folding into the content-hash work.

aalemayhu:
## Problem

genanki stamps every notetype with the export timestamp, and Anki only overwrites a notetype already in a user's collection when the incoming mod is newer than the local one. That meant every re-upload of a deck looked "newer" than whatever the user had, silently overwriting any CSS or template edit they'd made inside Anki. 22% of owners re-upload the same deck name at some point, so this is a real paying-user path.

## Fix

`N2AModel.to_json` (create_deck/helpers/n2a_model.py) now pins the notetype's `mod` to a frozen module-level constant (`NOTETYPE_MOD`) instead of the export timestamp. Anki no longer sees a same-version re-export as "newer," so it leaves the user's local CSS/template edits alone. The constant gets bumped only when a shipped template or CSS actually changes -- that's what pushes a real 2anki-side fix to returning users, same mechanism as before, just opt-in instead of automatic-on-every-export.

Note-level `mod` is untouched by this PR -- notes still always overwrite on re-upload. That's option 2 of the original design (needs a `card_guids` migration to track a content hash), split out to #4445 since it's too large to build to a trustworthy state in one unattended slot.

- **Scope**: shipped option 3 (notetype mod) only; deferred option 2 (note mod) to #4445. Trio call (pm + engineer): option 3 stands alone, has no schema change, and is the smaller/safer half; option 2's migration + ledger-seeding logic is exactly what Tier 3 + migration-reviewer exist to slow down, not something to rush overnight.
- **The constant's value** (`1789650998`) is the Unix timestamp at the moment this line was written. Riskiest assumption (engineer's flag): it must be >= every notetype mod value already sitting in a returning user's collection, which holds by construction -- every prior real export happened at some point in the past, so any timestamp captured "now" or later at write time is >= all of them. No prod-data check needed for this to hold.
- **UX**: designer's call was silent-and-correct (no new UI, no banner) since this repo's "precise, trustworthy" register means doing the right thing quietly, not announcing every re-upload's outcome. Did add one changelog line since 22% of owners will feel the behavior change on their next re-upload.
- **Found and fixed a real gap while verifying**: `create_deck/tests/test_anki_import.py`'s existing `test_returning_user_reimport_writes_field_ids_in_place` simulated "an old collection without field ids" by stripping ids from a *freshly built* apkg, relying on wall-clock time always making a later export's mod strictly newer. With mod now frozen, that simulation no longer produces a genuinely-older local copy, so the test would have started asserting the wrong thing (and could have started failing in CI's anki-matrix rows, which is where this only gets exercised -- it importorskips locally without the `anki` pip package). Backdated the simulated collection's mod in `_strip_notetype_ids` to actually predate the constant. Verified this catches the real regression: built a throwaway Python 3.12 venv with the real `anki==26.8.1` package (per `requirements.txt`'s version pin) and ran the full `test_anki_import.py` suite against real Anki import/re-import logic -- all 7 tests pass, including a new `test_reexport_does_not_clobber_a_users_local_css_edit` that I confirmed fails without this fix (reverted `n2a_model.py` locally, watched it fail with the user's CSS edit overwritten, then restored the fix) and passes with it.

## Testing

- `create_deck` full pytest suite: 142 passed, 1 skipped (environment-gated, no `anki` package in the default venv).
- Real-Anki verification (throwaway Python 3.12 venv with `anki==26.8.1` installed per `requirements.txt`'s pin): all 7 tests in `test_anki_import.py` pass, including the new CSS-clobber regression test, confirmed red-then-green against the actual fix.
- No TS/web files touched by this PR -- pure `create_deck/` Python change plus a `FEATURE.md` doc update and one changelog entry.

## Browser check

Not applicable -- no `web/src/` changes; this is a backend/export-path fix with no runtime-visible UI surface beyond the What's New entry.

Fixes #4407

Split follow-up (option 2, needs a migration): https://github.com/2anki/2anki.net/issues/4445


aalemayhu (commented):
Reviewed against security.md and code-quality.md (inline review). Pure Python change in create_deck/, no HTTP/DB/env-flag/auth surface touched. The frozen NOTETYPE_MOD constant is tested for both the override itself and stability across differing export timestamps, and the returning-user regression fixture was updated (with a documented reason) to keep proving the id-pinning behavior this change sits next to. Full create_deck pytest suite green (142 passed). No blocking findings.

aalemayhu (commented):
Reviewed against code-quality.md/testing.md (Python spirit — no knex/layer-violation checks apply) and the changelog voice guide. NOTETYPE_MOD is covered by direct unit tests plus a real-Anki regression test (anki==26.8.1, verified red-then-green against the fix in a throwaway venv). The _strip_notetype_ids backdating fix is a genuine, well-justified correction, not a hack. Changelog entry is clean (positive framing, no hedge words, no implementation leak). No security surface — pure Python export-path change, no networking/auth/DB. Clean.

aalemayhu:
Independently re-verified locally: built a throwaway Python 3.12 venv, installed anki==26.8.1 + genanki==0.13.1 per requirements.txt's pin, ran tests/test_anki_import.py against real Anki (7/7 pass). Also reproduced the regression directly — swapped in the pre-fix n2a_model.py, confirmed test_reexport_does_not_clobber_a_users_local_css_edit fails (the re-upload overwrites the CSS edit, exactly as described), restored the fix, confirmed green again. Venv discarded after.

aalemayhu (commented):
Reviewed against security.md, code-quality.md, testing.md, and changelog.md conventions. No security concerns (pure Python, no user input, no auth/DB/network paths touched). Engineering: comments explain WHY not WHAT, new constant has direct test coverage, no scope creep, pylint 10.00/10 on both changed helpers/ files. Fixed one changelog wording nit (hedge phrasing implying prior brokenness) before this review. The core fix was red-green verified against a real anki==26.8.1 install (not just mocked/unit-level), including a new regression test that reproduces the exact CSS-clobber bug and confirms the fix resolves it.

aalemayhu (commented):
**Verdict**: approve

Root-causes the actual bug (notetype mod = export timestamp = "always newer") with a minimal, correctly-scoped fix, and goes further than most PRs to prove it: a real Anki-collection regression test (`anki==26.8.1` in a throwaway venv, confirmed red-without-the-fix / green-with-it), not just a mocked-JSON assertion. The `_strip_notetype_ids` backdate fix in the existing returning-user test is the sharpest catch here -- without it, the frozen constant would have silently made that test's "old collection" indistinguishable from a fresh one, masking exactly the class of regression this PR fixes.

### Blocking
None.

### Nits
None -- the "Decisions made overnight" section already documents the riskiest assumption (constant >= any past export time, true by construction) and why option 2 was split to #4445 rather than compressed into this slot.

Security: N/A, pure Python constant + test change, no user input involved.
Testing: real Anki import/re-import proves the fix; existing coverage updated correctly so it keeps testing what it claims to.
Voice: changelog title is clean (no prior-brokenness hedge), matches VOICE.md conventions.

aalemayhu (commented):
Reviewed: root cause correctly identified and fixed, real-Anki regression test proves it, existing test updated to keep testing the right thing, changelog voice clean, no blocking findings.

PR #4448, opened 2026-09-17
fix: never ship an input card with nothing to type against

### LINKED ISSUE #4440: Notion-typed bold (**word**) doesn't reliably produce a usable Input card
### Problem
A user's Input-type note came out with a blank `Input` field, so any typed answer was marked wrong in real Anki, even though the front card rendered correctly. The word was made bold via Notion's typed markdown shorthand (`**word**`) rather than the Cmd+B toolbar.

Investigation of the live-Notion-toggle import path (`getInputCard.ts`) found no stripping bug — it reads Notion's own `annotations.bold` boolean directly, and the strict `enableInput && answer` AND-gate in `create_deck.py` should fall back to a plain Basic card rather than ship an Input card with a blank field. So the reported symptom doesn't cleanly match what that path can currently produce. Two live possibilities: the export went through a different path (HTML/zip upload has separate, older bold-detection logic in `DeckParser.ts`), or Notion's `**bold**` autoformat silently failed to trigger for this specific toggle title (leaving literal asterisks with `annotations.bold: false`). Root cause is unconfirmed.

Also worth flagging: a previous support reply told this user "both fixes are now tracked and queued" for this exact class of bug. No matching commit or tracking issue existed in this repo before this one — that claim was inaccurate and shouldn't recur.

(reporter: a Notion user testing input-card conversion, contacted via the support inbox)

### Proposed direction
Get the actual export (`.apkg` or Notion export zip) for the specific toggle so we can see the exact HTML/rich-text shape, trace which import path produced the note, and confirm whether the fault is Notion-side (autoformat not triggering) or 2anki-side, before writing any fix. Already asked the reporter for this file via support. Do not reintroduce an "already tracked" claim until root cause is confirmed.

aalemayhu:
## Problem

Investigating #4440 (Notion-typed bold doesn't reliably produce a usable Input card). That issue's own notes say root cause is unconfirmed without the reporter's actual export file, and flags a prior support reply that inaccurately claimed a fix was "already tracked" -- so this PR does not claim to resolve #4440 and does not close it.

While reading the HTML/zip upload bold-to-input path (`DeckParser.transformCard` / `treatBoldAsInput`), found two concrete, independently-real gaps that could plausibly produce the reported symptom (an Input card with a blank field that marks any typed answer wrong):

1. `transformCard` stamped `card.enableInput = this.settings.useInput` unconditionally, before checking whether the bold-span or typable-back path actually produced an answer. A `<strong>` span that ends up with no usable text (empty after stripping, or a match that doesn't round-trip) still left the card flagged as an Input note -- with `card.answer` never set.
2. `treatBoldAsInput` stored the bold span's raw `innerHTML` as the answer. A bold word that also carries nested inline formatting -- `<strong>mito<em>chondrion</em></strong>`, which a real Notion or rich-HTML export can produce -- stored the answer as `"mito<em>chondrion</em>"`, not the plain text a learner would actually type.

## Fix

- `card.enableInput` is now only ever set to `true` inside the branch that also sets a real, non-empty `card.answer` (both the bold-span path and the `isTypableAnswer` back-side fallback). Everything else defaults to a plain Basic card.
- `treatBoldAsInput` now stores the bold span's trimmed `.text()` for the answer instead of `.html()`, which also means a whitespace-only bold span no longer counts as an answer.

## Testing

- `DeckParser.test.ts`: updated the existing "long/multi-line answer stays input-disabled" test to assert `enableInput === false` (previously `true` with an empty answer -- same practical outcome via the Python-side AND-gate, but now correct at the source), and added a new test for an empty `<strong></strong>` span asserting the card falls back to a valid Basic note.
- Full server Jest suite: green except the documented environmental `getPageCount`/pdfinfo failures (no `pdfinfo` binary in this environment).

## Browser check

Not run -- no dev-server/browser access in this environment. Backend parser change with full unit coverage on both the new and existing behavior.

Related to #4440 -- narrows the plausible cause set but does not confirm it; leaving that issue open pending the reporter's file per its own stated proposed direction.


aalemayhu (commented):
Security: pure parser-logic change, no user-controlled URL/HTML-sink/path/env surface. Engineering: enableInput is now only ever set alongside a real answer (both branches), treatBoldAsInput correctly switched to trimmed .text() for the stored answer while keeping .html() for the DOM replace step; test coverage now includes the happy path, nested-formatting, whitespace-only, and fully-empty bold spans, plus the DeckParser full suite (102 passed). Voice: PR body and issue comment are careful not to overclaim resolution of #4440 given the issue's own unconfirmed-root-cause note; changelog entry is clean voice, no implementation leak. Clean.

aalemayhu (commented):
Reviewed diff directly: enableInput gating tightened to only the branch that sets a real answer, treatBoldAsInput switched to trimmed .text() so nested-formatting/whitespace bold spans cant produce a wrong or blank comparison value. Four tests cover both fixes plus the corrected pre-existing assertion. No security/layer/voice issues \xe2\x80\x94 pure parser logic, no new HTTP/SQL/HTML-render surface.

aalemayhu (commented):
Re-reviewed after the complexity refactor (extract applyInputAnswer, pure behavior-preserving extraction — same DeckParser suite green, 102/102). Addresses the real SonarCloud cognitive-complexity finding on transformCard. No new issues.

Issue #4449, opened 2026-09-17
feat: bridge the duplicate-guid warning into the conversion report


aalemayhu:
Split from #4408. The oversized-package check shipped in https://github.com/2anki/2anki.net/pull/4408 (bridged the existing apkgOversizeWarning.ts detection into the persisted job conversion report).

## Remaining scope from #4408, decided by trio review

**Build (same pattern as the oversize bridge):**
- Duplicate-guid bridge: `duplicateGuidWarning.ts` already computes the dropped-note count for the one-shot upload response; the persisted `ConversionReport` (job/sync path) never gets an entry for it. Same shape as the oversize fix in #4408 -- add a `reason_code: 'duplicate_guid'` entry to `buildConversionReport.ts`, thread the count in from wherever the job pipeline has it (check whether `countDuplicateGuids(ws.location)` -- used in `UploadService.ts` -- is reachable from `performConversion.ts`, or if the job path needs its own count), wire `entryCopy`/`entryTone` in `ConversionReportModal.tsx`, and translate the new key into all 10 locales.

## Decided NOT to build (trio review, do not resurface without new data)
- Image-only front with empty back: overlaps the already-shipped `empty_back` report entry: dedupe the signal before adding a new one.
- Unused media in the package: Anki drops it silently on import -- wasted bytes only, no user-facing harm.
- Deck nesting deeper than 10 levels / empty path components: Anki renames the empty ones to "blank" and keeps importing -- cosmetic, no user harm.

## Not in scope (original #4408 constraint, still holds)
Anything that needs the user's existing Anki collection (notetype conflicts, already-present notes).

PR #4450, opened 2026-09-17
feat: warn in the conversion report when a synced deck exceeds 100 MB

aalemayhu:
## Problem

AnkiWeb refuses to sync a package over 100 MB. That check already exists (`apkgOversizeWarning.ts`, shipped separately) and shows up on the one-shot upload response, but the persisted job conversion report -- what a Notion-sync (Ankify) job shows in `ConversionReportModal` on /downloads -- never carried it. A returning Ankify user had no pre-flight signal that their deck was too big; the sync would just silently fail to reach their phone/AnkiWeb.

## Fix

Bridge the existing 100 MB threshold into `buildConversionReport.ts` as a new `oversized_package` entry (`stage: 'output'`, mirrors the existing `truncated` entry -- a flag, not a count, so it's excluded from `blocks_skipped`). Threaded the built apkg's size (already computed by `BuildDeckForJobUseCase`, in MB) into the existing `buildConversionReport({...})` call in `performConversion.ts`. `ConversionReportModal.tsx` renders it with the existing `note` tone (nothing was skipped or failed) through the same entry-list pattern every other report reason already uses -- no new page, no new component.

- **Scope: shipped the oversize bridge only tonight.** pm + designer trio call: of #4408's six original checks, two (`cloze in non-cloze deck`, `package >100MB`) were already fully shipped elsewhere (#4402, and `apkgOversizeWarning.ts` respectively) and a third (`duplicate guid`) already exists as a warning on the one-shot upload path (`duplicateGuidWarning.ts`) -- none of that was obvious from the issue text, found by reading the actual codebase before scoping. Of the three genuinely-unbuilt checks (image-only-front, unused-media, deep/blank-nesting), pm and designer agreed all three are either cosmetic (Anki handles them gracefully with no user harm) or overlap an existing report entry -- **decided not to build them**, not deferred.
- **What's left, split to a follow-up:** the duplicate-guid check needs the *same bridge treatment* as this PR did for oversize (it has detection, just not a `ConversionReport` entry for the persisted job/sync path). Opened https://github.com/2anki/2anki.net/issues/4449 rather than compress it into this slot, since it needs its own count-source check in the job pipeline.
- **Copy:** no exact size interpolation (`{{size}} MB`) in the localized string -- `ConversionReportEntry` has no numeric payload slot beyond `count` (already repurposed as a flag here, matching the `truncated` entry), and overloading `count` to mean "megabytes" for this one entry read as more confusing than a static "over 100 MB" message. All 10 locales translated (not just en/de).
- **No new analytics event.** The designer review suggested a `report_warning_shown` event; skipped it here since it doesn't exist anywhere in this report system yet (not even for the existing `empty_back`/`assets_dropped`/`truncated` entries) and adding one requires the dual client/server `KNOWN_EVENTS` registration this repo enforces -- out of scope for a single-entry bridge, better proposed as its own change if wanted across the whole report.

## Testing

- New tests in `buildConversionReport.test.ts` (flags over 100 MB, doesn't flag at/under the limit, excluded from `blocks_skipped`) and `ConversionReportModal.test.tsx` (renders the localized copy with a Note tone, not Skipped).
- `pnpm test` (full server Jest suite) in a fresh worktree: green except the documented environmental `PythonExitError`/no-venv and `getPageCount`/pdfinfo failures (this worktree has no `create_deck/venv`).
- `pnpm --filter 2anki-web test:run` (full web vitest suite): 414/414 files, 3718/3718 tests, including the i18n locale-parity test (confirms all 10 locales have matching keys).
- `pnpm format:check` / `pnpm lint`: clean (one pre-existing formatting issue in the test file fixed; remaining lint warnings are all pre-existing, in files this PR doesn't touch).

## Browser check

Browser check: not applicable — no dev-server/browser access in this environment. Small, additive change to an existing, already-tested component (`ConversionReportModal`), following the exact rendering pattern of the `truncated` entry that already ships; RTL test coverage exercises the real render path including the tone badge and localized copy.

Toward #4408
Follow-up: https://github.com/2anki/2anki.net/issues/4449


aalemayhu (commented):
**Verdict**: approve

Correctly scoped: read the real codebase before building, found two of #4408's six checks already shipped elsewhere and a third half-shipped, and built only the genuinely-missing piece (the report-side bridge for the already-existing oversize detection). No new page, no new component -- reuses the established `ConversionReport` entry pattern exactly.

### Blocking
None.

### Nits
None -- the "no exact size in copy" tradeoff and "no new analytics event" decisions are both explained with real reasoning in the PR body, not just asserted.

Security: N/A, no new HTTP surface, no user input reflected anywhere.
Testing: real entry-shape assertions (not just "did not throw"), full server + web suites both green (documented environmental gaps excluded correctly).
Voice/i18n: all 10 locales translated with the established register per language, no prior-brokenness hedge language, matches VOICE.md sentence-case/no-exclamation conventions.
Layering: stays within services/NotionService + the web report component; no knex/layer violations.

aalemayhu (commented):
Reviewed: correctly scoped (verified two of six original checks already shipped elsewhere before building), clean bridge pattern, no blocking findings.

PR #4452, opened 2026-09-17
fix: default favorite type to page when the caller omits it

aalemayhu:
## Problem

Triaging `/ops/errors` surfaced two open recurring groups (2026-07-31, 2026-09-09, 2026-09-11): EN "Failed to update favorite" / DE "Favorit konnte nicht aktualisiert werden" from `RulesPage`.

Root cause: `RulesPage.toggleFavorite()` sends the optional `?type=` URL query param straight through to `POST /api/favorites`, and `isValidFavoriteInput` rejects the request with a 400 when `type` is missing. Several real navigation paths into `/rules/:id` never set that param — `PreviewPage`, the per-page list in `CardOptionsPage`, and both Ankify Notion pickers (`NotionSubscriptions`, `NotionPagePicker`) — so favoriting a page reached from any of those surfaces failed every time.

## Fix

`FavoriteService.create` now normalizes `type` to `'page'` whenever the caller omits it or sends something other than `'page'`/`'database'`, instead of rejecting the favorite. This is safe because `GetEnrichedFavoritesByOwnerUseCase` already tolerates a wrong type — it retries as a database on a Notion "not a page" error — so a wrong guess self-corrects at read time. Fixing the default in one place also avoids threading `type` through every current (and future) caller of `/rules/:id`, which is the actual reason four separate pages drifted out of sync with the one page (`SearchObjectEntry`) that set it correctly.

## Testing

- New `src/services/FavoriteService.test.ts` (TDD): defaults missing/unrecognized type to `page`, keeps an explicit `database`, still rejects a missing `object_id`.
- `npx tsc -p . --noEmit`: clean.
- Full server Jest suite: 752/753 suites, 7399/7433 tests passed — only the documented environmental `getPageCount`/pdfinfo failures (no `pdfinfo` binary in this environment).
- Full web vitest suite: 414/414 files, 3722/3722 tests passed.
- `pnpm format:check` / `pnpm lint` / `pnpm arch`: clean, no new findings on touched files.

## Browser check

Not run — no dev-server/browser access in this environment. This is a backend-only default-value fix with no UI change; covered by the new unit test exercising the actual normalization logic.

Fixes the two open `error_events` groups on `/ops/errors` for the favorite-toggle failure (kept unresolved there pending this fix landing).


aalemayhu (commented):
**Verdict**: approve

Restores favoriting from Preview/Card Options/Ankify without changing UI — fixes the recurring error groups on `/ops/errors`, ties to simpler (one less dead-end action).

### Blocking
- none

### Nits
- none (low-confidence-only notes on the test file's repo mock and `unknown` casts match existing codebase convention — not new deviations)

aalemayhu (commented):
Security/engineering/ux-voice forks all clean, no blocking findings. Approved to ship.

PR #4463, opened 2026-09-17
chore: remove the dead ai-converter-floor-v1 feature flag row

### LINKED ISSUE #4453: chore: remove dead ai-converter-floor-v1 feature flag
## What

`ai-converter-floor-v1` shows on `/ops/flags` (currently On, per #2726) but nothing in the codebase reads it.

`getFeatureFlag()` (`src/lib/featureFlags/getFeatureFlag.ts`) is the only way flag values are consumed at runtime, and it has zero call sites repo-wide (`grep -rn "getFeatureFlag(" src` matches only its own definition). Grepping for the flag key itself (`ai-converter-floor`, `converter_floor`, `converterFloor`) also turns up nothing outside a changelog entry.

The feature it was meant to gate (comprehensive AI mode / 200-500 cards per chapter, `web/src/pages/WhatsNewPage/changelog/2026-05-30-ai-converter-floor.json`) appears to have shipped unconditionally, the canary flag was never wired to a check, or was wired once and the check was later removed without deleting the flag.

Compare with `pricing_v2`, which is still load-bearing: `resolveCohort()` in `src/usecases/checkout/pricingV2.ts` branches on it live, so that one stays.

## Ask

- Confirm nothing depends on `ai-converter-floor-v1` reading false (i.e. no code path anywhere expects to check it later).
- Migration to delete the `ai-converter-floor-v1` row from the `feature_flags` table (this page only flips existing flags, rows come from migrations per the `/ops/flags` page copy).
- Delete `src/lib/featureFlags/getFeatureFlag.ts` too if this was its only intended consumer, or leave it if a near-term flag will use it.

Not urgent, pure cleanup, no user-facing effect either way.

aalemayhu:
## Status: ready for merge (rail — waiting for Alexander)
Migration — hard rail, self-merge blocked per repo convention. `migration-reviewer` reviewed clean (see PR comment): no blocking issues, safe to merge as-is.

## Summary
- `ai-converter-floor-v1` shows `On` on `/ops/flags` but `getFeatureFlag()` — the only function that reads flag values — has zero call sites repo-wide. The feature it gated (comprehensive AI mode / 200-500 cards per chapter) shipped unconditionally; the flag was left behind.
- Migration deletes the row from `feature_flags` by key. Reversible: `down` re-inserts it with its original description and `value: false`.

- **Did NOT delete `src/lib/featureFlags/getFeatureFlag.ts`, despite the issue suggesting it.** The issue's own grep only checked `getFeatureFlag(` call sites and found none — but I found that `invalidateFeatureFlagCache`, exported from that *same file*, is a live import in `src/usecases/ops/SetFeatureFlagUseCase.ts:5`, firing on every `/ops/flags` toggle. Deleting the file would have broken that import. I almost shipped this wrong — deleted the file first per the issue's suggestion, then caught it via a full grep before committing and reverted. Scope is now migration-only.
- **No `pnpm kanel` regeneration.** This migration only deletes/inserts a row (DML), it doesn't add or alter a table/column (DDL) — the kanel gate in `CLAUDE.md` applies to schema changes, which this isn't. `src/data_layer/public/FeatureFlags.ts` types are unaffected. Confirmed by `migration-reviewer`: zero diff in `src/data_layer/public/` for this change.
- **No test added.** Existing tests (`FeatureFlagsRepository.test.ts`, `SetFeatureFlagUseCase.test.ts`, `OpsRouter.test.ts`, `FlagsTab.test.tsx`) all use `ai-converter-floor-v1` as a generic example key against mocked repositories, not the live table — none of them assert this specific row exists, so none needed updating. `migration-reviewer` confirmed no migration in the repo carries a colocated test — consistent with existing convention.

## Verification
- [x] Applied against local dev Postgres and confirmed the full round trip: `migrate.latest()` deletes the row (`before`: `{key: 'ai-converter-floor-v1', value: true, ...}` → `after up: undefined`), `migrate.down()` restores it (`{value: false, ...}`, matching the original migration's insert), `migrate.up()` re-deletes it cleanly.
- [x] `grep -rn "ai-converter-floor\|converter_floor\|converterFloor" src web/src` after the change — only test fixtures remain, no runtime consumer.
- [x] `tsc -p . --noEmit` — clean
- [x] `pnpm format:check` — clean
- [x] `migration-reviewer` — clean, no blocking issues
- [ ] Sonar: not run locally; PR decoration will report

Fixes #4453


aalemayhu (commented):
## Migration review — `20261016000000_remove_ai_converter_floor_flag.js`

**Verdict:** ship it

**Findings:**

- **[locking]** No concern. `del()`/`insert()` target a single row by primary key (`key text primary()` per `20260806000000_add_feature_flags.js`) on a table that only ever holds a handful of feature-flag rows. This is a row-level lock held for the duration of one DML statement — near-instant, no risk to prod traffic regardless of table temperature.

- **[rollback]** `down` is a faithful *structural* rollback but not a faithful *state* rollback, and that's the right call here: it re-inserts the row exactly as the original `20260806000000_add_feature_flags.js` migration created it (`value: false`, same description), not whatever live value the row held at delete time (e.g. if someone had flipped it to `true` via `/ops/flags`). For a flag whose value is read by zero call sites (`getFeatureFlag()` has no consumers, confirmed via repo grep), the toggled value is inert — restoring the *definition* is the correct and sufficient semantic. Worth a one-line comment in the migration itself noting "restores original defaults, not the live value at delete time" so a future reader doesn't assume otherwise, but not a blocker.
- Note: `down` is not idempotent if run twice (`insert` on an existing PK throws) — normal for knex's up/down tracking, not a concern in practice.

- **[naming]** Filename matches `YYYYMMDDHHMMSS_snake_case.js` and sits correctly in the sequential date range of neighboring migrations (`...20261015000000_create_ai_credit_grants.js` → `20261016000000_...`). Description (`remove_ai_converter_floor_flag`) accurately matches the change (deletes the `ai-converter-floor-v1` row).

- **[kanel]** Confirmed genuinely unneeded. This is DML only — no `knex.schema` call, no column/table change. `src/data_layer/public/FeatureFlags.ts`'s shape (`key`, `value`, `description`, `updated_at`, `updated_by`) is a description of columns, not rows, and is unaffected by which flag keys exist. `git status` on the branch shows only the migration file added, no `data_layer/public/*` diff — consistent with "no schema change, no regeneration needed."

- **[data shape / FKs / defaults]** N/A — no column added, altered, or defaulted; no new FK introduced.

- **[test coverage]** Verified via grep: every remaining reference to `ai-converter-floor-v1` in `src/` and `web/src/` is either a mocked-repository test fixture (`FeatureFlagsRepository.test.ts`, `SetFeatureFlagUseCase.test.ts`, `OpsRouter.test.ts`, `FlagsTab.test.tsx`, etc., using it as a generic example key) or the historical changelog JSON. None hit the live table, so none needed updating for this migration.

**Pre-merge checklist:**
- [x] Lock behavior safe at scale — trivial single-row PK delete/insert on a small table.
- [x] `down` reverses `up` by inspection — restores original row definition; toggled-value loss is acceptable since the flag has no readers.
- [x] `pnpm kanel` correctly not re-run — DML only, no schema change, confirmed no diff in `src/data_layer/public/`.
- [x] No backfill involved — single-row DML, not a batch operation.

aalemayhu:
**Migration review: ship it.**

- Locking: trivial primary-key row DML on a single-row config table, no scale concern.
- Rollback: `down` restores the exact original create-migration insert shape (`value: false`, matching description) — verified byte-identical against `migrations/20260806000000_add_feature_flags.js`'s insert. Admin-toggle history isn't round-tripped, which is correct for a dead flag being removed.
- Naming: follows convention, sequences correctly after the latest migration.
- Kanel: correctly skipped — DML only, confirmed zero diff in `src/data_layer/public/` between this branch and `origin/main`.
- Scope: confirmed one file touched (the migration). `getFeatureFlag.ts` untouched; `invalidateFeatureFlagCache` at `SetFeatureFlagUseCase.ts:5` is a genuine live import, so leaving that file alone was the right call.
- Tests: no migration in this repo carries a colocated test — consistent with existing convention, not a gap.

No blocking issues. Safe to merge.

PR #4461, opened 2026-09-17
fix: classify busboy truncated uploads as an expected client fault

aalemayhu:
## Summary
- `isExpectedClientFault` now recognizes busboy's `Unexpected end of form` (thrown from `Multipart._final()` when a client truncates an upload mid-stream — closed tab, dropped network) as an expected client-side abort, same category as `Request aborted` / `ECONNRESET` / `MulterError` it already excludes.
- Before this, the plain `Error('Unexpected end of form')` had neither a `code` nor a `MulterError` name, so it slipped through and got recorded as a real server error in `error_events`, polluting `/ops/errors`.

Cosmetic dashboard-noise fix only — no functional bug, no user-visible behavior change, no changelog entry.

## Test plan
- [x] TDD: added a failing test (`returns true for a busboy truncated-upload abort`) first, confirmed it failed for the right reason, then made it pass.
- [x] `pnpm test -- src/lib/misc/isExpectedClientFault.test.ts` — 11/11 passed
- [x] `tsc -p . --noEmit` — clean
- [x] `pnpm format:check` — clean
- [ ] Sonar: not run locally; PR decoration will report

Fixes #4455

PR #4464, opened 2026-09-17
chore: relabel prune-dead-uploads as a between-sweeps check

### LINKED ISSUE #4454: chore: retire or relabel the manual prune-dead-uploads command
## What

The manual "Prune dead upload rows" ops command (`/ops/commands`, shipped in #4419) is now superseded by the automated daily self-heal added in #4435 (`deleteDeadUploadRowsInDatabase`), which the #4435 commit body describes explicitly as making this class of bug self-heal on a schedule instead of needing the manual button.

## Ask

Decide whether to:
- Remove the manual command entirely (route, controller, use case, `CommandsTab.tsx` button), or
- Keep it but re-label as an explicit manual-override / between-sweeps verification tool, so it doesn't read as the primary mechanism anymore.

Not urgent, no user-facing effect either way.

aalemayhu:
## Summary
- Trio call on issue #4454: the manual "prune dead upload rows" ops command is superseded as the *primary* mechanism by the daily automated self-heal (#4435), but not made redundant — relabel, don't remove.
- Renamed the panel heading from "Prune dead upload rows" to "Verify dead upload rows" and reworded the subtitle and the OpenAPI description to frame it as a between-sweeps check / immediate override, not the primary cleanup path. No route, controller, use-case, or button-label change.

- **Keep vs remove: keep.** All three trio legs converged independently. PM: the dry-run (`missingRows`/`sampleKeys`) is the only way to inspect what a sweep will act on before or between runs. Designer: same read on the UX side — reorder the mental model so "Check" reads as primary intent. Engineer: confirmed the automated self-heal genuinely runs on a schedule (`ScheduleCleanup.ts`, 24h interval, wired at boot in `server.ts`, not flag-gated) and that the manual path is a *separate implementation* (not shared code) with a real capability gap the cron lacks — on-demand dry-run verification.
- **Scope: copy-only, both `CommandsTab.tsx` and `OpsRouter.ts`'s Swagger description.** PM/designer only flagged the UI card; the engineer leg additionally found the OpenAPI description still framed this as backfilling old orphan-sweep damage with no mention of the automated path, so I updated both in the same PR rather than leaving the API doc stale.
- **Did NOT touch:** button labels ("Check dead rows" / "Prune dead rows" — already accurate), any run logic, or the automated helper. Engineer flagged that the manual path lacks the automated helper's `deletionVolumeAlert` anomaly guard — noted here as a real gap, not fixed, since sharing that guard is a separate, non-trivial change outside this issue's scope.

## Test plan
- [x] TDD: added a failing test (`frames the panel as a between-sweeps check, not the primary cleanup`) asserting the new heading and subtitle text, confirmed red for the right reason, then green.
- [x] `pnpm vitest run src/pages/OpsPage/CommandsTab.test.tsx` — 12/12 passed
- [x] `tsc -p . --noEmit` (server) and `tsc --noEmit` (web) — clean
- [x] `pnpm format:check` — clean
- [ ] Sonar: not run locally; PR decoration will report

Browser check: not applicable — admin-only `/ops` page (Alexander is the sole user), copy-only change with no interactive behavior change; the updated vitest assertions cover the rendered text directly.

Fixes #4454


aalemayhu (commented):
**Verdict**: approve

Copy-only relabel that keeps a real capability (dry-run/on-demand verification) instead of removing it — trio decision documented in the PR, all three review dimensions clean.

### Blocking
None.

### Nits
None.

aalemayhu (commented):
Security, engineering, and UX/voice review forks all came back clean — no blocking findings on this copy-only relabel (Swagger description + React heading/subtitle + matching test). Browser check out-clause in the PR body is honest: admin-only page, pure copy, covered by the new vitest assertion, no live walkthrough needed.

aalemayhu (commented):
Clean rebase onto main (picking up the just-merged #4461) — no content change from the previously reviewed SHA. Security/engineering/UX-voice review still holds.

PR #4467, opened 2026-09-18
chore: add a local-cleanup step to /ship's success path

aalemayhu:
## What
Adds step 7 (Local cleanup) to `.claude/commands/ship.md`, between prod verification and the failure/revert path.

## Why
`CLAUDE.md`'s Git section states the checkout-main/pull/branch-delete rule, but `/ship`'s own step list never called it — the Report step just prints the PR URL and deploy verdict. That gap meant the rule was honor-system prose an agent could (and did) forget after a real ship.

## How
New step 7 runs only on the success path (healthy deploy or no-deploy merge): `git checkout main && git pull --ff-only`, `git branch -D <branch>`, `git fetch --prune origin`, and `git worktree remove <path>` if a dedicated worktree was used. Explicitly keys off the queue's confirmed `MERGED` state from step 4 (not the enqueue moment) — this repo merges through a merge queue, so `gh pr merge` only enqueues, and deleting the branch before the queue actually lands it would break the queue's build. Skips silently if the checkout isn't on the just-merged branch (nothing to clean up).

## Testing
Doc-only change to a command's instructions — no code path to test. `/check` not applicable.

## Browser check
Browser check: not applicable — `.claude/` command file, no rendered output.

## Goal alignment
Indirect: keeps the harness's own workflow doing what it says it does, so future ships don't leave stale branches/worktrees for Alexander to notice and flag.


aalemayhu (commented):
**Verdict**: approve

Wires the checkout-main/branch-delete rule (previously honor-system prose in CLAUDE.md) into /ship's own step list as step 7, keyed off the queue's confirmed MERGED state. Round 1 caught the no-deploy path skipping cleanup; fixed and re-reviewed clean.

### Blocking
- none

### Nits
- none

Issue #4466, opened 2026-09-18
Upload dropzone: format-pill overlay blocks file-input click at 375px


aalemayhu:
## What

On `/upload` at 375px viewport, clicking the dropzone's "Choose files" area times out — Playwright reports a `<span class="UploadForm-module__formatPill___PfrTY">My Clippings.txt</span>` intercepting pointer events over the hidden `<input type="file" id="pakker">`. The format-pill list (`.zip .html .md .pdf .epub My Clippings.txt .docx ...`) appears to overlay the file input's clickable area at mobile width.

## Repro

1. `pnpm dev`, open `http://localhost:3000/upload` at 375×812.
2. Dismiss the quick-tour dialog.
3. Click the dropzone ("Drop your files here or Choose files").
4. Click is intercepted by a `formatPill` span instead of reaching the file input.

## Where

`web/src/pages/UploadPage` (or wherever `UploadForm-module__*` lives) — the format-pill list's stacking/pointer-events relative to the file input at narrow widths.

## Found during

Browser attestation for #4350 (shared-deck reweight) — unrelated to that PR's diff, so not fixed there. Drag-and-drop upload is presumably unaffected (native OS drop targets the whole zone); this is specifically the tap-to-choose path on mobile widths.

## Impact

Unverified against real mobile Chrome/Safari — Playwright's strict pointer-interception check may be stricter than a real touchscreen tap. Worth a manual mobile check before prioritizing; filing so it doesn't get lost.

aalemayhu:
Investigated. Root cause: `#pakker` is hidden via `clip: rect(0,0,0,0)`, which Chromium treats as non-hit-testable at any position — a direct click on that input always gets intercepted by whatever paints at its box, regardless of where it is positioned relative to the format pills. That is exactly what the original Playwright repro hit (targeting the raw input directly).

Clicking any visible content inside the wrapping `<label for="pakker">` — the format pills included — still opens the file chooser correctly via the browser's native label click-forwarding, which does not depend on the hidden input own hit box. Verified with a `filechooser`-event assertion at 375px, clicking a format pill directly: https://github.com/2anki/2anki.net/pull/4469. This matches what a real tap does (lands on the visible pill, bubbles to the label).

No real user-facing defect, so no source change went in. Closing; the PR above adds a regression test that locks in the working path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4469, opened 2026-09-18
test: lock in upload dropzone click-to-choose at 375px

aalemayhu:
## Summary

- Investigated #4466 (format-pill overlay reportedly blocking the file-input click at 375px).
- Root cause found: `#pakker` is hidden via `clip: rect(0,0,0,0)`, which Chromium treats as non-hit-testable at any position — a direct click on that input always gets intercepted by whatever paints at its box, regardless of where it is positioned. That matches exactly how the original Playwright repro found it (targeting the raw input).
- Clicking any visible content inside the wrapping `<label for="pakker">` — the format pills included — still opens the file chooser correctly, via the browser's native label click-forwarding, which does not depend on the input own hit box. Verified with a `filechooser`-event assertion at 375px (matches a real tap: a real touch/click also lands on the visible pill first, then bubbles to the label).
- No real user-facing defect found, so no source change. This PR only adds a regression test locking in the working path, so a future change that breaks label-forwarding (e.g. an errant `pointer-events: none`) gets caught.

## Browser check

`pnpm --filter 2anki-web exec playwright test tests/upload-dropzone.spec.ts` passes locally.

## Test plan
- [x] New Playwright test `web/tests/upload-dropzone.spec.ts` passes locally.
- [x] `pnpm format:check` clean.

Recommend closing #4466 with this finding, will comment on the issue.


aalemayhu (commented):
**Verdict**: approve

Test-only PR (single new Playwright spec, no source change). Locks in the real click-to-choose path at 375px after investigation found #4466 was a Playwright-locator artifact (clip-hidden input is non-hit-testable directly), not a user-facing defect. Security fork: no issues, no production code touched. Engineering fork: no issues, no layer/mocking/comment violations; low-confidence note on the hardcoded pill-text string match being coupled to real UI copy, not a rule violation.

PR #4470, opened 2026-09-18
feat: pin note mod to content-changed-at, not export time

### LINKED ISSUE #4445: feat: note-level content-hash mod for signed-in re-imports (card_guids ledger)

aalemayhu:
## What

Signed-in re-uploads of markdown/HTML/CSV/xlsx decks no longer overwrite edits a user made to a card inside Anki after their last upload — only when the card's actual content changed does the re-exported note carry a newer Anki `mod` timestamp.

## Why

Closes #4445. Every genanki note was stamped with export time on every re-upload; since Anki overwrites any guid-matched note whose incoming `mod` is newer, an unedited re-upload always looked "newer" than whatever the user had edited locally and silently clobbered it. 22% of owners re-upload the same deck name at some point — a paying-user path. (Companion fix #4447 already solved the same problem for notetype/CSS mod; this is the separate note-content half.)

## How

- Migration: `card_guids.content_changed_at` (nullable timestamptz, no backfill/default). Reuses the existing `fingerprint` already packed into `source_page_id` (from #4388) as the content digest — deliberately **not** a new `content_hash` column, a scope cut the trio synthesis made over the original issue's two-column ask.
- `resolveUploadIdentityGroup` (`src/lib/anki/uploadCardIdentity.ts`) now also resolves a `contentChangedAt` per card: held at the stored value on an exact fingerprint match (content genuinely unchanged), bumped to `now` in every other case (first sighting, an in-place answer edit, or a guarded fork).
- `DeckParser.applyUploadIdentityGuids` computes `now` once per parse, sets `card.mod`, and writes `contentChangedAt` back into the ledger entries.
- `card.mod` is threaded into `deck_info.json` the same channel `explicit_guid` already uses. `create_deck.py` passes it to a new `N2ANote` (`create_deck/helpers/n2a_note.py`), which reimplements genanki's `Note.write_to_db` to substitute the pinned value for the *notes*-row mod only (card-row/scheduling mod stays at export time) — genanki threads one export-wide timestamp through both, with no per-note hook, so a full override was the only option (confirmed by reading the vendored genanki source before writing this).
- Duplicate-guid fold-in the issue also asked for turned out to already be substantially covered by `resolveUploadIdentityGroup`'s existing fingerprint-based disambiguation (#4388) — no new logic needed there, only the regression test.
- Trio-reviewed spec: `git log -p -- Documentation/specs/note-reimport-mod.md` (removed after implementation, per the spec-lifecycle convention).

## Measuring success

No new analytics event this round (deferred — see Risks). The existing `upload_identity_replayed` event (owner-level `issued`/`replayed`/`guarded` counts, already shipped with #4388) is the closest existing signal; a day-7 check can sample `card_guids` `u:` rows for a returning owner and confirm `content_changed_at` stayed put across two back-to-back identical re-uploads.

## Testing

- `src/lib/anki/uploadCardIdentity.test.ts` — content-changed-at held/bumped/seeded across every decision branch (issued, replayed via exact match, replayed via answer-edit, guarded, legacy null row).
- `src/lib/parser/DeckParser.uploadIdentity.test.ts` — new `note mod` describe block: an unchanged re-import holds `card.mod`, a changed answer bumps it (fake-timers on `Date.now`, not wall-clock).
- `src/data_layer/CardGuidLedgerRepository.test.ts` — SQL-shape assertions for the new column on read, write, and upsert-merge.
- `create_deck/helpers/test_n2a_note.py` — writes a real `.apkg` via genanki `Package`/`Deck` and reads the sqlite `notes.mod` back: pinned value wins, unset falls back to export timestamp, two re-exports with the same pinned mod produce identical output.
- Full server jest suite run (not just targeted files, per the parser-shared-code rule): 744 suites / 7333 tests passing; the 9 pre-existing suite failures are the documented worktree gaps (`PythonExitError` — no venv/`CREATE_DECK_DIR` in a worktree checkout; `getPageCount.test.ts` — no `pdfinfo` binary) and one resource-contention flake in an untouched file (`BlockHandler.test.ts`, passes in 18s isolated) — verified each failure's error type individually before attributing it to environment, not this diff.
- Full Python `create_deck` suite: 145 passed, 1 skipped (pre-existing, unrelated).
- `pnpm run build`, `pnpm run arch` (0 errors), web typecheck/lint/vitest (414 suites / 3728 tests) all clean — no web/src changes in this diff besides the changelog JSON.

## Browser check

Not applicable — only a changelog JSON changed under `web/src/`, no runtime UI change.

## Changelog

`web/src/pages/WhatsNewPage/changelog/2026-09-18-reuploading-a-deck-keeps-your-card-edits.json`: "Re-uploading a deck keeps the changes you made to individual cards inside Anki" — deliberately worded to distinguish it from #4447's already-shipped "keeps the card style you edited" entry from the day before.

## Risks

- No new analytics event for the held-vs-bumped split (see Measuring success) — if this needs prod visibility beyond the existing `upload_identity_replayed` counts, that's a small follow-up, not blocking.
- `N2ANote.write_to_db` reimplements (not wraps) genanki's method — a future genanki version bump could silently drift from the vendored implementation it copies. `test_n2a_note.py` writes and reads back a real apkg, so a genanki upgrade that breaks this fails loudly, not silently.
- Rollback: migration `down` drops the column; the next re-upload for every affected owner re-seeds it as a first sighting (one extra mod bump per returning user), same shape as the guid table's own accepted rollback behavior.

## Goal alignment

More beautiful / more trustworthy — a re-upload stops destroying study work a paying user did inside Anki. No user-facing UI, so no funnel metric moves; the leading indicator is qualitative (support inbox silence on "my edits disappeared" reports) rather than a dashboard number.

## Decisions

Trio synthesis (full detail in the removed spec's git history): pm cut the issue's proposed `content_hash` column in favor of reusing the existing `fingerprint`; designer ruled out any UI surface (the server can't honestly claim "kept your edits" at export time, since it has no view into the user's live Anki collection) in favor of a changelog entry; engineer confirmed the genanki `Note.write_to_db` obstacle and found the subclass pattern already precedented by #4447's `N2AModel`. All three independently converged on treating the duplicate-guid fold-in as a test to confirm rather than new logic.

Tier 3 — hard-rail migration. `migration-reviewer` review requested; this PR stops at ready and waits for Alexander's manual merge per the autonomous-shipping policy.


aalemayhu (commented):
**Verdict**: approve (hard rail — waiting for Alexander to merge)

Migration-reviewer flagged one real issue pre-review (content_changed_at created as plain timestamp instead of timestamptz, an outlier against ~24 recent sibling migrations and a real TZ-dependent correctness risk given the column feeds epoch-second math) — fixed, verified locally (rolled back/reapplied, confirmed timestamp with time zone), and kanel confirmed as a no-op. Security fork: no issues. Engineering fork: no issues — every touched file has a matching test change, no layer violations, no scope creep, Date.now() called once per parse not per-card.

PR #4480, opened 2026-09-18
fix: lead the apkg-as-zip reject with pick-a-different-file, not /print

### LINKED ISSUE #4478: UX: users retry rather than follow the apkg-as-zip reject CTA to /print
## What

`isAnkiPackageZip` (shipped 2026-09-12, #4411) correctly rejects a `.zip` upload that's actually a renamed/re-exported `.apkg` — detection is solid, zero false positives observed. The client (`UploadForm.tsx`) matches the error and swaps in a "That's already an Anki deck" panel with a CTA to `/print` (apkg-to-PDF, a working, separate feature).

The problem is downstream of the message: users aren't following the `/print` CTA, they're retrying the same rejected upload instead.

## Evidence (prod logs, 2026-09-12 through 2026-09-18)

- 13 occurrences over 6 days — 100% the zip-content variant (a genuinely renamed `.apkg`, not the plain `.apkg`-extension case).
- Clusters into repeated-attempt bursts, not one-off hits: one mobile Safari session hit it 4x in 2.5 minutes (all `400`s, same upload endpoint), two more sessions hit it 2-3x within a couple of minutes each on separate days.
- No sign of a false positive anywhere in the sample — every hit is a real renamed-apkg zip.

## Why this matters

The message and redirect exist and are correct, but at real volume (~2 sessions/week) people are bouncing off the wall and retrying rather than converting. Possible causes worth testing:
- The `/print` CTA doesn't stand out enough, especially on mobile (the clearest retry-burst was an iPhone Safari session).
- "Print as PDF" isn't the thing a confused user expects when they believe they uploaded a Notion export — the framing may need to explain *why* retrying won't help before offering the alternative.

## Ask

UX pass on this error state — trio review needed since it's user-facing copy/flow. Not urgent (modest volume), filing so it isn't lost.

aalemayhu:
## Summary
- Issue #4478: prod logs showed users retrying the same rejected upload (one mobile Safari session 4x in 2.5 minutes) instead of following the `/print` CTA — the panel offered PDF export, a goal the user doesn't have when they believe they uploaded a Notion export.
- Trio (pm/designer/engineer) synthesis: flip the hierarchy. Primary action is now "Choose a different file" (wired straight to the file picker, not a reset-then-hunt two-tap flow), with a why-line explaining a finished Anki deck can't be re-converted. `/print` is demoted to a quiet secondary link, kept because it's still a real option for someone who does want an offline copy.
- Ships `apkg_reject_shown` / `apkg_print_cta_clicked` analytics events (client + server allowlists) so the fix is measurable — no event fired on this path before.
- Routes the copy through i18n (`upload.form.apkgReject*` keys in `common.json`, all 10 locales) — the prior copy was hardcoded English, the only error state in this component that wasn't.

## Test plan
- [x] `UploadForm` unit tests updated for the new copy/hierarchy and pass
- [x] i18n parity test (406/406) — all 10 locales have matching keys
- [x] Full web vitest suite green (414/414 files, 3729/3729 tests)
- [x] Server tsc, web typecheck clean
- [x] `pnpm format:check` clean tree-wide
- [x] Changelog entry added (`2026-09-18-apkg-reject-explains-why.json`)

Closes #4478

## Browser check
Notes: ran `pnpm --filter 2anki-web test:golden-path` (machine-backed attestation per `.claude/docs/browser-attestation.md`) — 4/4 passed, no console/page errors at 375px. The apkg-reject panel change itself is covered by the updated `UploadForm.test.tsx` unit tests (new copy, hierarchy, and file-picker click behavior).


aalemayhu (commented):
**Verdict**: approve

Fixes a real retry-loop (issue #4478) by matching the CTA to the user's actual intent — simpler and faster for the confused-upload recovery path.

### Blocking
- none

### Nits
- none

aalemayhu (commented):
Three-fork review (security/engineering/ux-voice) clean, no findings. Fixes issue #4478 retry-loop.

PR #4481, opened 2026-09-18
fix: render MCQ options as a static list instead of dropping them

aalemayhu:
## Summary
- 2anki's own `n2a-mcq` note type builds its multiple-choice options entirely client-side, via an embedded `<script>` that reads the `Multiple Choice` field at runtime and injects option elements into an empty `<form>`.
- `ApkgPreviewService.sanitizeCardHtml` correctly strips `<script>`/`<form>` from card HTML for security (can't execute arbitrary script from an uploaded deck) — but that means every surface built on `ApkgPreviewService.renderCard` rendered MCQ cards as just the bare question, with the entire answer-choice list silently missing:
  - PDF export (`/print`)
  - the live in-browser apkg preview page
  - the shared-deck page
  - MCP card reads
- Mirrors the existing image-occlusion special case (`renderIoCard`/`isImageOcclusionTemplate`): detect the note type by its `Multiple Choice`/`Correct Answer` fields and render the options server-side as a static numbered list — front shows all options unmarked (a self-test worksheet), back re-lists them with the correct one marked (✓ + bold + highlight) plus the `Extra` explanation. If `Correct Answer` doesn't match any option verbatim (a source-deck authoring bug), falls back to a labeled "Correct answer: …" line instead of guessing.
- Fixing this in the shared `ApkgPreviewService` (not scoped to just the PDF use case) fixes all four consuming surfaces in one change.

## Test plan
- [x] New `ApkgPreviewService.test.ts` coverage: options render on front, correct option marked on back, Extra renders, no-match fallback, no `<script>`/`<form>` leak, card isn't dropped from the deck list
- [x] `src/services/ApkgPreviewService/` full suite green (12 suites, 114 tests)
- [x] Verified against a real MCQ `.apkg` end to end — rendered the actual PDF via `ExportApkgToPdfUseCase`/`PdfRenderService` locally and visually confirmed the output (options + checkmark + explanation, matches the intended layout)
- [x] Server `tsc --noEmit` clean
- [x] `pnpm format:check` clean tree-wide
- [x] Full server Jest suite: 752/753 suites, 7422/7456 tests pass — the one failing suite (`getPageCount.test.ts`, pdfinfo) is the documented pre-existing environmental failure, unrelated to this change
- [x] Changelog entry added (`2026-09-18-mcq-decks-print-with-answer-choices.json`)


aalemayhu (commented):
**Verdict**: approve

Fixes a real correctness bug (MCQ cards silently missing their options) across four surfaces at once by fixing it at the shared render layer, mirroring the existing image-occlusion pattern — more beautiful, and removes a class of blank-card complaints.

### Blocking
- none

### Nits
- `ApkgPreviewService.ts` `NOSONAR` on `MCQ_OPTION_SPLIT_REGEX` could use an inline reason per security.md convention (non-blocking, cosmetic)

aalemayhu (commented):
Two-fork review (security/engineering) clean, no blocking findings. Fixes MCQ options silently missing on 4 surfaces.

Issue #4479, opened 2026-09-18
fix: PDF export times out silently on large decks that bypass the free-tier card cap


aalemayhu:
## What

PDF export ("Print as PDF" on an `.apkg`) fails with a generic 500 `"PDF generation failed."` on large decks, instead of a clear limit message.

## Root cause

- `MAX_CARDS = 1000` in `src/usecases/apkg/ExportApkgToPdfUseCase.ts` is bypassed whenever `isPaying(res.locals)` is true (`src/controllers/ApkgController.ts`'s `exportPdf` — any paid subscription, not just Ankify/Auto-Sync). A paying user sails straight past the cap that would otherwise stop a free user with a specific "This deck has N cards, PDF export supports up to 1000" message.
- With the cap bypassed, `ExportApkgToPdfUseCase` builds one HTML document with every card as a table row. Each row ships its own `<style scoped>` block per side — `scoped` was removed from Chrome years ago, so on a large deck thousands of these apply globally, forcing full-document style recalculation.
- `PdfRenderService.ts` has a hard 30s timeout (`PDF_TIMEOUT_MS`) for the MathJax-settle wait and the `page.pdf()` call.
- Reproduced against a 9,799-card / 55MB test deck (3 sub-decks, image/media heavy): `TimeoutError: Timed out after waiting 30000ms` on the MathJax `waitForFunction`, ~46s total before Puppeteer gives up.
- That `TimeoutError` isn't `ApkgTooLargeError` / `CardLimitExceededError` / `InvalidApkgError`, so it falls into the generic catch in `ApkgController.ts` (`sendPdfExportError`) → 500 `"PDF generation failed."` — no indication of what actually went wrong.

## Not the cause

- multer's 100MB upload limit (file was 55MB, well under)
- corruption / invalid apkg (parses cleanly, correct card/deck counts and media)

## Fix shape (needs a product call, not mutually exclusive)

1. Cap PDF export by card count for **everyone**, including paying users, at whatever the renderer can reliably finish inside the timeout — surfaced via the existing `CardLimitExceededError` message instead of silently bypassing.
2. If large decks should genuinely be supported: raise `PDF_TIMEOUT_MS`, and stop duplicating a `<style scoped>` block per row (hoist card CSS to one `<style>` in `<head>` — `scoped` is a no-op in modern Chrome anyway, so today's approach buys nothing but the recalculation cost).

## Repro

[...]

PR #4482, opened 2026-09-18
fix: say a deck is too large to print instead of a generic 500

aalemayhu:
## Summary

A very large deck that times out in the PDF renderer used to return a generic 500 `"PDF generation failed."`. Paying users bypass the 1000-card cap, so they are the ones who hit it, with no next step.

- `PdfRenderService` turns Puppeteer's `TimeoutError` (page load, MathJax wait, or `page.pdf`) into a typed `PdfRenderTimeoutError`. Other render failures pass through untouched.
- `sendPdfExportError` answers `400` with a specific message, without `console.error`, and records a `pdf_render_timed_out` event so the size of the affected slice can be read.
- Server-only change. The client already shows the server `message` verbatim for a 400.
- Changelog entry added.

Refs #4479. This PR does **not** close it: the issue's main fix shape is removing the paid-user bypass of the card cap, which is a quota change and is left for Alexander.

- Status code: chose `400` over `402/403/5xx`. The client special-cases 401/402/403 (upgrade CTA, which would be a false promise to a paying user) and 5xx risks proxy error pages and Sentry noise. `400` matches `CardLimitExceededError`. Assumption: nothing upstream rewrites 400 bodies. If wrong, change the status in `sendPdfExportError`.
- Copy: "This deck is too large to turn into a PDF in time. Split it into smaller decks in Anki and print them one at a time." (designer's call). It deliberately avoids the phrase the client uses to show "Upgrade for unlimited", and states no card threshold because the real cutoff is unmeasured. Swap if you'd word it differently.
- Localization: the message is English-only, shown raw, like the existing card-limit message. The designer preferred a localized `print.tooLarge` key across 10 locales plus `role="alert"` on the error paragraph; I did not build that here because it turns a server-only fix into a web change unrelated to the reported symptom. Follow-up if you want it.
- Telemetry: added `pdf_render_timed_out` (userId only, no card count; the controller doesn't know it) so the morning cap decision has a count instead of a guess.
- Scope: built the typed timeout mapping only. Deliberately did NOT build:
  - removing the paid bypass of `MAX_CARDS` (quota rail, your call);
  - raising `PDF_TIMEOUT_MS` (safety limit, no measurement behind a new number);
  - skipping MathJax when a deck has no TeX (a missed-delimiter false negative would silently drop math; pm wanted it, engineer left it out, I sided with the smaller PR);
  - hoisting the per-row `<style scoped>` blocks (engineer verified that keeping each unique block in last-occurrence order preserves today's cascade, but the render-time gain is unmeasured);
  - base64 media dedup / chunked or async rendering (the issue's follow-up rewrite).

## Test plan

- [x] `PdfRenderService.timeout.test.ts`: timeout in load, MathJax wait and `page.pdf` each map to `PdfRenderTimeoutError`, browser is closed, other errors are not wrapped (puppeteer mocked at the edge). Red before the change (class missing), green after.
- [x] `ApkgController.test.ts`: timeout answers 400 with the message, no `console.error`, and tracks `pdf_render_timed_out` for the owner.
- [x] `tsc -p . --noEmit`, `pnpm arch` (0 errors), `pnpm format:check`, oxlint on changed files, web typecheck, web vitest (3729 passed), web lint (warnings only, none in changed files).

Browser check: not applicable — the only `web/src/` file is a changelog JSON, no rendered output changes.


aalemayhu (commented):
**Verdict**: approve

Simpler failure path for paying users with very large decks: a render timeout now returns a specific, actionable 400 instead of a generic 500, and is measurable via pdf_render_timed_out. Security, engineering and voice reviews found no blocking issues; puppeteer TimeoutError identity confirmed against real Chrome. Paid-cap bypass is intentionally untouched (quota decision left open on the issue).

### Blocking
- None

### Nits
- The owner-id expression is now duplicated once in ApkgController; wait for a third use before extracting.

PR #4484, opened 2026-09-18
fix: translate the card option labels and descriptions

aalemayhu:
## Summary

`/card-options` (and the settings modal, which renders the same form) showed 32 option checkboxes in English for every user. Their labels and descriptions came from the server (`supportedOptions.ts`, `GET /api/settings/card-options`) and were rendered raw, while the rest of the page was already translated.

- New `cardoptions` namespace in all 10 locales: `options.<option key>.{label,description}` plus `premiumBadge`. The glob loader picks it up, so `index.ts` is unchanged.
- `CardOptionsForm` resolves each label and description with `t('cardoptions:options.<key>.label', { defaultValue: <server text> })`. That covers the general list, the grouped lists, the section-tags row (which renders outside the checkbox list) and the `Premium` badge. A key with no entry falls back to the server text.
- `cardOptionsI18nParity.test.ts` (server jest): every option in `supportedOptions.ts` must have an English entry whose text equals the server text, no orphaned entries, and no key containing `.` or `:`. Adding, dropping or rewording an option without the matching JSON edit fails CI. Locale-to-English key parity is already enforced by `web/src/lib/i18n/locales/parity.test.ts`.
- Changelog entry added.

## Decisions (please review)

- Client-side translation keyed by option key, not server-side localization: the server stays English and stays the key source, and no API shape changes. Assumption: option keys stay stable (they already are the settings keys).
- One PR, all 9 non-English locales: the parity test forces the key to exist in all 10 files, so staging would only paste English into the six low-traffic locales, which the fallback already does. Translations were produced per locale by separate agents against each locale's existing terms (German uses "Stapel" and "Anki-Tag"; Toggle, Cloze, Anki, Notion and format names stay verbatim; emoji, cloze syntax and literals untouched).
- `Premium` badge goes through a key, and every locale's value is "Premium", the same as the existing `account.premium` in all 10 locales.
- For English users the label now comes from `en/cardoptions.json`, not the server. The parity test keeps the two identical, so an option reword needs both edits.
- Existing `CardOptionsForm.test.tsx` fixtures used invented labels for real keys (`Cloze deletion cards`, `Only convert toggles with the 🍒 emoji`); I changed them to the real server labels so they keep matching what production sends.
- Not built: no wording rewrites of the English text, no toggle, and no translation of the other English-only surfaces found in the audit (separate follow-up).

## Needs a native speaker (machine-quality, flagged by the translators)

The same three or four strings were uncertain in nearly every language, so those are the ones worth a human pass first, starting with de, pt, es (the locales with real traffic):

- `enable-input` (Anki's type-in-the-answer field, rendered as a concept rather than the word "input")
- `basic-reversed` and `reversed` (no earlier reversed-card term in most locales)
- `cherry` and `section-tags` (a fixed mode name for the "Cherry-pick" idiom; pl uses a coined "Wybór selektywny")
- `block-id-identity` (long, and its On/Off phrasing reads oddly in several languages)
- fr `download-pdfs` (could read as a user download rather than bundling PDFs into the deck)

## Test plan

- [x] `CardOptionsForm.i18n.test.tsx`: a translated label and description render instead of the server English (red before the change); the section-tags row translates (red before); the shipped German text renders for `add-notion-link`; an unknown key falls back to the server text.
- [x] `cardOptionsI18nParity.test.ts`: passes on real data; I removed one key and added a ghost key and it failed, then restored.
- [x] Independent check of all 9 files: same 32 keys in the same order, no empty strings, no untranslated copies, no exclamation marks, protected tokens present.
- [x] `tsc -p . --noEmit`, `pnpm arch` (0 errors), `pnpm format:check`, web typecheck, web vitest (3760 passed), web lint, jest for `CardOptionsController`.

## Browser check

Notes: ran `pnpm --filter 2anki-web test:golden-path` (4 passed: landing, signed-in dashboard, two account states, at 375px, no console errors). That spec does not open `/card-options`, so I did not view the translated page in a live browser. The rendered German output is covered in jsdom by `CardOptionsForm.i18n.test.tsx` against the shipped `de/cardoptions.json`.


aalemayhu (commented):
**Verdict**: approve

More beautiful and consistent for non-English users: the 32 card option labels and descriptions on /card-options now render in the selected language, with the server text as fallback and a server-side test that fails CI when options and translations drift. Security, engineering and voice/i18n reviews found no blocking issues; namespace registration confirmed via the glob loader (no ns allowlist); no HTML in any locale string.

### Blocking
- None

### Nits
- es cloze label ("Borrado por oclusión (Cloze)") and pt enable-input label read awkwardly; both reuse existing product wording and are in the PR body native-review list.

PR #4488, opened 2026-09-19
fix: answer 401 invalid_token for a bad MCP access token, not 400

aalemayhu:
## Summary

Waits for Alexander: `src/services/mcp/oauth/` is a hard-rail path (OAuth), so this is left ready for a manual merge.

`KnexOAuthProvider.verifyAccessToken` threw `InvalidClientError` for an unknown, revoked or expired access token, or one whose owner no longer exists. The SDK's `requireBearerAuth` only turns `InvalidTokenError` into a 401 with a `WWW-Authenticate: Bearer error="invalid_token"` header. Any other OAuth error becomes a **400 `invalid_client`**. It now throws `InvalidTokenError`, so a dead token gets the 401 the MCP authorization spec expects, which is what tells a client to refresh.

## What the investigation found (prod logs, 2026-09-11 to 2026-09-19)

The `POST /mcp` 400s the server check showed are two different errors:

- **69-byte body, about 270 requests** (166 `POST` and 107 `GET /mcp`, all `Claude-User`). Byte-for-byte this is `{"error":"invalid_client","error_description":"Access token expired"}`; the messages "Invalid access token" and "Access token revoked" are also 20 characters, so they serialize to the same 69 bytes. Clients hit it in bursts (POST, POST, GET within about a second) and then stop: only 1 of 273 was followed by a `/mcp/token` refresh from the same address within 20 seconds. It ran at roughly 10 a day, then 143 on 09-17 and 85 on 09-18. This is what this PR fixes.
- **198-byte body, about 980 requests.** Matches only the SDK's `Bad Request: Unsupported protocol version: <date> (supported versions: …)`, which is also spec-conformant behaviour for an unsupported `MCP-Protocol-Version` header. In the observed sessions it comes right after a token call and is followed by a successful handshake (`initialize`, `initialized`, a 12 KB `tools/list`), so it looks like a client version probe, not a failure. Three came from the Claude Code CLI. **Not changed here**, and I could not see the header value (the access log does not record it). We are already on the latest SDK (1.30.0, supports up to 2025-11-25), so an upgrade will not change it.

## Decisions (please review)

- Chose to change the error class in the provider rather than wrap `requireBearerAuth` in a route-level mapper: the provider is the one place that decides a token is invalid, and the SDK already does the right thing for `InvalidTokenError`.
- Assumption: Claude refreshes on a 401 `invalid_token` as the spec says. The logs show it refreshing through `/mcp/token` in normal flows, but I have not seen it react to this specific response. Please check after deploy (below).
- No changelog entry: this is a protocol-level status change and the user-visible effect is unverified until the logs show clients recovering.
- Not built: logging the rejected `MCP-Protocol-Version` value for the 198-byte case. Cheap to add if you want to know what Claude is sending.

## Verify after deploy

On the prod box, count `POST|GET /mcp` responses of `400 69` per hour (expect about zero) and check that 401s on `/mcp` are followed by a `POST /mcp/token 200` from the same client.

## Test plan

- [x] New `bearer auth through the SDK middleware` tests drive the real `requireBearerAuth` with the real provider: an unknown, revoked or expired token and a token with a missing owner must answer 401 with `error="invalid_token"` (all four were red with `Expected: 401, Received: 400` before the change), plus a positive control that a valid token gets 200.
- [x] All 20 server suites whose path mentions `mcp` (210 tests), `tsc -p . --noEmit`, `pnpm arch` (0 errors), `pnpm format:check`.


aalemayhu (commented):
**Verdict**: approve (rail: needs a manual merge)

Simpler and more correct auth failure for MCP clients: a bad, revoked or expired access token now gets the 401 invalid_token that tells a client to refresh, instead of a 400 invalid_client that clients stopped on. Security and engineering forks found no blocking issues: no rejection path is weakened, WWW-Authenticate values come from constants (the presented token is never echoed), nothing else depended on the old 400, and the test drives the real SDK middleware with a valid-token control. The one sturdiness suggestion (assert the token row exists) is in 3fd55e9.

### Blocking
- None

### Nits
- None

PR #4489, opened 2026-09-19
fix: show semester passes in the paid-value monitor

aalemayhu:
## What

The `/ops` paid-value monitor now covers 120d (Semester) passes.

## Why

The monitor only knew the `24h` and `7d` pass kinds, and it bounded its pass query at `now + 7d`. A 120d pass expires months out, so the SQL dropped it before the window logic ran; an unknown kind would also have defaulted to a 24h window that starts in the future. Result: every Semester buyer was invisible to the "paid users with no deck yet" check, on the one pass SKU that is selling. Found in an `/ops` data sweep, where the monitor checked fewer passes than the sales panel showed for the same week.

## How

- Add `120d` to the kind-duration map and raise the query lookahead to the longest kind.
- Two new tests, red first: a semester pass bought inside the window counts as value delivered, and the pass query reaches a full semester ahead.

## Checks

- `pnpm test` on `src/services/ops`, `OpsRouter`, `UserPassRepository`: 194 pass. `tsc -p . --noEmit` clean. `oxfmt --check` clean on both files.
- No changelog entry: ops-only, nothing user-visible.
- Not a hard-rail path (no glob matches).


aalemayhu (commented):
**Verdict**: approve. Makes the paid-value monitor see Semester passes, so the ops check for paid users with no deck yet covers the pass that is actually selling. Security fork: no issues. Engineering fork: no issues; the widened lookahead only affects the user-pass query, the anonymous and subscription paths are unchanged, and both new tests fail without the fix. One low-confidence note (the 120d constant duplicates the one in usecases/passes, but importing it would invert the layering and this is only the second occurrence), not worth acting on.

PR #4490, opened 2026-09-19
feat: offer the Semester Pass on the limit page

aalemayhu:
## What

The signed-in limit wall now offers the Semester Pass next to the Day and Week Pass, and leads with the Unlimited plan (order B from the preview). The Semester click is tracked as `plan: semester_pass` on the existing `paywall_upgrade_clicked` event and checks out through the same `startPassCheckout('120d', undefined, 'limit-wall')` path the pricing page uses.

Also: a presentational `LimitWall` (extracted from `LimitPage`) and a DEV-only `/dev/limit-preview` route that renders both section orders plus the redirecting and checkout-error states.

## Why

The Semester Pass is the pass that is actually selling, but the wall people hit at the free cap only rendered Day and Week (`PassCards` supports Semester; the wall never passed the handler). An `/ops` data sweep showed that most free-tier uploads end at this wall and that few people who see the paywall buy.

Serves **faster** (the right offer at the moment of intent, one click) and **simpler** (no signup-or-subscribe fork for a one-term need).

## Decisions

Trio review ran before code (pm, designer, engineer, in parallel).

- **Agree:** put Semester on the signed-in wall; cost is bounded (the pass carries a fixed AI-credit allowance with a hard ceiling below its price); no cap change; no new pass tiers.
- **Agree:** do not put passes on the anonymous wall. Signup-first is the working path there.
- **Deferred, not built:** the cancel-flow swap (offer a pass in place of the pause offer). PM and engineer argued it risks raising churn on the one surface that owns retention; the designer wanted a reason-aware version. It needs its own trio pass.
- **Conflict, decided by Alexander:** subscription first or passes first. PM leaned subscription-first for lifetime value, the designer pass-first for conversion. Alexander picked **B, Unlimited first** from `/dev/limit-preview`. `LimitPage` uses `unlimited-first`; a test pins the order. The `order` prop and the passes-first variant stay so the preview keeps showing the comparison; either can be deleted later if the preview is retired.
- **Rail boundary:** `PassCards.tsx` lives under `PricingPage`, which matches the `pricing` hard-rail glob, so it is untouched. This PR only edits `LimitPage/`, `LimitPreviewPage/`, `App.tsx` and a changelog entry.

## Review round 1 fix

The ux reviewer found the checkout error rendered below all three pass cards, so at 375px a failed click showed its error more than a screen away. It now sits directly under the pass section label and scrolls into view (`scrollIntoView?.({ block: 'nearest' })`, the same optional-call idiom `PhotoToFlashcardsPage` uses). Residual trade-off the round-2 reviewer named: after a failed click on the Week or Semester card the page scrolls up to the alert and the user scrolls back down to retry. Acceptable for a rare error path; revisit if wall checkout errors show up in support mail or logs.

## Follow-ups, not in this PR (review nits, all low stakes)

- The variant-B error note in the preview says "next to the buttons"; it is "above the passes".
- The preview scrolls to the last error variant on load because both error variants scroll on mount (real usage never mounts with an error).
- `LimitWall.test.tsx` restores the stubbed `scrollIntoView` inline rather than in `afterEach`, and has no negative test for a null error.
- Passes and buttons keep `PassCards`' hardcoded English product names (`Day Pass`, `Week Pass`, `Semester Pass`); that component is a hard rail and was not touched.

## Measuring it

Funnel on existing events: `paywall_shown` (surface `limit-wall`) -> `paywall_upgrade_clicked` (`plan: semester_pass`) -> `checkout_completed` (pass `plan`, `surface`). Day-7 read: does the semester click fire from the wall, and do total wall clicks hold. Unlimited now leads, so read its click share against the passes as well. The missing piece, `checkout_started` for pass checkouts, lives in `PassCheckoutController` (a hard rail) and ships as a separate PR for a hand merge.

## Browser check

Notes: walked `/dev/limit-preview` at 375px on the dev stack, twice (again after the review fix below). It renders the same `LimitWall` the limit page uses: variant B shows the Unlimited plan before the passes, the Semester Pass button is present, in the redirecting variants only that card turns into the disabled redirecting label, and in both checkout-error variants the alert follows the "Pay once" label and precedes the three pass buttons; 0 console errors. Not walked: the real `/limit` page (needs a signed-in free user at the cap); the container behavior is covered by `LimitPage.test.tsx`. Four stacked cards at 375px is still long; a follow-up could collapse Day and Week behind a link, which needs new copy in all 10 locales.

## Tests

- `LimitPage.test.tsx`: order pinned (Unlimited before the passes); Semester button renders; click calls checkout with `120d` and the wall surface and tracks `semester_pass`.
- `LimitWall.test.tsx`: both orders, error alert, only the redirecting pass disabled.
- `LimitPreviewPage.test.tsx`: smoke test of the preview variants.
- Full web suite passes; `tsc` and tree-wide `pnpm format:check` clean; production build emits no preview chunk (checked before the order change).


aalemayhu (commented):
**Verdict**: approve. Adds the Semester Pass to the signed-in limit wall and leads with the Unlimited plan (the order the owner chose from the preview). Round 1 found one blocking ux issue, the checkout error rendering below all three pass cards; fixed by placing the alert under the pass section label and scrolling it into view. Round 2: security no issues, engineering no blocking issues, ux no blocking issues. Recorded in the PR body as low-stakes follow-ups: a preview note that says next to the buttons where it means above the passes, the preview scrolling to the last error variant on load, and two test tidy-ups. Known residual trade-off: after a failed Week or Semester click the page scrolls up to the alert.

PR #4493, opened 2026-09-19
fix: unwrap Notion default-highlight marks so dark mode stays readable

aalemayhu:
## What

When a Notion HTML export is converted, text wrapped in `<mark class="highlight-default" data-notion-highlight="default">` is now unwrapped (`unwrapNotionDefaultHighlights`, called in `DeckParser.loadAndNormalizeDOM`). Real colour highlights (`highlight-yellow_background`, `highlight-red`, and so on) are left alone.

## Why

A support report: after taking an AI-made deck from Anki into Notion and back to Anki, only the cards that came from the Anki side looked wrong, with a highlighted look and text that disappears in Anki's dark mode. Resetting the highlight to default in Notion did not help.

What the production data shows (read-only, user 22331):

- Entry point: the file-upload converter (`POST /api/upload/file` into `PrepareDeck`, the only production `new DeckParser` site; the two canary jobs and a test helper also construct it), not a live Notion sync. The `uploads` rows have no Notion page id. The Anki-to-Notion step ran through `apkg_import` jobs a few days earlier.
- In the deck produced on 2026-09-19, 13 of 144 notes carry all of their text (front and back) inside `<mark class="highlight-default" data-notion-highlight="default">`. The other 131 notes, typed in Notion, have none. So the wrapper is on exactly the notes that came in through the Anki import.
- The note type's CSS is our template plus the export's own `<style>` block (`extractStyles`). The export block sets `.highlight-default { color: rgba(44, 44, 43, 1) }`, a near-black text colour. On Anki's dark canvas that text is dark on dark.
- Setting a highlight back to "default" in Notion cannot remove a wrapper Notion still exports, which matches "I set it to default but it is still there".

Serves **simpler** (the deck just works, in either theme).

## Not the note-type changes

Asked whether the recent note-type work caused this. It did not: #4422, #4436, #4447 and #4470 change note type ids, the IO stock kind and the `mod` timestamps, and none touches CSS or card HTML. The `mark` wrapper and the near-black colour come from Notion's export.

One related consequence worth knowing: since #4447 Anki keeps the CSS of a note type it already has, so a CSS-only fix would not reach existing collections. This fix changes the card HTML instead.

How a corrected deck reaches an existing collection depends on the GUID. On a signed-in upload, a card with a Notion block id gets its stored GUID replayed from the `card_guids` ledger (`applyLedgerGuids`), and having no `mod` of its own (the #4470 content-changed path skips cards with a `notionId`) it takes the export time, so Anki overwrites the note in place. That is the reporter's case: they are signed in and their ledger has rows for these blocks from earlier uploads. A card with no stored GUID (an anonymous upload, or a block seen for the first time) gets a content-derived GUID (`guid_for(deckName, fields[0], type)`), so once its field changes it imports as a new note beside the old one.
## What I could not confirm

- **The yellow.** The deck's CSS makes `<mark>`'s own background transparent, so I could not reproduce a yellow from the file. Unwrapping removes `<mark>` entirely, so no browser default can apply either. The invisible text is explained and fixed; whether the yellow on the reporter's screen is gone is confirmed only when they re-import.
- **`.highlight-default_background`.** The export CSS also colours this class near-black, and it is not unwrapped. I have not seen Notion emit it, so it is left as is.
- **Why Notion writes the wrapper.** The Anki importer (`ApkgToNotionBlocksService`) sends an explicit `color: 'default'` annotation on every text run, and the data suggests that is what Notion exports as a default highlight. I could not verify it without a real Notion workspace, so the importer is unchanged. Worth a check as a follow-up.

## Tests

- `unwrapNotionDefaultHighlights.test.ts`: unwraps the Notion shape, keeps inline formatting, handles a data-attribute-only mark, unwraps every mark, nested marks, a mark in a summary and in a table cell, and leaves colour highlights alone, including one nested inside a default mark.
- `DeckParser.defaultHighlight.test.ts`: through the real parser, invented stand-ins in the 2026 `details.toggle` shape and the 2024 `display:contents` shape produce a card with no `<mark>` on front or back, and a real colour highlight survives.
- The first version of both files failed before the change, for the right reason (missing module; `<mark>` still on the card). The review pass also built the fixtures with the unwrap disabled and confirmed the parser tests fail then.
- Full server suite, run before the extra test cases were added: 757 suites pass, including the parser, golden and canary suites (with the Python venv). The extra cases pass on their own. The one failure is `getPageCount`, which needs `pdfinfo` and is environmental. `tsc -p . --noEmit` and tree-wide `pnpm format:check` clean.

Browser check: not applicable — the only `web/src` change is a changelog JSON entry.


aalemayhu (commented):
Review verdict: pass. Three review rounds ran on this branch. The code and tests were clean from the first round (the cheerio unwrap was exercised on nested, block, summary, table-cell and adjacent marks with no content dropped or reordered; real parser builds with the unwrap disabled and enabled differ only by the mark; no fixture or canary baseline contains a mark). Rounds one and two blocked on wording in FEATURE.md, the commit message and the PR body about how a re-imported corrected deck reaches an existing Anki collection; both were corrected against the code (applyLedgerGuids, loadKnownGuids, create_deck.py guid selection, the n2a_note.py mod fallback) and the reporter's ledger rows in production, and the third round came back clean. One non-blocking nuance is left as is: the content-derived GUID sentence assumes the block-id-identity recovery switch is off, which is the default. Extra test cases added after round one pass. Not a hard-rail change.

PR #4494, opened 2026-09-19
fix: read ops conversion success from events, not the jobs slice

aalemayhu:
## What

The ops conversion tiles (free and paid volume, success rate, blocked by plan) now read the `conversion_succeeded` and `conversion_failed` events instead of the `jobs` table. This is row 1 of the ops-tiles workstream in the data-sweep plan.

## Why

The `jobs` table only records Notion, MCP and apkg-import conversions, a small slice of what people actually convert. Uploads, the biggest path, never wrote a job row, so the success-rate tile could not go red however the pipeline behaved. Reading the events sees every successful conversion. It does not yet see every failure (see Known gap).

## Definition

- **Succeeded:** `conversion_succeeded` events in the last 7 days.
- **Technical failure:** `conversion_failed` events whose reason is not a paywall or empty-deck reason. The classification is the shared one in `classifyFailureReason.ts` (the same one the upload funnel uses), so the two tabs cannot drift.
- **Plan blocked:** `conversion_failed` with reason `monthly_limit` or `anonymous_cap`. These are excluded from the success rate and reported separately, as before.
- **Success rate:** succeeded ÷ (succeeded + technical failures). Null when a tier has neither.
- **Free vs paid:** paid means the user has a Stripe customer id (unchanged from the jobs version). Free now also includes uploads without an account, because those are real free conversions. The tab text says so.

## Verification

- Ran the exact query for both tiers read-only against production (rendered with the same knex builder, `default_transaction_read_only = on`). It executes on real Postgres, and the counts agree with a separate aggregate I ran first by reason and by tier.
- Mutation check: folding plan blocks into the denominator makes three service tests fail; reverted.
- Note for whoever reads the tile next: most failed attempts are the two paywall gates, not errors. That is the leak the blocked-by-plan tiles show. The technical success rate will read high until the gap below is closed, so treat it as an upper bound.

## Known gap (found in review)

Failures on the Notion, MCP and apkg-import path are all emitted (`performConversion`). On the upload path, `conversion_failed` is only emitted for limits, `upload_incomplete`, `ai_credits_exhausted`, `empty_deck` and image errors. The generic sync catch and the async `.catch` in `UploadService` never track it, so Python, parser, timeout and PDF failures on uploads are invisible to this tile. The success rate is therefore still an upper bound on the upload path, for the same reason the jobs version never went red. The tab hint now says "Only failures the server records as events are counted." The follow-up is to emit `conversion_failed` from those two paths with the job failure reason code; it is the next row of the plan and changes the upload hot path, so it gets its own PR.

## Not in this PR

- The `conversion_failed` emission gap above.
- The "top failure reasons" list and the weekly failed-conversions series on the same tab still read `jobs`. They are the next rows of the plan; changing them here would have doubled the diff.
- `ai_credits_exhausted` is not in the shared paywall list, so it counts as a technical failure if it ever appears as a `conversion_failed` reason. It did not appear in the window I read. Worth a decision when the credits work settles.

## Removed

The six jobs-based methods this replaces (`countFreeConversions7d`, `countPaidConversions7d`, `computeFreeSuccessRate7d`, `computePaidSuccessRate7d`, `countFreePlanBlocked7d`, `countPaidPlanBlocked7d`) and their tests. Their cases live on as service tests with a fake events repository, and the events query has generated-SQL tests (`props->>'reason'` is Postgres-only, so an in-memory sqlite test cannot stand in).

## Tests

- `ConversionMetricsService.test.ts`: volume, success-rate maths, plan blocks kept out of the denominator, null when a tier has no data, one tier failing while the other still reports, both tiers read over seven days.
- `EventsMetricsRepository.sql.test.ts`: join shape, free and paid filters, the technical-failure clause, binding order, and the row mapper.
- Full server suite: 757 suites pass. The one failure is `getPageCount`, which needs `pdfinfo` and is environmental. `tsc -p . --noEmit` and tree-wide `pnpm format:check` clean. Web suite run as well (text change only).

Browser check: not applicable — the only `web/src` change is explanatory text in the owner-only ops tab, covered by the OpsPage tests. The tab needs the owner session and the live API, which the dev server does not have.


aalemayhu (commented):
Review verdict: pass. Round one verified the SQL (no column ambiguity against users, anonymous events land in free and never paid, deleted users keep their events, the technical-failure clause is character identical to the funnel's), the reason classification against every reason the upload and conversion paths emit, that no caller of the six removed jobs methods remains, and that the tests would fail if the tier filters were swapped or plan blocks joined the denominator. It blocked on a PR body claim that events see every conversion, when conversion_failed is not emitted from the generic upload catch paths, and on a test fixture that looked like live counts. Both are fixed in 52cca2a462ac, with the tab hint and swagger summary now stating what the tile misses, and the emission gap recorded as the next plan row. Round two on that fix commit was clean; its one nit (the body should say Notion, MCP and apkg-import failures are all emitted) is applied. Read-only production run of the exact query recorded in the body. Not a hard-rail change.

aalemayhu (commented):
Review verdict: pass. The reviewed change (two rounds, both clean at 52cca2a462ac) is unchanged. The only commit since is d2488717e7a6, a mechanical fix for the one open Sonar finding (typescript:S4323): the inline `number | string | null` union, repeated four times in EventsMetricsRepository.ts, is now a single `PostgresNumeric` alias. I checked the diff (six lines added, four removed, types only), ran `tsc -p . --noEmit`, the two affected suites (39 tests) and the tree-wide format check, all clean. No agent re-review was run for this commit, and no behaviour changed. Not a hard-rail change.

PR #4492, opened 2026-09-19
fix: rename the Unlimited plan to Pro so the name stops overpromising

aalemayhu:
## What

Renames the paid subscription from "Unlimited" to "Pro" everywhere users read the plan name: the pricing card, comparison table, FAQ, limit wall, account page, checkout success, nav badge, abandoned-checkout and price-lock-in emails, docs (English and German), SEO answer pages, the JSON-LD offers, and the MCP and IAP strings. Plan, price and entitlement are unchanged.

It also covers what crawlers and AI assistants read: the prerendered `/pricing` title and description, the `/notion-marketplace` description, `llms.txt`, and the JSON-LD offers. The Playwright checkout spec now clicks "Get Pro".

Accuracy fixes that ride along, all about the same misleading word:

- Chat is metered by the AI-credit pool, so the chat rows in `limits.md` and `chat.md` now say "Uses AI credits" instead of "Unlimited".
- Photo-to-deck draws on the same pool, so the photo-to-deck docs (English and German) say "Paid plans use AI credits" instead of "Paid plans are unlimited".
- The Day and Week pass copy said "24 hours / 7 days of Unlimited access". It now says "full access", because passes carry a fixed AI-credit allowance.
- On the Anki-to-Notion landing page, the line where the plan name changed also said "up to 5,000 cards per import"; the server caps paid imports at 10,000 notes, so it now says that.

## Why

Since prepaid AI credits shipped, the plan includes a monthly credit allowance for AI features, so a plan literally named "Unlimited" overpromises. The earlier PR fixed the capability claims (mind maps, photo-to-deck, imports). This one fixes the name itself.

Serves **simpler**: a name with no capability claim in it cannot be falsified by a future cap.

## Decisions

Trio review ran before code (pm, designer, engineer, in parallel).

- **pm:** rename to Pro, a name with zero capability content. Keep "unlimited" only where it is literally true for paying users. No announcement email.
- **designer:** the plan name is a protected string, so it stays verbatim in all 10 locales and the change is a mechanical token swap, not a re-translation. Capability words ("Unbegrenzte Karteikarten" and so on) stay.
- **engineer:** nothing reads `product.name`, so renaming the Stripe product is safe for logic. Internal identifiers stay.
[...]

PR #4491, opened 2026-09-19
fix: stop calling capped features unlimited

aalemayhu:
## What

Stops selling capped features as unlimited. Six strings change in all 10 locales (60 values):

| Where | Before | After |
| --- | --- | --- |
| Photo-to-deck free-limit message | "Upgrade for unlimited" | "Paid plans include AI credits for more" |
| Mind maps near-cap notice | "Upgrade for unlimited" | "Your plan includes N mind maps at a time" |
| Mind maps limit screen | "Upgrade for unlimited mind maps" | "... Delete one to make room" |
| Import free-plan nudge | "Upgrade for unlimited imports" | "Upgrade for larger imports" |
| Native app FAQ | "AI deck generation and unlimited use are optional paid upgrades" | "AI deck generation is an optional paid upgrade, and paid plans lift the free limits" |
| Unlimited plan card (pricing) | "AI flashcards, photo-to-deck, and chat" | "... — 300 AI credits a month" |

Also a guard test (`unlimitedClaims.test.ts`) that fails if any of the five capped-feature strings says "unlimited" (or its equivalent) in any locale, and if the plan card loses its credit count.

## Why

A subscriber on the plan named Unlimited wrote in confused by "0 AI credits left" (support inbox). Checked in code, the false claims were: mind maps (a subscriber gets a tiered cap, not unlimited), photo-to-deck (credit-gated, returns a hard stop at zero), and imports (capped per import for paid too). Left alone because they are true: unlimited cards, conversions and prints for a paid plan, and image occlusion.

Serves trust in the paid product (honest copy). No feature change.

## Decisions

Trio review ran before code (pm, designer, engineer, in parallel).

- **Agree:** keep the plan name "Unlimited" and fix the claims. The card count really is unlimited for paid; the confusion is scoped to AI credits.
- **Conflict, resolved:** the engineer read photo-to-deck as uncapped for paid; PM and designer read it as credit-gated. The credits code and its 402 test confirm PM and designer, so the photo string changed.
- **Resolved by checking code:** designer proposed "300 credits a month" for the plan card. Subscription credits reset on the billing-day anniversary, so annual plans also get monthly windows. The claim holds. Legacy $2 plans (100 credits) do not see this card.
[...]

PR #4495, opened 2026-09-19
fix: photo limit message no longer says unlimited (server)

aalemayhu:
## What
The server-side photo-to-deck free-cap error told users to "Upgrade for unlimited." Photo-to-deck is credit-metered for paying users (it draws on the AI credit pool), so "unlimited" was false. The message now names the free cap and the AI-credit path that lifts it: `Free plan is N photos per month. You've used M. Paid plans include AI credits for more.`

## Why
Part of workstream D, "stop claiming unlimited" (PLAN row 8a), following #4491 (capped-feature strings) and #4492 (plan renamed to Pro). Since prepaid AI credits shipped, photo-to-deck is capped/metered, so the copy must not promise unlimited.

## How
Changed the string in `makeFreeQuotaReachedError` (`src/usecases/imageOcclusion/PhotoToFlashcardsUseCase.ts`). The error's `used`/`limit`/`status` shape is unchanged, so both consumers keep working:
- Web client (`PhotoToFlashcardsPage`) already builds its own message from `used`/`limit` and shows the corrected copy (shipped earlier) — it ignores the server string.
- MCP (`McpToolsService.photoToDeck`) surfaces the raw message text; that string is now honest.

Test fixtures in `UploadService.test.ts` and `McpToolsService.test.ts` that carried the old wording were updated so they stop asserting the false claim.

## Measuring success
Grep prod logs for the photo 429 path: `conversion_failed` events with `reason: image_429` still fire; the surfaced message no longer contains "unlimited". No metric moves — this is copy correctness.

## Testing
- Added a use-case test that pins the corrected message and asserts it never matches `/unlimited/i` (TDD: watched it fail on the old string first).
- Full server suite: 7505 passed, 25 skipped, 7 todo; 757 suites passed. The only failures are the documented environmental `getPageCount` pdfinfo cases (no local `pdfinfo` binary), unrelated to this change.
- Full web suite: run for completeness (this PR does not touch `web/src`).
- `tsc -p . --noEmit` clean; tree-wide `pnpm format:check` clean.

## Browser check
Not applicable — server-only, no `web/src` changes.

## Changelog
No changelog entry — the web photo-limit message already shows the corrected copy (shipped in a prior PR), so no web "What's New" audience change; this PR only aligns the server string that reaches MCP clients.

## Risks
Low. String-only change; the error shape consumers depend on is untouched. Rollback is reverting one line.

## Goal alignment
Serves "simpler" (honest, accurate copy — a user hitting the cap gets the real next step instead of a false promise). No metric moves; correctness fix.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer checked the photo free-cap 429 message against the use case: it fires only for non-paying users at the monthly free cap, and paid users are metered by AI credits, so the new wording is accurate. The error shape (status, used, limit) is unchanged for the web page, the upload path and MCP, and no consumer matches on the old text. The server string now agrees with the English web copy. No blocking findings; non-blocking notes are a redundant assertion, a hard-coded 5 in the pinned test string, and a softer "lift the limit" FAQ line in the convert landing config that can be a follow-up. Not a hard-rail change.

PR #4496, opened 2026-09-19
fix: record upload failures as conversion_failed so the tile can go red

aalemayhu:
## What

Upload failures that never produced a `conversion_failed` event now do. A shared predicate, `isExpectedUploadState`, separates the known user-input states from pipeline failures on both upload paths.

- **Synchronous upload catch** (`UploadService.handleUpload`): the final fall-through, and a missing `pdfinfo` binary (`pdf_unreadable`, because that is our tooling failing). The reason for the fall-through is the existing `jobFailureReasonCode` (`parser_crash`, `python_crash`, `worker_timeout`, `unknown`, and so on).
- **Asynchronous AI upload `.catch`**: unexpected background failures, same classification, with a credits error that lost its class across the worker boundary recorded as `ai_credits_exhausted`.
- **New `isExpectedUploadState(err)`** (`src/lib/upload/isExpectedUploadState.ts`): true for the known user-input and expected states. The sync fall-through and the async branch both use it, replacing the `instanceof` list the async path had. It is a closed list, so a new user-input error class has to be added there or it is counted as a technical failure.

## Why

Review of https://github.com/2anki/server/pull/4494 found that the ops success-rate tile, now sourced from conversion events, was still an upper bound on the upload path: the generic catch paths never tracked a failure, so Python, parser, timeout and tooling failures on uploads were invisible. This closes that gap so the tile can go red when the pipeline actually breaks.

Serves **faster**, indirectly: a broken upload path shows up the day it breaks instead of as a support email.

## The worker boundary (found in review of this PR)

Uploads convert in a worker that serialises an error down to `{ message, name }`, and the parent rebuilds a plain `Error` (`jobFailureReason.ts` documents this). Every `instanceof` check on that path therefore fails in production, including the list this PR first reused. The first version of this change would have counted image-only uploads, oversized decks and Claude parse or large-section states as technical failures and made the tile read too pessimistic, and my first tests missed it because they mocked the use case and threw the real classes.

`isExpectedUploadState` matches by `name` and by message, so it works on a rebuilt error:

- names: `EmptyDeckError`, `EmptyContentError`, `PythonZeroCardsError`, `ClaudeParseError`, `ClaudeLargeSectionError`, `ImageOnlyContentError`, `DeckTooLargeError`, `IncompleteZipError`, `EpubNoAnnotationsError`, `EpubTooLargeError`;
- messages: the truncated-conversion message, the password sentinel, `docx_parse_failed`, `pdfinfo_failed` (a corrupt PDF), `pdfinfo_password`, "already an Anki deck", and the two "nothing highlighted" messages for EPUB and Kindle clippings;
- HTTP coded client faults (`isExpectedClientFault`), which includes the in-process credits stop the sync path already tracks itself, so it is not double counted.

Deliberate calls: a corrupt PDF and an unreadable docx are problems with the user's file, so they are not technical failures; a missing `pdfinfo` binary is ours, so it is. The async list already treated docx as expected, so the two paths now agree.

## What changed for logging

The async catch used to log an error level line for worker-rebuilt user-input states because its `instanceof` list never matched; those now log at info like the in-process ones. No response, status or job-state behaviour changes.

## Effect on the tile

After deploy, technical failures on uploads start appearing and the free and paid success rates move from "upper bound" toward the truth. Expect a visible step at the deploy time; that is the fix working, not a regression. History before the deploy stays as recorded.

## Known limits

- One rare double event: on the sync path `conversion_succeeded` is emitted before `incrementCardUsage`, so if that throws the request records both a success and a failure.
- Malformed OPML and Brainstorms JSON mind-map files throw plain errors with no stable name, so they still count as technical failures on the upload path. Small volume; a follow-up can give them a named error.
- Other `UploadService` paths that can fail (`restartClaudeJob`, the image path) were not audited for missing events. The image path emits its own `image_*` reasons.

## Verification

- Tests first: the positive cases failed before the change for the right reason (no event emitted).
- Mutation checks: removing the expected-state guard from the fall-through makes two tests fail; removing the name matching from the predicate makes fifteen fail (the worker-boundary cases in both the unit tests and the upload service tests).
- Reason strings bucket correctly in the shared classifier (`classifyFailureReason.ts`): the reasons emitted here are technical, and `ai_credits_exhausted` is the existing reason the sync path already emits.

## Tests

- `isExpectedUploadState.test.ts`: every expected name (as a rebuilt plain error), the message cases, the credits stop, and the values that must not be expected.
- `UploadService.test.ts`, new suite "failure events for the success-rate tile": parser crash, worker timeout and unclassified error on the sync path; missing `pdfinfo`; a corrupt PDF, unreadable docx and Anki-package-as-source not counted; the worker-rebuilt user-input states not counted on both the sync and async paths; client hang-up and empty content not counted; async unexpected failure for the signed-in owner; async credits error that lost its class; no second event for the in-process credits error.
- Full server suite: 758 suites pass. The one failure is `getPageCount` (needs `pdfinfo`, environmental, as always). `tsc -p . --noEmit` and tree-wide `pnpm format:check` clean. No web change, so no web suite.

No changelog entry — internal analytics event, nothing a user sees.


aalemayhu (commented):
Review verdict: pass. Three review rounds. Round two found the predicate missing incomplete-zip, EPUB, Kindle and password-PDF user-input states; caee72958a93 adds them, and the final reviewer verified every added name and message against the throwing source and confirmed the worker boundary keeps the name. Remaining gaps (malformed OPML or Brainstorms JSON, corrupt EPUB zip, oversized-archive messages) are low volume, counted as technical, and listed in the PR body. Full server suite passed on the prior head; tsc and tree-wide format check clean; the upload suites (206 tests) pass on this head. Not a hard-rail change.

PR #4497, opened 2026-09-19
fix: guard the upload funnel signup stage against pre-09-08 account_created

aalemayhu:
## What
The upload-funnel tab derives its signup stage from the `account_created` event. That event under-fired until 2026-09-08 — it only fired on the password signup path, missing every OAuth and magic-link signup — so any window reaching back before the fix undercounts signups and shows a fake download-to-signup rate. The default 30-day window is affected. This adds a `signup_reliable` flag to the funnel payload and teaches the tab (and the Copy-for-Claude prompt) to say plainly that the signup stage is unreliable for those windows instead of presenting a wrong number.

## Why
Workstream C, "Fix the ops tiles that lie", in PLAN.md: `account_created` under-fired before the fix, so the upload funnel's signup stage moves for the wrong reason. The emit fix landed in commit eb4384049 (#4363) on 2026-09-08, mid-day. Because it deployed part-way through 2026-09-08, that whole day is contaminated, so the reliable-since cutoff is the conservative `2026-09-09T00:00:00Z` — the first fully-clean day. The reliable event going forward is `account_created` itself (server-side, all paths, carries `signup_origin` for the per-origin breakdown), so switching to the client-side `signup_completed` would lose OAuth/magic-link coverage and origin attribution — the correct fix is a date guard, not an event swap.

## How
- `UploadFunnelService` computes `signup_reliable = since >= 2026-09-09T00:00:00Z` and returns it on both the success and error paths. Counts and rates are left as recorded (honest raw values); the flag marks them untrustworthy.
- The tab shows a warning banner (gated on `stages != null` so it never stacks with the error banner), replaces the signup count tile and the download-to-signup rate hero with an em dash + honest footnote, and dashes the per-origin "Signed up" column when the window is unreliable.
- `buildClaudePrompt` adds one standalone note above the per-origin JSON when unreliable and swaps the task line for a signup-free one, so a copied prompt cannot blame the undercounted signup step. It reads a missing `signup_reliable` field as reliable (`?? true`), matching the tab.
- The `/api/ops/upload-funnel` swagger documents `signup_reliable`.

## Measuring success
`GET /api/ops/upload-funnel?window=30d` returns `signup_reliable: false` today (2026-09-19; 30d window starts 2026-08-20). As the window ages past the cutoff, the same call on a 7d window returns `signup_reliable: true`. Confirm in the ops tab: the signup notice banner shows for 30d/90d and is absent for 7d.

## Testing
- Unit tests added:
  - `UploadFunnelService.test.ts`: signup reliable after the cutoff, at the exact cutoff, one millisecond before the cutoff (unreliable), before the cutoff, and on the repository-error path.
  - `UploadFunnelTab.test.tsx`: unreliable window renders the notice and dashes the Signup tile and origin-row cell (scoped with `within`); reliable window shows no notice.
  - `buildClaudePrompt.test.ts`: unreliable payload adds the note + signup-free task; reliable payload keeps the plain task; a payload missing `signup_reliable` reads reliable.
- Full server suite: 757 passed suites / 7509 tests; only failure is the documented environmental `getPageCount` (pdfinfo ENOENT), excluded.
- Full web suite (run alone): 422 files / 3846 tests, all pass.
- `tsc -p . --noEmit` clean; web `tsc --noEmit` clean; tree-wide `oxfmt --check` clean.

## Browser check
Browser check: not applicable — internal ops dashboard tab behind an ops login with production data; not walked in a browser. The tab is exercised by the added Vitest render tests (notice banner, dashed count/rate, per-origin column).

## Changelog
No changelog entry — internal ops dashboard change.

## Risks
Low. Additive `signup_reliable` field; existing counts/rates unchanged. If a consumer ignores the flag it sees the same numbers as before. Rollback is a straight revert.

## Goal alignment
None of simpler / faster / more beautiful directly — this is ops hygiene. It makes the upload funnel trustworthy so the paid-conversion decisions built on it (workstream C) rest on real numbers. Metric: none new; read the `signup_reliable` field on `/api/ops/upload-funnel`.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings: the cutoff comparison is epoch milliseconds on UTC values, the flag is set on the success and error paths, the payload type is consistent between server and web with no typed mock broken, the Ops page has no i18n so hardcoded English matches its surroundings, and the CSS tokens exist in all five themes. Its non-blocking notes are folded into the head commit: the cutoff moved to 2026-09-09 because the account_created fix deployed mid-day on 09-08, the exact-cutoff and just-before-cutoff cases are tested, the vacuous tab assertions were replaced with `within` checks on the Signup tile and origin row, the Copy-for-Claude prompt carries a standalone note and a signup-free task line when unreliable (defaulting to reliable when the field is missing, matching the tab), the swagger documents `signup_reliable`, and the warning no longer stacks with the error banner. I read the final diff myself. Not a hard-rail change. The other `account_created` consumers are left for a follow-up recorded in the plan.

PR #4499, opened 2026-09-19
fix: exclude MCP saves from ops job-duration percentiles

aalemayhu:
## What
The ops Performance tab's job-duration percentiles were computed as `last_edited_time - created_at` over every done job. Two things poisoned them: MCP saves create and complete their job row in the same call, so they land at about one millisecond and drag the median down; and restarted / re-synced jobs reuse the same row (`onConflict ignore`), so `created_at` stays at first creation while `last_edited_time` moves forward, inflating the tail to hours or days. This excludes `type='mcp'` from the duration percentiles and the slowest-jobs list, and relabels the metric honestly.

## Why
Workstream C, "Fix the ops tiles that lie", in PLAN.md: performance percentiles use `last_edited_time - created_at`, so re-edited jobs read as hours or days and MCP jobs at ~1ms drag the median.

## How
- `PerformanceMetricsService`: added `AND type IS DISTINCT FROM 'mcp'` to the shared done-jobs filter, so MCP saves leave both the percentile query and the slowest-jobs query. Extracted `buildDurationPercentilesSql()` / `buildSlowestJobsSql()` as pure exported functions so the generated SQL is assertable (the queries need Postgres-only `percentile_disc` and can't go through the knex builder).
- `PerformanceTab`: the duration subtitle now says the metric is created → last done update, MCP excluded, and that restarts/re-syncs make the tail an upper bound, not a per-run duration. The slowest-jobs subtitle notes MCP is excluded.

## What this PR does NOT do (needs a new field, not a migration here)
There is no per-run start timestamp on `jobs` (only `created_at` and `last_edited_time`, and `created_at` is frozen at first creation across restarts/re-syncs). So the restart/re-sync inflation on p95/p99 cannot be removed with the existing columns — the rows are individually indistinguishable from a genuinely slow single run. A real per-run conversion duration needs a dedicated `started_at`/`finished_at` pair written per run. That migration is deliberately out of scope; this PR fixes what the existing data supports (MCP exclusion) and labels the residual honestly. No per-run start/finish column was added.

## Status breakdown still counts MCP (intentional)
Only the timing metrics (duration percentiles + slowest-jobs list) exclude `type='mcp'`. The 24h terminal-status breakdown query is left unchanged and still counts MCP saves, which are always `done`. So the status-breakdown done count will exceed the duration-sample count by the MCP volume. This is intentional — the status breakdown is a truthful outcome count, not a timing sample. Noted in the `/api/ops/performance/metrics` swagger description too.

## Measuring success
`GET /api/ops/performance/metrics` — the `durations` p50 for the 24h/7d windows should rise now that the ~1ms MCP rows no longer sit in the sample, and `count` drops by the MCP volume. The slowest-jobs list should no longer contain `type: 'mcp'` rows.

## Testing
- Unit tests added (`PerformanceMetricsService.test.ts`): the SQL is built through a pg-dialect `knex({ client: 'pg' })` and asserted via `pg.raw(buildSql(), bindings).toSQL()` (mirroring `EventsMetricsRepository.sql.test.ts`), with a placeholder-vs-binding count check. Coverage: duration SQL times created→last-edited on done jobs, requires both timestamps, computes p50/p95/p99 + total, and excludes `type='mcp'`; slowest-jobs SQL excludes `type='mcp'` and orders by duration reading `last_edited_time` as the completion time.
- Full server suite: 757 passed suites / 7510 tests; only failure is the documented environmental `getPageCount` (pdfinfo ENOENT), excluded.
- Full web suite (run alone): 422 files / 3842 tests, all pass.
- `tsc -p . --noEmit` clean; web `tsc --noEmit` clean; tree-wide `oxfmt --check` clean.

## Browser check
Browser check: not applicable — internal ops dashboard tab behind an ops login; not walked in a browser. The only web change is two subtitle strings on the Performance tab (the first one is a long sentence, so its wrapping is unchecked); behavior is unchanged.

## Changelog
No changelog entry — internal ops dashboard change.

## Risks
Low. The SQL filter is additive (`type IS DISTINCT FROM 'mcp'` keeps null-type legacy rows in). Response shape unchanged. Rollback is a straight revert. Residual risk: the tail percentiles are still inflated by restarts/re-syncs — now labeled as such rather than silently wrong; fully fixing it is the deferred per-run-timestamp field.

## Goal alignment
None of simpler / faster / more beautiful directly — ops hygiene. It makes the performance tile trustworthy (median no longer dragged by MCP, tail honestly labeled) so latency regressions can actually be read. Metric: none new; read `durations` on `/api/ops/performance/metrics`.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings: `type IS DISTINCT FROM 'mcp'` keeps legacy NULL-type rows, the only writer of `type='mcp'` is the one-call MCP save, the shared filter feeds both queries with unchanged binding order, and no interface or typed mock changed. Its non-blocking notes were folded in on the head commit: the SQL tests now build the query with a pg-dialect knex and check the binding count, the swagger text and PR body say the status breakdown still counts MCP saves and that the span is an upper bound, and the title no longer overclaims. I read the delta myself. Web change is two subtitle strings on an internal ops tab. Not a hard-rail change.

PR #4501, opened 2026-09-19
fix: blocks cache counts hits (await the update)

aalemayhu:
## What
`BlocksCacheRepository.get` incremented the per-owner `fetch` hit counter with a Knex `.update()` that was built but never executed (knex builders are lazy; nothing awaited or `.then`-ed it), so the counter never moved. It now fires the update on a cache hit, matching the sibling `ConversionResultCacheRepository.bumpHit`: a fire-and-forget `.then(undefined, onError)` so the query actually runs but never blocks or fails the read, with an atomic `fetch + 1` in SQL rather than a value computed from the earlier SELECT.

## Why
Workstream E / PR-queue row 12 in the ops sweep: the `blocks` cache hit counter never increments, so there was no signal for how often the cache serves a page. This restores that bookkeeping without changing cache behavior. Scope is deliberately limited to the counter — no TTL pruning and no migration (those are a separate row).

Notes on the counter's meaning:
- This turns on exactly one UPDATE per cache hit (fire-and-forget, off the read path — the conversion-worker pool is small, so the read is never made to wait on it).
- `fetch` is per cache generation: `save` upserts and resets it to 1, so hits for the current generation are `fetch - 1`.
- Nothing in `src` reads `fetch` yet; it is bookkeeping for later ops/eviction decisions.

## How
Knex query builders are thenables that only execute when awaited or `.then`-ed. The counter update is now dispatched through a private `bumpFetch` that calls `.then(undefined, onError)` (so it runs) and is invoked with `void` from `get` (so `get` returns the cached block immediately). On failure it logs the error message only via `console.warn` (knex error strings inline SQL bindings, so the whole error object is not logged), then the read proceeds unaffected.

## Measuring success
The `blocks.fetch` column now advances on cache hits (previously flat): `select max(fetch), avg(fetch) from blocks` trending up over time confirms it. Unit-covered by `BlocksCacheRepository.test.ts`.

## Testing
- Unit tests (real better-sqlite3):
  - `get increments the fetch counter for the owner` — save (fetch=1) then get; after flushing the fire-and-forget update, the row's fetch is 2. Watched it fail first (stayed 1 with the un-executed builder), then pass.
  - `get returns the cached payload and warns when the counter update fails` — the `fetch` column is dropped after the row is saved so the real update rejects; get still returns the payload and `console.warn` is called (spied).
- Full server suite: 7506 passed; only `getPageCount` (2 tests) fails, the documented environmental pdfinfo ENOENT.
- Web suite: unchanged — no web files touched.
- `pnpm exec tsc -p . --noEmit`: clean (compiles tests too). Tree-wide `pnpm format:check`: clean.

## Risks
Low. The read path is unchanged except that the counter update now actually runs, off the read path and isolated from it. No schema change, no cache-eviction change. Worst case, a DB error on the counter is logged and the read proceeds as before.

## Changelog
no changelog entry — internal cache bookkeeping.

## Goal alignment
Serves none of simpler / faster / more beautiful directly — internal reliability/observability fix (restores a signal that was silently dead). Metric: none new; `blocks.fetch` movement in prod is the confirmation.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings and confirmed the root cause: the hit-counter update on `blocks` was a lazy knex builder that was never executed, so the counter never moved. Its non-blocking notes are folded into the head commit: the update now fires off the read path with an explicit rejection handler (matching `ConversionResultCacheRepository.bumpHit`, so a cache hit does not wait on an extra write and the two-connection worker pool is not held), it increments atomically in SQL instead of writing a value computed from the earlier read, and it logs the message only at warn level. The tests run the real repository on in-memory sqlite: one asserts the counter actually reaches 2, the other drops the column so a real update rejects and asserts the read still returns the payload and a warning is logged. The PR body no longer claims an unhandled rejection in production, and states that this turns on one UPDATE per cache hit and that nothing reads the counter yet. I read the final diff myself. Not a hard-rail change.

Issue #4049, opened 2026-08-11
fix: return-rate ops query is O(n²), stacks up and starves the DB pool


aalemayhu:
## Symptom

- The Return rate panel (`/ops/` return-rate tab) renders em-dashes and "No cohorts in this window" even though the events table has a large `conversion_succeeded` cohort in the 90d window.
- The whole `/ops` section (including `/ops/errors`, whose own queries run in ~2 ms on a 229-row table) intermittently becomes very slow.

Both have the same root cause. Observed live on prod: **8 concurrent copies of the return-rate query, each running 4–6+ minutes**, spawned by the ops page's auto-refresh, starving the DB connection pool. (They were `pg_cancel_backend`-ed as mitigation on 2026-08-12.)

## Root cause

`ReturnRateMetricsService.buildReturnsQuery` (`src/services/ops/ReturnRateMetricsService.ts:183`) computes `first_return_at` with a **correlated subquery per cohort row**:

```sql
(SELECT MIN(e2.created_at) FROM events e2
  WHERE COALESCE(e2.user_id::text, e2.anonymous_id) = COALESCE(e1.user_id::text, e1.anonymous_id)
    AND COALESCE(e2.props->>'source', 'unknown') = COALESCE(e1.props->>'source', 'unknown')
    AND e2.name = 'conversion_succeeded'
    AND e2.created_at > e1.created_at)
```

The cohort is tens of thousands of rows, the identity expression isn't indexable, so this is O(n²). The request never completes; on top of that `getMetrics` wraps everything in a bare `catch {}` that returns nulls, so the UI can't tell "empty" from "broken".

## Fix

1. Replace the correlated subquery with a single-scan window function:
   ```sql
   LEAD(created_at) OVER (
     PARTITION BY COALESCE(user_id::text, anonymous_id), COALESCE(props->>'source', 'unknown')
     ORDER BY created_at
   ) AS first_return_at
   ```
   Semantics are identical (next success by the same identity+source), and `fetchReturns`/`computeReturnRates` already reduce to the earliest gap per key.
2. Surface query errors in the response (`error` field) like `UploadFunnelService` does — drop the silent `catch {}`.
3. Add a statement timeout for ops analytics queries so a future regression can't starve the pool again.

Per `.claude/rules/testing.md`, assert the generated SQL shape with a pg-dialect knex `toString()` test.

PR #4502, opened 2026-09-20
fix: count a return only after a whole day, on cohorts old enough

aalemayhu:
## What

The ops return-rate tile now measures a real return.

- **A return** is a successful conversion at least 24 hours after an identity's first one, within 7, 14 or 30 days of it. A repeat minutes later no longer counts.
- **Mature cohorts only.** Each window counts only identities old enough to have finished it, so a two-week-old identity is not in the 30-day denominator. The tab shows that denominator (`Share of N new identities old enough for this window` on the three overall cards, `(N)` beside every source rate), and reads `—` (with "None old enough yet" on the overall cards) when nobody has finished a window. It says nothing about denominators when the query failed.
- **By source of the first conversion.** An identity is counted once, under the source of its first conversion, instead of once per source it ever used.
- **New means new.** The scan looks 30 days before the 90-day cohort window (120 days of conversions), so an identity is only a newcomer if it has no conversion between 120 and 90 days ago.
- Copy-for-Claude and the swagger text carry the same definition, plus the per-window `n`.

## Why

The sweep found this tile overstated returns by roughly a factor of two: same-session repeats counted as returns, and immature identities sat in every denominator. Workstream C row 3 in the plan. The activation half of the row (time to first deck as a share downloading within 24h) is a separate PR.

Serves none of simpler, faster or more beautiful directly. It is instrument hygiene: the number that decides whether to invest in retention has to be a number.

## How it is measured

One scan of `events` with a window function (first conversion and its source per identity) and a `FILTER` for the first conversion a day or more later. No correlated subquery, no `LEAD` (the old query stacked up under the ops page's auto-refresh, see #4049). The 30 second timeout with cancel is kept.

I ran the exact query read-only on prod as an aggregate: it is valid on Postgres and finished in about 150 ms.

## Tests

- Tests first; the new server and tab tests failed against the old code for the right reason (missing API and old wording).
- Server: generated SQL asserted through a Postgres-dialect knex (single scan, no `LEAD`, bindings in placeholder order); the pure computation for the 24-hour rule, the window boundary, maturity per window, per-window denominators, source bucketing and the empty cohort; `getMetrics` mapping database rows with `Date` timestamps and the error path.
- Mutation checks: dropping the whole-day rule fails one test, dropping the maturity rule fails four, and turning either `>=` boundary into `>` fails a boundary test.
- Web: new `ReturnRateTab.test.tsx` (definition text, denominators, the dash for an unfinished window, per-source counts), and the Copy-for-Claude prompt test.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). Full web suite: 424 files, all pass. `tsc -p . --noEmit`, web typecheck and tree-wide `pnpm format:check` clean.

## Known limits

- An identity is an account or an anonymous id, so a visitor who later signs up counts as two identities. Same as before.
- "New" is judged from 120 days of events: an identity whose only earlier conversions are more than 120 days old reads as new.
- The three windows have different denominators (a 7-day rate covers younger identities than a 30-day rate), so the rates can be non-monotonic. The denominators are shown.

## Browser check

Browser check: not applicable — internal ops dashboard tab behind an ops login with production data; not walked in a browser. The tab is covered by the new render tests.

## Changelog


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It ran the query on a local Postgres fixture and confirmed one row per identity, tie handling, the 24-hour and 120-day boundaries, that `FILTER` inside the `GROUP BY` is valid, that `Raw.timeout(ms, { cancel })` exists in the installed knex, that pg raw returns `{ rows }`, and that the response change is additive with only the ops controller, the tab, the prompt and tests as consumers. Its non-blocking notes are folded into the head commit: the swagger and PR wording now describe the 120-day scan correctly, the first-conversion source has a deterministic tie-break on `id`, source rows sort by size so they do not reshuffle on refresh, the tab no longer says nobody is old enough when the query failed, the stale "cohorts" copy and the prompt's `n` wording are fixed, and the tests now pin the exact 24-hour and window-age boundaries, the 30-second cancel timeout and a real database rejection. I mutation-checked both boundary comparisons and read the final delta myself. Not a hard-rail change.

aalemayhu (commented):
Review verdict: pass. The reviewed change (a fresh read-only review, then a fix commit for its non-blocking notes) is unchanged. The only commit since is 4a8075e19039, a types-only refactor for the one failing SonarCloud condition (new duplicated code density: the server and web return-rate types mirrored each other): the server window types are now derived from the single windows list, and the web type keeps only the fields the tab reads. No behaviour changed. I read the diff, and the full server suite (758 suites pass, only the environmental `getPageCount` fails), the full web suite (424 files), `tsc -p . --noEmit`, the web typecheck and the tree-wide format check are clean. Not a hard-rail change.

PR #4503, opened 2026-09-20
fix: replace time to first deck with download shares after signup

aalemayhu:
## What

The ops Conversions tab drops "Time to first deck" and shows two shares of new accounts instead:

- **Downloaded within 24h of signup**: the share of new accounts that downloaded a deck in their first 24 hours. This includes the deck made before signing up.
- **Made a deck after signing up**: the share that downloaded a deck at least 10 minutes after signup, still within 24 hours of it.

The cohort is accounts created in the last 30 days that are at least a day old, and never earlier than 2026-09-09, when `account_created` started firing reliably (the same constant the upload funnel uses, now exported from `UploadFunnelService`). If that cohort has not reached a whole day yet, nothing is queried and both tiles read `—`. Copy-for-Claude carries both shares (asserted in the prompt test). The card footnote names the 2026-09-09 floor.

## Why

The median was an artifact. A deck made before signing up downloads the moment the account is created, so the median measured the held-deck path, not activation. I checked prod read-only: for accounts since the reliable date, most first downloads land within a minute of signup and about nine in ten within ten minutes. So "within 24h" alone mostly restates the held-deck path, and the second tile is the one that says whether a new account made something new. The 10 minute cut-off is named `HELD_DECK_DOWNLOAD_MINUTES` next to the query.

Workstream C row 3 in the plan (the return-rate half is https://github.com/2anki/server/pull/4502). Serves none of simpler, faster or more beautiful directly; it is instrument hygiene so activation is read from a number that means something.

## How it is measured

One query (a subquery for accounts, left-joined to `events` for downloads): accounts (first `account_created` per user in the window) left-joined to their own `deck_downloaded` events inside `[account_at, account_at + 24h]`, counting distinct users overall and distinct users with a download 10 minutes or more after signup. It uses the existing `(user_id, created_at)` index. I ran the exact generated SQL read-only on prod: it is valid Postgres and finished in about 10 ms, and it returns the same counts as an independent `EXISTS` formulation.

The old `medianMinutesToFirstDeck` query, its row mapper and `formatDurationMinutes` (now unused) are deleted, and `time_to_first_deck_median_minutes_30d` is replaced by `new_accounts_downloaded_24h_rate_30d` and `new_accounts_downloaded_after_signup_24h_rate_30d`. The only consumers are the ops Conversions tab, its Copy-for-Claude prompt and the tests (the Today snapshot only carries the type in a test fixture); nothing outside the repo reads the payload.

## Tests

- Server: the generated SQL through a Postgres-dialect knex (the join condition, the 24 hour window, the 10 minute settle time as a bound parameter, and bindings in placeholder order), the row mapper (Postgres counts arrive as strings), and the service: rates from counts, no share for an empty cohort, the 30-day start and one-day end, the reliable-date floor, and no query before the cohort reaches a day.
- Mutation checks: widening the 24 hour window fails a SQL test; turning the reliable-date floor from `max` into `min` fails three service tests.
- Web: the Conversions tab tests for both cards and their footnotes and for the dash on nulls; the prompt and payload type tests.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). Full web suite: 423 files, all pass. `tsc -p . --noEmit`, web typecheck and tree-wide `pnpm format:check` clean.

## Known limits

- The accounts are `account_created` events with a user id; the cohort is small right now because the reliable date is recent, so both shares will be noisy for a while.
- The 10 minute cut-off is a judgement. A held deck downloaded more than 10 minutes after signup would count as a new deck, and a genuinely new deck within 10 minutes would not, so the second tile is an estimate in both directions.

## Browser check

Browser check: not applicable — internal ops dashboard tab behind an ops login with production data; not walked in a browser. The tab is covered by render tests for both cards.

## Changelog


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It ran the exact generated SQL on a local Postgres against a temp-table fixture and compared it with an independent `EXISTS` formulation (same counts), and checked the boundaries at exactly 10 minutes, exactly 24 hours, and the cohort start and end, in UTC and a non-UTC session; it confirmed the join, the bindings order, the service's cohort maths and early return, no import cycle, no remaining reference to the removed field, helper or query, and no typed mock left to break the deploy typecheck. Its non-blocking notes are folded into the head commit: the ten minute settle time is now a bound parameter instead of an interpolated constant (I validated it as a real prepared statement, read-only, on prod), the card footnote names the 2026-09-09 floor, the Copy-for-Claude lines are asserted, the zero-account test asserts both shares, and the PR body no longer overstates "single scan" or "errs toward undercounting". I read the final delta myself. Not a hard-rail change.

PR #4504, opened 2026-09-20
chore: stop emitting two duplicate page impression events

aalemayhu:
## What

Two page impression events stop being emitted:

- `upload_ai_badge_viewed`, which fired on the upload page for every AI badge state change on top of `upload_page_viewed`.
- `home_ai_anon_badge_viewed`, which fired on the landing page for every anonymous view on top of `landing_page_viewed`.

Both are removed from the client call sites and from the client event allowlist. The server allowlist keeps the two names for now, so a stale browser tab does not add an unknown-event warning to the logs for each event (an unknown event is not rejected, it logs one `console.warn` and the row is recorded). `upload_page_viewed` and `landing_page_viewed` stay as the one impression per page. Existing rows in the `events` table are left alone.

## Why

A read of the last 7 days of the `events` table on prod showed five near-identical impression events making up about two fifths of all rows (each fires once per page view), and nothing in `src/` reads two of them. Extra writes on the hottest table for numbers nobody looks at.

I grepped every reference before removing anything:

- `upload_ai_badge_viewed` and `home_ai_anon_badge_viewed`: only the allowlists, the call site and one test list. Nothing reads them.
- `photo_entry_point_viewed` is the **third** of the five and I am deliberately keeping it. The documented photo-to-deck funnel query in `Documentation/queries/photo-to-deck-funnel.sql` lists it in its informational `funnel_counts` section (its attempt rate uses `upload_started` and `photo_entry_point_clicked`), so removing it would change what that query reports. If that funnel is not used any more, it can go in a follow-up along with the query.

Serves **faster**, slightly: fewer writes and fewer client requests per page view, and a smaller, more honest `events` table. Metric: the share of the events table taken by these names, read with `select name, count(*) from events where created_at > now() - interval '7 days' group by 1 order by 2 desc` a week after deploy; expect both names to drop to their stale-tab tail.

## Tests

- Tests first, and they failed for the right reason: new tests assert the upload page and the landing page do not emit the removed events (`UploadPage.test.tsx`, `HomePage.test.tsx`).
- The existing parity test (every web event exists on the server) passes; the server still accepts the two names.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). Full web suite: 423 files, all pass. `tsc -p . --noEmit`, web typecheck and tree-wide `pnpm format:check` clean.

## Known limits

A browser that still runs an older cached bundle keeps sending the two removed names until it reloads; the server records them as before and nothing reads them. The server allowlist entries can be removed in a later cleanup, once stale bundles are gone.

One small loss: the removed upload event carried the AI badge state, so `upload_page_viewed` is no longer split by badge state. Nothing in the repo used that split.

## Browser check

Browser check: not applicable — two analytics calls removed, no rendered change; the page tests assert the events are no longer emitted while the page views still are.

## Changelog

No changelog entry — internal analytics change, nothing a user sees.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings: no reader, rate or funnel of either removed event anywhere in the repo (src, web, docs, SQL, scripts, workflows, locales), the removed effects did only call `track` so nothing became unused, an unknown event is recorded normally with one warning line rather than rejected, and the two new tests would fail if the events were re-added while the sibling tests still assert the two page views. Its non-blocking notes are folded in: the server allowlist now keeps the two names so a stale tab does not add warnings to the logs, and the PR body corrects three claims (an unknown event is not flagged in the row, the photo funnel query lists `photo_entry_point_viewed` in an informational section rather than using it as the denominator, and the two-fifths figure now says where it came from). The full server and web suites, `tsc -p . --noEmit` and the tree-wide format check pass on this head. Not a hard-rail change.

PR #4506, opened 2026-09-20
fix: stop the photo hint, limit page and occlusion link overpromising

aalemayhu:
## What

Four user-facing strings that overpromised now say what is true, in all 10 languages:

- **Photo drop hint** (`tools:photo.dropHint`): "Add as many as you need" becomes "Add several photos at once". Free is 5 photos a month and paid photos use AI credits, so there is no "as many as you need". The format list and size are separated by full stops instead of a dash.
- **Limit page label** (`accountx:limit.skipCap`): "Skip the cap for good" becomes "Skip the monthly card cap". The cap returns when a plan ends, and this names which cap.
- **Image occlusion upgrade links**: the free-plan count notice (`ImageQueue`) and the Notion import drawer (`NotionImportDrawer`) said "Upgrade for unlimited". They now use the existing `occlusion.upgradeToAddMoreImages` ("Upgrade to add more images"), and the now-unused `occlusion.upgradeForUnlimited` key is deleted from all 10 locales.
- **French occlusion strings** that said "Passe à l'illimité" (`upgradeToAddMoreImages`, `upgradeToAddMore`, `paywallImages`) now say "offre payante", matching the mind-map string fixed earlier. No other language had the claim.

## Why

Follow-up to the audit behind https://github.com/2anki/server/pull/4505. Paid image occlusion builds accept 20 uploaded images at a time (`upload.array('images', 20)` in `src/routes/ImageOcclusionRouter.ts`; free is 3), so "unlimited" was false, and the photo hint and limit label implied no cap where one exists. Serves **simpler**: a user is not surprised after upgrading. Metric: none moves directly; it is trust hygiene.

## Decisions (trio synthesis)

- **PM:** fix these now, split locale keys from the English copy, keep true scoped claims, and leave the image occlusion cap decision (reword versus raise) to Alexander. I reworded the upgrade link to say "add more images", which is true either way.
- **Designer:** "Skip the monthly card cap" keeps the purchase job without sounding apologetic; separating the hint with full stops removes the dash and the ambiguity about the format list; swapping to the existing "Upgrade to add more images" key needs no new translation.
- **Engineer:** the only tests pinning old strings are the limit page and wall tests (updated), and no other test reads the photo hint or the occlusion upgrade text. The credit-literal parity test and the pricing-page keys (`pricing.pass.imageOcclusion`, the comparison table cell, `pricingFaq.ts`) are rail paths, so they go in a separate rail PR for Alexander.
- **Guard:** `unlimitedClaims.test.ts` now covers `occlusion.paywallImages`, `occlusion.upgradeToAddMore` and `occlusion.upgradeToAddMoreImages` in all 10 locales. It failed on the old French text and I checked it fails again if the old claim is restored. The photo hint and limit label contain no unlimited word, so their guards are the component tests.

## Tests

- Tests first: the limit page and wall tests (new label), a photo page test (no "as many as you need"; "Add several photos at once."), and an image queue test (free plan count notice offers "Upgrade to add more images", no "unlimited" link) failed before the change.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). Full web suite: 424 files, all pass. `tsc -p . --noEmit`, web typecheck and tree-wide `pnpm format:check` clean. Locale JSON parses in all 10; no placeholders are involved.
- Native review of the eight non-English translations is still pending; the changes are short, direct translations of the English.

## Known limits

The Notion import drawer has no render test, so its link text is covered only by sharing the same key as the image queue. Not changed here: the pricing page's "Unlimited image occlusion" pass card and comparison table cell, and the "Paid plans lift the limits" subheading, which sit in rail paths and follow separately.

## Browser check

Browser check: not applicable — text-only changes to translated strings and two link labels, no layout or logic change; not walked in a browser.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. All 20 edited locale files parse and keep their trailing newlines with no reformatting noise; all 10 locales got the photo hint and limit label edits, the deleted `occlusion.upgradeForUnlimited` key is gone everywhere and has no remaining reader, the French occlusion edits match the earlier mind-map wording, no unlimited words remain in the changed keys, "Add several photos at once" is true (the photo input accepts multiple files with no per-batch cap), and the image queue test really exercises the free-plan count notice. Its non-blocking notes are folded into the head commit: the French skip label now says to pass over the cap instead of exceeding it, the Italian hint is more idiomatic, the changelog title no longer hints at prior brokenness, and the PR body cites the 20-file route. The full web suite (424 files), the web typecheck and the tree-wide format check pass on this head; the server is untouched since its last full run (758 suites, only the environmental `getPageCount` fails). Not a hard-rail change.

PR #4508, opened 2026-09-20
fix: pass descriptions name their AI credits, not full access

aalemayhu:
## What

The Day Pass and Week Pass were described as "full access" in four places that search engines, AI assistants and readers quote. They now say what a pass includes, with the credit allowance the code grants (`'24h': 300`, `'7d': 500` in `src/lib/claude/aiCredits/allowance.ts`).

- `web/index.html` Offer structured data: "24 hours of Pro features with 300 AI credits" and "7 days of Pro features with 500 AI credits".
- `web/public/llms.txt`: the same, in the pricing list.
- `reference/plans.md` (English) and its German mirror: the intro line, both section leads, and the "What you get" bullets now name the credits; the description reads "when a one-time pass makes sense".
- `hardcodedLimitClaims.test.ts`: a targeted guard that keeps "full access" and "Vollzugriff" out of those four files.

## Why

Follow-up to https://github.com/2anki/server/pull/4505 and https://github.com/2anki/server/pull/4506. "Full access" reads as no limits, but AI conversion, photos and chat draw on credits, so the claim overpromised. Serves **simpler** and more trustworthy pre-purchase copy. Metric: none moves directly; watch refund and churn-reason tags citing a credits mismatch.

## Decisions (trio synthesis)

- **PM:** name the credits rather than soften the phrase; keep the price mismatch between `llms.txt` ($4/$9) and `plans.md` ($6/$12) out of this diff because pass prices are Alexander's call.
- **Designer:** "Pro features with 300 AI credits" scans the same length as "full access" and tells a buyer what the pass unlocks.
- **Engineer:** the pricing FAQ (`pricingFaq.ts`) still says "full access" on `main`; it sits in a rail path and is fixed in https://github.com/2anki/server/pull/4507, so the guard is scoped to four files instead of a tree-wide scan that would fail here until that PR merges.

## What this does not change

Pass prices, the "Unlimited conversions" line in the plans table (a no-monthly-card-limit claim), and the German string quality beyond these lines.

## Tests

- Tests first: the guard failed on the old `llms.txt` (both lines quoted) before the change.
- Full web suite: 425 files, 3955 tests, all pass. Web typecheck and tree-wide `pnpm format:check` clean.
- Server: untouched, no server code reads these files (grepped `src` and `scripts`); the server suite was not re-run for this diff.
- Native review of the German lines is still pending.

## Browser check

Browser check: not applicable — text-only changes to structured data, llms.txt and two docs pages, covered by the guard test; not walked in a browser.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It confirmed the credit numbers match `src/lib/claude/aiCredits/allowance.ts` (300 for the Day Pass, 500 for the Week Pass), that "Pro features" is true because a pass sets the same paid flag a subscription does and every limit gate reads it (card limit, upload caps, PDF pages, print quota, image occlusion, mind maps, chat, templates), that both JSON-LD blocks in `web/index.html` still parse, that the German lines use the same "KI-Credits" term as neighbouring docs, that the guard test failed on the old text and scans only files that should not contain the phrase, and that no hard-rail path or user-identifying content is touched. Its one wording note (the changelog title hinting at earlier copy being wrong) is fixed in the head commit. Left as follow-ups, both pre-existing: the pass prices disagree between `index.html`/`llms.txt` ($4/$9) and `plans.md` ($6/$12), and the German plans table shows the Pro price differently from the English one; both are pricing decisions for Alexander. The full web suite (425 files, 3955 tests), the web typecheck and the tree-wide format check pass; the server is untouched and was not re-run.

PR #4509, opened 2026-09-20
chore: report upload network failures to the ops error log

aalemayhu:
## What

A browser-side network failure on an upload ("Failed to fetch", "Load failed") is now reported to the ops error log, so the Errors tab can see it.

- `reportUploadNetworkFailure` (`web/src/lib/reportClientError.ts`) reports under a bucketed message, `Upload network failure (<type>, <size range>)`, with the size ranges under 10 MB, 10-50 MB and over 50 MB. The type is a short extension or "unknown type"; the filename is never sent. Elapsed seconds and the browser's wording ride in the row's `context`, which is stored in `error_events` but not shown by the Errors tab yet (read it with a query). The failures group into a handful of rows instead of one row per user.
- `UploadForm.tsx` calls it from the existing upload catch, only for network failures. The `upload_failed` analytics event is unchanged.

## Why

`reportClientError` skips those messages as transient (a deliberate filter for background polls and deploy handoffs), and `upload_failed` lives in the `events` table, which no ops tab reads. So the Errors tab showed nothing for the upload failures while they were a steady share of started uploads. Serves **faster** indirectly: it gives the upload-path investigation real timing data. Metric: the Errors tab shows the `Upload network failure (...)` groups within a day of deploy, read at `/ops/errors`; the elapsed-seconds split is a query on `error_events.context`.

## What the existing telemetry says

Read-only prod aggregates from `upload_failed` over 14 days: every failure is `reason: network`; the daily failure share is steady with no deploy-time bursts (only four single minutes with three failures); and failures are spread across sizes, mostly under 10 MB. That weakens the hypothesis that large uploads time out on the long synchronous request. `upload_started` carries no size, so a per-size failure rate is not computable yet; the elapsed seconds added here will show whether failures are instant drops or timeouts.

## Decisions (trio synthesis)

Internal ops tooling, trio optional; not run.

- **Not done here:** resolving the two stale error groups (an ops-tab click, not code), and hiding dead ops panels (which ones are dead is not established). Adding size and extension to `upload_started` for a real failure rate is a possible follow-up.
- **Known limits (from review):** a reload or tab close mid-upload can also throw a network error, so some rows are abandonment; `upload_failed` already counts the same population. The server dedupes by message hash and IP for five minutes, so a row count means distinct episodes, not failures; `upload_failed` stays the real count. The extension check accepts any short alphanumeric suffix, not only known upload types.
- **No changelog entry:** internal ops visibility, no user-visible behavior change.

## Tests

- Tests first: `reportUploadNetworkFailure` tests (skipped-as-transient cause now reported, bucketed messages, elapsed seconds and cause in context, silent when offline, unsafe extension replaced) and `UploadForm` tests (a network failure with a selected file posts its type and size range; none selected posts unknown type and size; a non-network error posts nothing) failed before the change.
- Mutation check: forcing the network gate to always report fails the non-network `UploadForm` test.
- Full web suite: 425 files, all pass (3964 tests before the last test-only commit; the two changed files re-run green). Web typecheck and tree-wide `pnpm format:check` clean.
- Server: untouched (the ingest controller already accepts any message and context; no server-side filter drops these); the server suite was not re-run for this diff.

## Browser check

Browser check: not applicable — no visible change; the report is a background request covered by component tests; not walked in a browser.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It confirmed the bucketed message clears every filter in `reportClientError` (offline still suppresses it, as intended), that the message carries only a validated extension and a size range with no filename, that the server ingest controller and use case do not filter on message content (they dedupe by hash and IP for five minutes and rate-limit at 10 a minute), that the `UploadForm` catch refactor keeps the same extension and size behavior for no dot, several dots and no file, that no hard-rail path or user-identifying content is touched, and that no changelog is needed for an internal ops change. Its non-blocking notes are handled in the head commit and the PR body: the Errors tab does not display `context`, so the elapsed-seconds claim now says to read it with a query; a reload or tab close mid-upload can also throw a network error and the server dedupe means a row count is distinct episodes, both now listed as known limits; the negative `UploadForm` test now waits for the failure event before asserting, and a selected-file test covers the extension and size derivation. Left as follow-ups: a visibility-state field in the context, a known-extension allowlist, and showing `context` in the Errors tab. The full web suite (425 files) passes on the previous head, the changed test file and the web typecheck and tree-wide format check pass on this head; the server is untouched and was not re-run.

PR #4511, opened 2026-09-20
fix: free plan copy states the anonymous 21-card cap

aalemayhu:
## What

The free plan was described as "100 cards per month, no account required" in three places that search engines, AI assistants and readers quote. Without an account a single conversion is capped at 21 cards (`ANONYMOUS_CARD_CAP` in `CheckMonthlyCardLimitUseCase.ts`); the 100 a month (`MONTHLY_CARD_LIMIT`) needs a free account. They now say both, in the same words the docs already use.

- `web/index.html`: the Free offer description and the site description in the structured data.
- `web/public/llms.txt`: the Free line in the pricing list.
- `web/src/pages/AnswersPage/answersConfig.ts`: the PDF answer said "No account required for PDFs under the free plan limit".
- `hardcodedLimitClaims.test.ts`: a targeted guard that keeps "no account required" out of those three files.
- Changelog entry, since the answers page text is user-visible.

## Why

Follow-up to https://github.com/2anki/server/pull/4505 and https://github.com/2anki/server/pull/4508. A visitor who reads "no account required" next to "100 cards" and drops in a 60-card file is refused after the conversion, which is the surprise these audits are removing. Serves **simpler** and more trustworthy first contact. Metric: none moves directly; watch the `anonymous_cap` share of `conversion_failed` (read in the ops Conversions tab) for a decline over the next weeks.

## Decisions (trio synthesis)

- **PM:** state both numbers rather than drop "no account"; the no-account path is a real strength, only the number attached to it was wrong.
- **Designer:** the docs phrasing ("without an account ... 21 cards; a free account ... 100 a month") is already the house wording, so reuse it.
- **Engineer:** the same claim also sits in the pricing FAQ (`pricingtable.json` in all locales and `pricingFaq.ts`), which are hard-rail paths; that is a separate rail PR for Alexander, so the guard is scoped to the three files here instead of a tree-wide scan that would fail until then.

## What this does not change

The pricing page FAQ answer (rail, separate PR), and whether 21 is the right anonymous cap (product decision).

## Tests

- Tests first: the guard failed on the old `index.html` before the change.
- Full web suite: 426 files, 3985 tests, all pass. Web typecheck and tree-wide `pnpm format:check` clean.
- Server: untouched; the server suite was not re-run for this diff.

## Browser check

Browser check: not applicable — text-only changes to structured data, llms.txt and one answers entry, covered by the guard test; not walked in a browser.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It traced the anonymous cap in `UploadService.ts` (a conversion that produces more than 21 cards without an account is rejected, per conversion, with no daily or lifetime counter) and the monthly limit of 100 for signed-in free users across the upload, async, Notion, chat and MCP paths, and confirmed the new wording matches the docs (`help/limits.md`, `start-here/upload-a-file.md`, `what-is-2anki.md`, English and German). Both JSON-LD blocks in `web/index.html` still parse. The guard test covers exactly three files, fails on all three old texts and passes on the new, and leaves the pricing FAQ files alone on purpose. The diff touches no hard-rail path and no user-identifying content, no test pins the old answers-page sentence, and the changelog id, type and date are valid. The only remaining places that still say "No account required for one-off conversions" next to 100 cards are `pricingtable.json` in the ten locales and `pricingFaq.ts`, which are rail paths and follow in a separate PR for Alexander. Its polish note on the changelog title is fixed in the head commit; a guard for "no account needed" wording variants is left as a possible follow-up. The full web suite (426 files, 3985 tests), the web typecheck and the tree-wide format check pass; the server is untouched and was not re-run.

PR #4514, opened 2026-09-20
chore: keep Node connections open longer than the Apache proxy

aalemayhu:
## What

Two lines in `src/server.ts` (through a small tested module): the Node HTTP server now sets `keepAliveTimeout` to 65 s and `headersTimeout` to 66 s. Node's default `keepAliveTimeout` is 5 s and nothing in the repo overrode it.

## Why

About 3 to 4% of started uploads fail in the browser with a network error ("Failed to fetch" / "Load failed"), every day, across file types and mostly on small files, with no burst at deploy time (read from the existing `upload_failed` events over 14 days). Anonymous visitors fail more often than signed-in users (roughly 5% against 2%), on both Chromium and Safari.

What I verified on the box, read-only:

- The site is fronted by Apache (`Server: Apache`), listening on 443 (h2 enabled) and proxying to Node on port 3001 with a bare `ProxyPass` (no `ttl`).
- Apache's `KeepAlive` is On with `KeepAliveTimeout 5` (browser-to-Apache). Node's `keepAliveTimeout` defaults to 5 s plus a 1 s buffer on Node 22.20 and nothing in the repo overrode it.
- Apache pools backend connections and cannot retry a POST that hits a connection Node has just closed, which the browser sees as a network failure. Raising Node's limit moves that race window from about 6 s to about 66 s, where far fewer request gaps land.

What I did not verify, and the reviewer flagged: Apache's error log is not readable by my user, so I have not seen the reset lines. A 3 to 4% failure rate is higher than a loopback race alone would normally give, so the browser-to-Apache 5 s leg (shared by the anonymous and signed-in paths) may matter as much. Treat this as a low-cost mitigation, not a diagnosed fix. The check for you: `grep -c "AH01102\|error reading status line" /var/log/apache2/error.log`, or `502` on `POST /api/upload/file` in `access.log`. If those are rare, this change will not move the number.

A structural follow-up, not in this PR: `ttl=` (under 60) on the Apache `ProxyPass` line written by `scripts/deploy-blue-green.sh`, so Apache retires pooled backend connections before Node would. That file is a deploy-pipeline rail and needs your call.

This is a hard-rail path (`src/server.ts`), so it waits for a manual merge. It serves **more stable**: the metric is the share of `upload_failed` events with `reason: network` among `upload_started`, currently steady at roughly 3 to 4% a day. Read it 3 to 7 days after deploy. If it does not drop, the cause is elsewhere and this change is harmless to keep.

## What this does not change

No timeout is lowered or removed; `requestTimeout` (Node's default 300 s) is untouched. Idle sockets are held a little longer, which Node closes on shutdown (`gracefulShutdown` closes idle connections first and severs the rest after 2 s).

## Tests

- Tests first: `httpServerTimeouts.test.ts` failed (module missing), then passes: keep-alive above Apache's 5 s, headers timeout above keep-alive.
- The `server.ts` wiring itself is one call and is not unit-tested.
- Full server suite: 759 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). `tsc -p . --noEmit` and tree-wide `pnpm format:check` clean. Web untouched, not re-run.
- No changelog entry: `chore:`, not claiming a user-visible fix until the number moves.


aalemayhu (commented):
Review verdict: pass (rail PR, waiting for Alexander's manual merge). A fresh read-only reviewer found no blocking findings. It verified on the pinned Node 22.20 that `keepAliveTimeout` and `headersTimeout` are plain properties that can be set after `createServer` and before `listen`, take effect for new connections, and do not conflict with `requestTimeout` (300 s, untouched) or `server.timeout` (0). Nothing else in `src/`, `scripts/` or `.github/` sets these, graceful shutdown still finishes promptly because it closes idle connections first, and a longer keep-alive adds no drain time to the blue-green cutover. Its notes are folded in: a why-comment now explains that the values sit above the proxy's idle limit, and the PR body was rewritten to state only what was verified. The mechanism is honest about its limits: Apache's error log is unreadable to the reviewer and to the author, so the reset lines have not been seen, the effective Node default is about 6 s (5 s plus a 1 s buffer), and a 3 to 4% failure rate is higher than a loopback race alone would give, so this is a low-cost mitigation to be judged by the `upload_failed` network share 3 to 7 days after deploy, not a diagnosed fix. Suggested follow-up, not in this PR: a `ttl=` on the Apache `ProxyPass` in `scripts/deploy-blue-green.sh` (a deploy-pipeline rail). Full server suite (759 suites, only the environmental `getPageCount` fails), `tsc -p . --noEmit` and the tree-wide format check pass; web untouched.

PR #4515, opened 2026-09-20
fix: quote the fetch column in the blocks cache counter update

aalemayhu:
## What

The blocks cache hit counter (`BlocksCacheRepository.bumpFetch`) updated `fetch` with a raw `fetch + 1`. `fetch` is a reserved word in Postgres, so the unquoted right-hand side is a syntax error. Every counter update has failed since https://github.com/2anki/server/pull/4501 deployed, and the failure is only logged as a warning. The right-hand side is now a quoted identifier, built by one exported helper (`incrementFetchCounter`) that the repository uses.

## Evidence

- Production error log (pm2, both colors): the same warning `[blocks-cache] fetch counter update failed: update "blocks" set "fetch" = fetch + 1 ... - syntax error at or near "fetch"` repeated thousands of times on 2026-09-20 and is by far the largest single entry in the log; first occurrence shortly after the deploy of #4501.
- Parse-only check against production Postgres (`PREPARE`, nothing executed): the shipped statement fails with `syntax error at or near "fetch"`; the fixed statement (`set "fetch" = "fetch" + 1`) prepares cleanly.
- Why it shipped green: the existing counter test runs against SQLite, which accepts the bare form. `.claude/rules/testing.md` already describes this trap (mocked or SQLite tests cannot reproduce Postgres-only syntax).

## Why

Serves **more stable and honest data**: the counter that decides whether the blocks cache earns its keep (PLAN row 12) has never counted a hit, every cache hit now fires an update that cannot succeed (an extra pool connection and a warning line each time), and the log is drowned in one line. Metric: the `[blocks-cache] fetch counter update failed` line goes to zero after deploy (`grep -c` on the pm2 error log), and `blocks.fetch` starts rising for re-read pages.

## Tests

- Tests first: a Postgres-dialect knex (`knex({ client: 'pg' })`, no connection) asserts the generated SQL is exactly `update "blocks" set "fetch" = "fetch" + 1 where ...` (it first failed because the helper did not exist), and a second test captures the SQL the repository emits on a cache hit and failed on the old code's bare `fetch + 1`. The listener is removed in a `finally`.
- The existing behavioural counter tests still pass.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). `tsc -p . --noEmit` and tree-wide `pnpm format:check` clean. Web untouched.
- A scan of the other raw SQL expressions in `src/` found no other unquoted reserved word.
- No changelog entry: internal, no user-visible behavior change.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It verified with knex 3.3.0 (no DB connection) that `raw('?? + 1', ['fetch'])` renders `"fetch" + 1` on pg and a quoted identifier on better-sqlite3, that `update({ fetch: <raw> })` renders `set "fetch" = "fetch" + 1` on pg while the old `raw('fetch + 1')` rendered the invalid bare form, and that the `??` idiom is already used elsewhere in the repo. It found no other instance of the bug class (the other raw increments use non-reserved column names) and confirmed `blocks.fetch` is `integer NOT NULL DEFAULT 0` with a unique index on `(object_id, owner)`, so each update is a single-row indexed write. On behaviour, the fix turns one failing statement per cache hit into one successful single-row update, which is negligible extra load and cannot break a conversion because `bumpFetch` stays fire-and-forget with a rejection handler. The parse-only check on production Postgres (shipped statement fails with a syntax error, fixed statement prepares) is in the PR body. Its test notes are folded in (the query listener is now removed in a `finally`, and the PR body no longer overstates why the first test failed). No rail path and no user-identifying content. The full server suite (758 suites, only the environmental `getPageCount` fails), `tsc -p . --noEmit` and the tree-wide format check pass; web untouched. Morning check after deploy: `grep -c "fetch counter update failed"` on the pm2 error log should go to zero.

PR #4516, opened 2026-09-20
chore: record usage events for mind maps, image occlusion, apkg import

aalemayhu:
## What

Three surfaces had no usage event, so a keep-or-remove verdict (CLAUDE.md "Surface lifecycle") was impossible for them. One event each, at the point of success, with counts only:

- `mindmap_created` (client, `MindmapList.tsx`): after a new map exists. No props.
- `image_occlusion_created` (server, `ImageOcclusionController.create`): after the deck is built, before it streams. Props: `image_count`, `occlusion_count`. Guest builds carry no user id.
- `apkg_imported` (server, `ImportApkgToNotionUseCase`): when the import job is marked done. Props: `note_count`, `truncated`.

Only the client event is added to both `KNOWN_EVENTS` lists (web and server); the two server-emitted events, like `mcp_tool_called`, do not go through that allowlist.

## Why

Alexander has pending keep-or-remove decisions and these surfaces are unmeasured (prod shows no event for them in the last 30 days apart from a rare mindmap excluded-nodes event). Serves none of simpler, faster, more beautiful directly: it is measurement. It earns its place because the alternative is deciding removals with no data. Metric and read: after 30 days (2026-10-21) count distinct users per event; a low count is the input to the keep-or-remove call, which stays Alexander's. A proposed bar from the PM view is 20 distinct users, offered as a starting point, not a decision.

## Decisions (trio synthesis)

- **PM:** instrument only the three surfaces with a real success moment; skip the homepage showcase (marketing, not a surface), the Dropbox chooser (on the removal row) and Google Drive import (a thin wrapper on `upload_started`).
- **Designer:** nothing user-visible changes; `track` posts with `keepalive`, so an event fired before a redirect is not lost; no privacy copy change while props stay counts.
- **Engineer:** client one-liner for mindmap create; the controller for occlusion (the use case does not know the user); the use case for apkg import (the client only sees completion through polling and would double-fire).
- **Conflict, resolved:** the engineer also offered a mindmap export event and a Drive/Dropbox picked event. Left out on purpose to keep the set minimal and to not measure what may be removed.

## Known limits

- Guest image occlusion builds are recorded without a user id, so distinct-user counts undercount guests.
- The image occlusion event fires after the deck is built, not after the browser saves the file.

## Tests

- Tests first: the new tests failed before the change (the mindmap test on a missing call; the controller and use case tests on a missing event); the "records nothing" cases (limit reached, failed import, cap refusal) guard against firing on failure.
- Full server suite: 758 suites pass; the only failure is `getPageCount` (needs `pdfinfo`, environmental). The server `test` job failed twice on an earlier head (once in CI, once in the merge queue) in `NotionAPIWrapper.forbiddenBlocks.test.ts` with `this.repository.insertEvents is not a function`. Cause, found and fixed here: `ImageOcclusionRouter.test.ts` mocks `EventsRepository` as an empty object, and the new event made its success path create the real events sink on top of it, so the sink's 5 s flush timer later failed in a slower suite. The router test now mocks the sink instance like the other route tests and asserts the event is recorded through the route.
- Full web suite: 427 files, 3999 tests pass. Both typechecks and tree-wide `pnpm format:check` clean.
- No changelog entry: internal instrumentation, no user-visible change.

## Browser check

Browser check: not applicable — a background analytics call after an existing action, no visible change, covered by a component test; not walked in a browser.


aalemayhu (commented):
Review verdict: pass. A fresh read-only reviewer found no blocking findings. It confirmed each event is recorded only on success: `image_occlusion_created` after the deck is built and before it streams (a 400, a coded error such as the free-tier image cap, or an unexpected error all return or rethrow before the record call), `apkg_imported` right after the job is marked done inside the try (so the record call cannot mark a finished job failed), and `mindmap_created` only when the create call resolves with an id (not when the free-limit modal shows). `getEventsSink().record` is synchronous and non-throwing, owners are numeric ids or absent in production, and the props are counts and booleans only with no filenames, titles or note text. Server-emitted events correctly stay out of the client allowlist, like `mcp_tool_called`, while the client event is in both lists and the parity test passes. Its one hardening note is folded into the head commit: the events sink flushes one multi-row insert, so a non-numeric user id would fail the whole batch and drop other events; the controller now uses `getOwnerId` and the import use case falls back to null, each with a test, and the mindmap test asserts exactly one call. Left as optional follow-ups: switching to the `track()` helper the rest of the server uses, `mkdtemp` for the test temp files, and the same unguarded `Number(user.id)` in `McpRouter.ts`. No hard-rail path, no user-identifying content, no changelog needed for internal instrumentation. The full server suite (758 suites, only the environmental `getPageCount` fails), the full web suite (427 files, 3999 tests on the earlier head, the changed test file re-run), both typechecks and the tree-wide format check pass.

aalemayhu (commented):
Review verdict: pass, carried over from the fresh review of the previous head (fac3a715bb8e) plus a delta check. Since that review the branch was rebased onto main (after https://github.com/2anki/server/pull/4515 merged) and one test-only commit was added: `ImageOcclusionRouter.test.ts` now mocks the events sink instance and asserts that one `image_occlusion_created` event is recorded for the signed-in user through the real route. That fixes the cause of two server `test` failures on this PR (a leaked real sink built on top of an empty `EventsRepository` mock, whose 5 s flush timer failed a slower suite), and it adds route-level coverage of the new event. No production code changed in the delta. The earlier findings all still hold (events recorded only on success, non-throwing sink, numeric or absent owner with a `getOwnerId` guard, counts-only props, allowlist parity, no rail path, no user-identifying content). The full server suite (758 suites, only the environmental `getPageCount` fails), `tsc -p . --noEmit` and the tree-wide format check pass on this head.

PR #4517, opened 2026-09-21
fix: Notion picker starts on your shared pages, not a search for anki

aalemayhu:
## Summary

The `/notion` picker seeded its search box with `anki`. Notion search matches page titles only, so a connected user saw just the pages with "anki" in the title and read every other page they had granted as missing. A real support report this week: three pages granted in Notion's consent screen, one shown.

- The box now starts empty. An empty box is served by the existing cached top-level pages endpoint (drops database rows, caps at 50, stops early, 60s in-memory plus DB tier), not the uncached full `search` that can make up to 20 sequential Notion calls and return every database row. Typed queries keep the full title search.
- A response for an old query can no longer land after the box changes or is cleared.
- Both post-login redirects drop the `q=anki` seed and go to `/search`. The now-unused `getSearchPath` helper is deleted.
- When the box is empty and pages are shown, the picker reuses the existing `import.pagePickerHelp` line ("Missing a page? Check your Notion sharing settings"). No new strings, no new translations.
- `connectToNotion` (OAuth connect and re-consent) now clears the owner's cached top-level pages, in memory and in the DB, before the fire-and-forget pre-warm. Before this, a user who had just granted new pages could still read the old cached set. It mirrors what `disconnect` already does, and a failed cleanup is logged and does not fail the connect.
- The web mock server gets a matching `/api/notion/top-level-pages` route.

Which of simpler / faster / more beautiful: simpler (the box shows your pages) and faster (one cached call instead of up to 20 for an empty box). Metric: Notion `conversion_succeeded` per connected user and "pages not showing" support tickets, read at day 7 and day 30.

Deliberately left out after the trio review: the placeholder change (10 locales), removing the sessionStorage-saved query, virtualization, a refresh button.

## Testing

- New `NotionService.connectToNotion.test.ts`. The reconnect test failed before the fix (old page set returned after reconnect) and passes now.
- `useSearchQuery.test.tsx`: empty default, empty box uses the top-level endpoint, typing uses `search`, spaces-only counts as empty, stale typed response is ignored after clearing. Redirect tests updated, one added for the login default, new `SearchPresenter.test.tsx` for the help line.
- Full web suite: 4007 passed. Full server suite: 7588 passed, only the known environmental `getPageCount` (`pdfinfo`) failures.
- `/check` clean (server build, arch, web typecheck, web lint, web tests). `pnpm format:check` clean tree-wide.
- Security review of the branch diff found nothing introduced by this change. The `useHandleLoginSubmit` files match the hard-rail list (login), so this PR waits for a manual merge. The OAuth connect change itself is not a rail path.

## Browser check

Notes:
- `pnpm --filter 2anki-web test:golden-path`: 4 passed at 375px, no console errors (landing, signed-in dashboard, account pages).
- Walked `/notion` at 375px against the mock API (with the connect-link response overridden to report a connected workspace): the empty box posts `top-level-pages` with an empty query and lists the page under the sharing help line. Typing "Sample" posts to the full `pages` search and hides the help line. Clearing the box returns to `top-level-pages`, the help line and the URL without `?q=`.
- That manual walk logged console errors, all 404s from routes the mock server does not implement (favorites, usage, preferences, pricing, event tracking, AI credits) plus the Vite HMR websocket. None come from the changed code. The mock server's `get-notion-link` also returns `connected` where the app reads `isConnected`, so the picker only renders with that override. Not fixed here.


aalemayhu (commented):
**Verdict**: approve

Simpler and faster: the /notion picker shows the pages a user granted instead of a hidden search for "anki", and the empty box costs one cached call instead of up to twenty.

### Blocking
- None. Security, engineering and ux/voice forks each found nothing at confidence 7 or above.

### Nits
- None.

Hard-rail path (login redirect default in useHandleLoginSubmit), so this waits for a manual merge.

PR #4519, opened 2026-09-21
chore: drop Write permission rules Claude Code never matches

aalemayhu:
## What

Remove five `Write(...)` allow rules from `.claude/settings.json`:

- `Write(src/**/*.test.ts)`
- `Write(web/src/**/*.test.ts)`
- `Write(web/src/**/*.test.tsx)`
- `Write(Documentation/specs/*.md)`
- `Write(Documentation/ankify/*.md)`

## Why

Claude Code only matches `Edit(path)` rules; `Edit` covers all file-editing tools including Write. Every session start prints one warning per rule: "is not matched by file permission checks — only Edit(path) rules are."

No access changes. Each removed rule is already covered by an existing `Edit` rule:

- the three test-file rules by `Edit(src/**)` and `Edit(web/src/**)`
- the two Documentation rules by their identical `Edit(Documentation/...)` twins

## Goal check

Does none of simpler/faster/more beautiful for flashcard creation: it is harness config only. It earns its place by silencing five startup warnings on every session.

## Checks

- `.claude/settings.json` still parses as JSON; `oxfmt --check` clean.
- Hard-rail path (`.claude/`): needs manual merge, not `/ship`.


aalemayhu (commented):
**Verdict**: approve. Security and engineering reviews found no issues. The five removed Write(...) allow rules are each covered by an existing Edit(...) rule, so access is unchanged; only allow entries were removed, no hooks or deny rules were touched, JSON still parses, and the branch is one commit directly off main. Serves simpler: five startup warnings gone. Hard-rail path (.claude/settings.json), so it waits for a manual merge.

PR #4520, opened 2026-09-21
feat: deliver the first 21 cards to anonymous visitors over the cap

aalemayhu:
## What

Behind a default-OFF DB flag (`anonymous_partial_delivery`), an anonymous single-file upload that produces more than 21 cards returns a 200 apkg with the first 21 cards and an `X-Cards-Held-Back` header, instead of being refused and redirected to `/limit?kind=anonymous`. The upload result shows a neutral notice with the held-back count that tells the visitor to create a free account and drop the file again. There is no server-side unlock: anonymous output is streamed once and never stored.

## Why

The free-tier wall is the main leak in `conversion_failed`; real conversion failures are a small share. An anonymous visitor does the upload, waits, and gets nothing. Signed-in free users over the monthly limit already get a partial deck on the job-queue path.

## How

- **Server:** `cardLimit` flows from `UploadService.handleSyncUpload` through `GeneratePackagesUseCase` into the worker and `DeckParser`, which truncates with `truncateDecksToCardLimit` after card transforms so exactly 21 cards ship. The apkg is built inside the worker before the cap check, so a re-run was not needed.
- **Arm:** flag on splits by `sha256(anonymous id)`; no anonymous id gets control; signed-in users and flag OFF are untouched.
- **Eligibility (added in review):** only a single non-zip file of a parser-backed format is eligible (`isPartialDeliveryEligible`). Zip and multi-file uploads keep today's refuse path and are tagged with no arm. Without this a multi-file treatment upload would have been truncated in a batch response with no notice.
- **Events:** treatment lands on `conversion_succeeded` with `card_limit_partial`, `cards_held_back`, `arm`, and a card-count bucket. Control cap-hits carry `arm` and the same bucket. Two client events, `anonymous_partial_notice_shown` and `anonymous_partial_signup_clicked`, are in both `KNOWN_EVENTS` lists. `PAYWALL_REASON_PATTERNS` is untouched.
- **Web:** `AnonymousPartialNotice` beside the success state, `role="status"`, 10 locales with plural forms, signup goes to `/register?redirect=/upload`.
- **Migration:** data-only seed of the flag row (the Ops flags tab only toggles existing rows). No schema change, so no kanel run. Migration review: ship it, no must-fix.

## Measuring success

Read in prod `events` after flipping the flag. Day 7: signup-start rate among cap-hit anonymous ids, by `arm`. Week 6: pay rate per cap-hit, by `arm`. Roll back if signup-start falls below control at day 7 or the pay rate trends below control. Cut by card-count bucket, since decks just over 21 cards are where a partial nearly satisfies the visitor. Zip and multi-file cap-hits carry no arm and are excluded from both arms.

## Testing

- Failing tests first for the eligibility rule, then the fix. Real generation test for truncation (`DeckParser.cardLimit.test.ts`), controller-level tests for flag off, treatment, control, at-cap, no anonymous id, zip and multi-file.
- Full server Jest: 7624 passed, 2 failed (`getPageCount` needs `pdfinfo`, the documented environmental failure), 2 suites skipped (`PythonExitError` venv).
- Full web vitest: 4043 passed. Server `tsc -p .` clean, web typecheck clean, `pnpm format:check` clean, `depcruise` 0 errors.
- Designer review before ready: no must-fix. Review fan-out: security clean; engineering nits addressed (dead multi-file budget removed, PrepareDeck and reversed-card tests added); UX and voice blocker addressed (changelog entry removed).

## Changelog

no changelog entry — the feature ships dark behind a default-OFF flag and later reaches only half of anonymous single-file uploads, so a What's New line would be untrue at merge. The entry ships with the PR that rolls the flag out (a review finding; the entry was removed from this PR).

## Risks and known gaps

- **Zip uploads are not covered** by this change, and zips were the largest single format at the wall. Threading a running card budget through `getPackagesFromZip` (several `PrepareDeck` call sites) is a separate change: https://github.com/2anki/server/issues/4522.
- **Cannibalisation:** some visitors who would have paid may take the 21 cards and leave. The week-6 pay-rate guardrail is the detector; do not declare a win on signups alone.
- **Signed-in sync path has the same dead end.** `handleSyncUpload` runs `CheckMonthlyCardLimitUseCase` for signed-in owners and throws instead of truncating, while the job-queue path truncates. Not changed here: https://github.com/2anki/server/issues/4523.
- **Contrast:** the filled signup button is white on `--color-primary`, about 3.7:1 in the light theme. It matches the shared primary button, so it is a token-level follow-up.
- **ru and pl headline** grammar ("21 карточка готовы") needs a native-speaker pass.
- **Hard rail:** limits path plus a migration. This PR waits for a manual merge; the post-deploy steps are in https://github.com/2anki/server/issues/4521.

## Browser check

Not run. This checkout has no dev server running and the notice only renders when the flag is ON and the visitor is in the treatment arm. To verify locally after the migration is applied: set `feature_flags.value = true` for `anonymous_partial_delivery`, upload a single markdown or HTML file that makes more than 21 cards without signing in (use an anonymous id whose hash byte is even), and check the notice at 375px and the console.

## Goal alignment

Faster: a deck arrives immediately instead of a dead-end redirect. Simpler: one fewer wall. Metric: signup-start rate among cap-hitters at day 7, pay rate per cap-hit at week 6, both in prod `events`.

Spec: recoverable from the `docs: add spec for anonymous partial delivery` commit on this branch.


aalemayhu (commented):
**Verdict**: approve. Faster and simpler: an anonymous visitor over the 21-card cap now gets a deck instead of a dead-end redirect, behind a default-OFF flag. Round 1 fan-out: security clean, engineering nits fixed (dead multi-file budget removed, PrepareDeck and reversed-card tests added), UX and voice blocker fixed (changelog entry removed, out-clause in the body). Round 2 fan-out on this head: security, engineering and UX and voice all clean. Hard-rail PR (migration plus limits path), so it waits for a manual merge.

PR #4525, opened 2026-09-22
chore: log pdftoppm kill signal alongside exit code

aalemayhu:
## What
Captures the `signal` argument alongside `code` in the `pdftoppm` `close` handler in `src/lib/pdf/convertPage.ts`, and folds it into the rejection message using the same `code=… signal=…` format the sibling `getPageCount.ts` already uses.

## Why
When a PDF-to-image page conversion is killed, the close handler read only the exit `code` and dropped the `signal` Node passes on that event, so prod logged `pdftoppm process exited with code null` with no way to distinguish a `SIGKILL` (OOM) from a `SIGTERM` (timeout) from a plain non-zero exit. That is the exact information needed to triage a killed conversion, and today it is unrecoverable after the fact.

## How
The error message is the only field that survives the conversion worker serialization boundary: `src/usecases/uploads/worker.ts` copies just `message`/`name` into the `UploadGenerationFailure`, and `buildWorkerError` in `GeneratePackagesUseCase.ts` rebuilds a plain `Error` from `message`. A structured property on the thrown `Error` would be dropped crossing that boundary, so the signal is embedded in the message. Format mirrors the sibling `getPageCount.ts` (`code=${code ?? 'null'} signal=${signal ?? 'none'}`) so existing `signal=` log greps and tooling apply unchanged.

Pure diagnostics: no change to conversion behavior, retry logic, or timeouts. No prod safety limit touched.

## Measuring success
The next killed page conversion logs `pdftoppm process exited with code=null signal=SIGKILL` (OOM) or `signal=SIGTERM` (timeout) instead of the ambiguous `code null`. Confirm in prod by grepping the conversion worker logs for `pdftoppm process exited with code=` and reading the `signal=` field.

## Testing
- Unit tests added: `src/lib/pdf/convertPage.test.ts` — mocks `child_process.spawn` (external edge only) and asserts the signal surfaces in the rejection for `SIGKILL` and `SIGTERM` (code null), that a non-zero exit with no signal reads `code=1 signal=none`, and that a clean exit still resolves to the page image path.
- Scoped run (this is a narrow, non-shared-hot-path change — not the parser core): `pnpm test src/lib/pdf/convertPage.test.ts` → 4/4 pass. `pnpm exec tsc -p . --noEmit` clean. `pnpm format:check` (tree-wide) clean.
- `pnpm test src/lib/pdf/` shows 2 failures in `getPageCount.test.ts` — the documented environmental `pdfinfo` failures (see `.claude/rules/testing.md`); that file is untouched by this PR.

## Browser check
Not applicable — server-side diagnostics only, no rendered output.

## Changelog
No entry — internal diagnostics only, invisible to users (per `.claude/docs/changelog.md`).

## Risks
Minimal. The only behavioral change is the text of one error message on the already-failing path; the resolve path and thrown-on-error semantics are unchanged. Rollback is a one-line revert.

## Goal alignment
None of simpler / faster / more beautiful directly — internal diagnostics. It earns its place by making a killed-conversion failure diagnosable from logs alone, which shortens the path to fixing conversions that OOM or time out (indirectly protecting "get a clean deck back").

## Trio
Skipped — pure instrumentation, no user-visible behavior change and no design decision.


aalemayhu (commented):
**Verdict**: approve. Pure diagnostics — captures the pdftoppm kill signal alongside the exit code so a future OOM-vs-timeout crash is diagnosable from logs. Security: no issues (signal is a runtime enum, not user input; no argv/spawn change). Engineering: no issues, format genuinely matches the sibling getPageCount.ts convention (verified), TDD confirmed real, no prod safety limit touched, mock is external-edge only.

PR #4526, opened 2026-09-22
chore: add chat_message_sent usage event

aalemayhu:
## What
Adds a single new analytics event, `chat_message_sent`, fired once per user-submitted chat message from the one clean submit point in `ChatPanel.sendMessage`. The event carries one minimal, content-free prop, `is_new_conversation`, and is mirrored into both the web and server `KNOWN_EVENTS` allowlists.

## Why
The Chat surface has no per-message usage event today — the only allowlisted chat event is `chat_all_conversations_deleted`, which measures deletion, not active use. There is a `chat_messages` DB table showing real activity, but no lightweight analytics event mirrors it, so a keep/remove verdict on the surface can't be read from event data alone. This event closes that gap.

No trio: pure instrumentation addition, no design decision, no UI change.

## How
- `track('chat_message_sent', { is_new_conversation })` fires in `sendMessage` immediately after the empty-content guard, before the API request. This is the single point where a user's message is actually submitted (composer send button and Enter both route through it).
- Deliberately NOT fired on: streamed tokens, template-change regenerate (`regenerateLastTurn` is a separate function), or re-renders. Each manual re-submit after a failure counts as one new user-sent message, which is the intended semantics — a retry is a fresh submission by the user.
- `is_new_conversation` = `activeConversationId == null`, i.e. this message starts a brand-new conversation with no server row yet. Combined with the total count, this gives conversations-started vs. continued-turn ratio.
- No message content, conversation content, or any user text is included in the props — a business rule (`.claude/rules/security.md`: never log sensitive user content) as much as a privacy one.
- Added the identical string to both `web/src/lib/analytics/events.ts` and `src/types/AnalyticsEvents.ts` in this PR; `events.parity.test.ts` stays green.

## Measuring success
Query the events store for `name = 'chat_message_sent'` grouped by day, and split by the `is_new_conversation` prop. A non-zero, sustained daily count is the usage signal for the surface's keep/remove verdict; `is_new_conversation=true` counts new conversations started, and `false / true` ratio approximates messages-per-conversation.

## Testing
- Unit tests added (`web/src/components/ChatPanel/ChatPanel.test.tsx`, new `ChatPanel — message sent analytics` describe):
  - fires `chat_message_sent` exactly once with `{ is_new_conversation: true }` on the first send of a new conversation
  - fires exactly once with `{ is_new_conversation: false }` when sending into an existing conversation
  - does NOT fire on a template-change regenerate
- Parity test `web/src/lib/analytics/events.parity.test.ts` passes (event present in both allowlists).
- `pnpm --filter 2anki-web typecheck`, `pnpm exec tsc -p . --noEmit`, and `pnpm format:check` all green.

## Browser check
Browser check: not applicable — event-only instrumentation, no rendered output change. The sole source change is a `track()` call in an existing submit handler; no new strings, elements, copy, or layout.

## Changelog
No entry — internal analytics instrumentation, no user-visible change.

## Risks
Minimal. Analytics failures are already swallowed in `track()` and never break the user flow. An unknown event name would silently record with `unknown: true`; the parity test prevents that by asserting the name exists in the server allowlist too. Rollback is a straight revert of these four files.

## Goal alignment
None of simpler / faster / more beautiful directly — this is internal instrumentation. It earns its place by giving the Chat surface the usage evidence the Surface-lifecycle policy requires to make a keep/remove verdict, so effort stays on surfaces users actually use.


aalemayhu (commented):
**Verdict**: approve. Adds chat_message_sent, the only per-message usage event on the Chat surface, needed to get a keep/remove verdict this surface currently can't produce from data alone. Security: confirmed content-free payload (is_new_conversation only, no message/conversation text). Engineering: event-name parity confirmed in both KNOWN_EVENTS lists, single fire point confirmed (once per submission, not per token/re-render/regenerate), no layer or mocking violations.

PR #4527, opened 2026-09-22
chore: fire landing_page_viewed on SEO landing pages and pricing

aalemayhu:
## What
Fires the existing `landing_page_viewed` analytics event from two more top-of-funnel surfaces: the shared `LandingPage` component (one mount point covering all ~20 SEO topic pages — NotionToAnki, PdfToAnki, QuizletToAnki, PowerpointToAnki, McatAnki, etc.) and `PricingPage`. Every call — including HomePage's existing one — now carries a `{ path }` prop so the funnel can tell the entry pages apart.

## Why
Only `HomePage` fired `landing_page_viewed`, so the entire SEO landing family and `/pricing` were invisible to funnel analytics — only homepage visits counted as "Landing." That undercounts real top-of-funnel traffic and makes SEO-page conversion impossible to measure. Reusing the existing event (already in both `KNOWN_EVENTS` allowlists) keeps this a pure instrumentation change with no new event name.

## How
- Reused the existing `track('landing_page_viewed', ...)` — no new event, no allowlist change.
- Added a `{ path }` prop (the distinguishing prop the call previously lacked): `/` for HomePage, `copy.pathname` for each landing page (canonical slug regardless of `/convert/*` aliasing), `/pricing` for PricingPage. `path` is not in the server's PII-strip pattern (`/email|token|password|filename|content|title/i`) and the value is a short slug well under the 1 KB props cap, so no server change is needed.
- Fired once per mount via the same `useRef` guard + `useEffect` pattern HomePage already uses (mirrored, not reinvented).

## Measuring success
Query the events sink for `landing_page_viewed` grouped by `props->>'path'`: rows with `path` = `/notion-to-anki`, `/pdf-to-anki`, `/pricing`, etc. should appear within minutes of deploy (previously only `/` — actually absent before this PR — was recorded). Day-7: SEO landing pages and `/pricing` show non-zero "Landing" volume in the funnel breakdown.

## Testing
- Unit tests added:
  - `LandingPage.test.tsx` — `it.each` over notion + pdf copy asserting `landing_page_viewed` fires exactly once on mount tagged with the page's `path`.
  - `PricingPage.test.tsx` — asserts a single `landing_page_viewed` with `{ path: '/pricing' }` on mount.
  - `HomePage.test.tsx` — existing mount test tightened to assert `{ path: '/' }`.
- `pnpm --filter 2anki-web` — the 3 touched files: 83 tests pass. Web typecheck clean. `pnpm format:check` (tree-wide) clean. Web lint: warnings only, all pre-existing, none on the added lines.

## Browser check
Browser check: not applicable — event-only instrumentation, no rendered output change.

## Changelog
No entry — internal analytics instrumentation, no user-visible change.

## Trio
No trio — pure instrumentation: no UI, copy, or design decision.

## Risks
Low. No new event name, no server change, no rendered-output change. Worst case an over-count if a page double-mounts, but the `useRef` guard fires once per mount. Rollback = revert the branch.

## Goal alignment
Serves none of simpler / faster / more beautiful directly — it is funnel instrumentation. It earns its place by making SEO-landing and pricing top-of-funnel volume measurable, which is the prerequisite for deciding where to invest in making the product faster/simpler. Metric: `landing_page_viewed` count by `props->>'path'`, read from the events sink, checked day-7 post-deploy.


aalemayhu (commented):
**Verdict**: approve. Fixes the top-of-funnel blind spot — only the homepage fired landing_page_viewed, so all ~19 SEO landing pages and PricingPage were invisible to funnel analytics. Reuses the existing event, no new allowlist entries. Security: path is always a fixed developer string, never request-derived. Engineering: single fire point confirmed to cover all topic pages via the shared LandingPage component, mount-once guard mirrors HomePage's existing pattern exactly, no KNOWN_EVENTS touched.

PR #4530, opened 2026-09-22
fix: darken primary button color to meet WCAG AA contrast

aalemayhu:
## What
Darkens the filled primary-button color token (`--color-primary`) plus its hover and focus-ring in the **light**, **purple**, and **hotpink** themes so the light/near-white button label meets WCAG AA (4.5:1) contrast. Dark and gold already passed and are untouched.

## Why
An accessibility audit found the filled primary button (`background: var(--color-primary)`, `color: var(--color-bg-primary)`) fails WCAG AA for normal/semibold text in 3 of 5 themes: light 3.68:1, purple 4.17:1, hotpink 3.35:1. The primary button is a core interaction surface rendered across the app; failing AA text contrast on it is a real accessibility defect.

## How
One shared-token change in `web/src/styles/base.css` per affected theme (9 lines total):
- **light**: `--color-primary` `#3b82f6` → `#2563eb`, hover `#2563eb` → `#1d4ed8` (= the theme's existing `--color-text-link-hover`), `--color-focus-ring` rgba updated from the old-primary literal to `rgba(37, 99, 235, 0.5)`.
- **purple**: `--color-primary` `#7c6cc4` → `#6b59b3`, hover `#6b59b3` → `#5d4ea3` (= existing `--color-text-link-hover`), focus-ring → `rgba(107, 89, 179, 0.45)`.
- **hotpink**: `--color-primary` `#ec4899` → `#cf1f6b`, hover `#db2777` → `#9d174d` (= existing `--color-text-link-hover`), focus-ring → `rgba(207, 31, 107, 0.45)`.

The 6 files that hardcode white text on `var(--color-primary)` inherit the fix automatically because they reference the token for their background — confirmed in `AnkifyPage.module.css` (`background: var(--color-primary)`) and `ReviewPanel.module.css` (`background: var(--color-primary); color: #ffffff`).

## Measuring success
No runtime metric — this is a static contrast correction. Verification is the recomputed WCAG relative-luminance contrast ratio of each new primary against its theme's `--color-bg-primary`, independently computed (not taken from the designer's numbers):

| Theme | New `--color-primary` | `--color-bg-primary` | Ratio | AA (≥4.5:1) |
| --- | --- | --- | --- | --- |
| light | `#2563eb` | `#ffffff` | 5.17:1 | PASS |
| purple | `#6b59b3` | `#faf9fd` | 5.41:1 | PASS |
| hotpink | `#cf1f6b` | `#fff7f9` | 4.91:1 | PASS |

Old-value ratios reproduced exactly against the audit (3.68 / 4.17 / 3.35). Re-runnable against `base.css` at any time.

## Testing
- Unit tests added: none — this is a CSS custom-property value change; grep confirmed no test asserts on the old or new hex values, so there is no stale assertion to update.
- `pnpm --filter 2anki-web typecheck`: pass
- `pnpm --filter 2anki-web test`: 429 files / 4049 tests passed (full suite, not scoped — shared base file)
- `pnpm format:check`: clean (tree-wide)

## Browser check
Browser check: not applicable — no dev server or Chrome channel is available in this environment to visually verify; the change is a value-only edit to shared design tokens and the contrast improvement was verified by independent WCAG relative-luminance computation.

## Changelog
No changelog entry — accessibility color correction, not a new capability. Colors shift slightly darker but brand identity and layout are unchanged.

## Risks
Low. Only 9 token values change across 3 themes; dark and gold are untouched. Every new primary is darker than the old one, so any surface using `--color-primary` as a background with light text gains contrast, never loses it. Rollback: revert the single commit.

## Goal alignment
More beautiful — the primary button now reads clearly and meets WCAG AA in every theme. Metric: WCAG contrast ratio of primary-on-background per theme (light 5.17, purple 5.41, hotpink 4.91), computed directly from `base.css`; no ongoing prod metric since it is a static value.


aalemayhu (commented):
**Verdict**: approve. Fixes a real WCAG AA failure on the filled primary-button pattern in 3 of 5 themes (light 3.68→5.17, purple 4.17→5.41, hotpink 3.35→4.91), cascading correctly to ~65 dependent files via the shared token. Engineering: scope confirmed exactly 9 lines across the 3 named themes, no stale test references to the old hex values. A11y: contrast ratios independently recomputed and confirmed for all three, hover states confirmed still darker than resting (no lighter-on-hover regression), confirmed base.css has no text-on-primary usage that this could regress. Literal diff double-checked by the orchestrator against both reviewers' claims — matches exactly.

